PR #2, opened 2022-03-03
feat(across-v2):add react-query hooks 

Gamaranto:
This PR replaces the current redux toolkit hooks with `react-query` ones.

### DIFF
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -1,7 +1,7 @@
 import { useQuery } from 'react-query';
 import { useConnection } from 'state/hooks';
 import { useBlockNumber } from './useBlockNumber';
-import { getBalance } from 'utils';
+import { getBalance, getBalances } from 'utils';
 
 /**
  * @param token - The token to fetch the balance of.
@@ -13,9 +13,9 @@ import { getBalance } from 'utils';
 export function useBalance(token: string, account?: string, blockNumber?: number) {
 	const { chainId, account: connectedAccount } = useConnection();
 	const accountToQuery = account ?? connectedAccount;
-	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId ?? 1);
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId);
 	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
-	const { data: balance, ...delegated } = useQuery(["balance", accountToQuery, blockNumberToQuery], async () => {
+	const { data: balance, ...delegated } = useQuery(["balance", token, accountToQuery, blockNumberToQuery], async () => {
 		const balance = await getBalance(chainId!, token, accountToQuery!, blockNumberToQuery);
 		return balance;
 	}, {
@@ -25,6 +25,31 @@ export function useBalance(token: string, account?: string, blockNumber?: number
 	});
 	return {
 		balance,
+		...delegated
+	}
+}
+
+/**
+ * 
+ * @param tokens - The tokens to fetch the balance of.
+ * @param account - The account to query the balance of 
+ * @param blockNumber - The preferred block number to execute the query. Note, past blocks require an archive node. 
+ */
+export function useBalances(tokens: string[], account?: string, blockNumber?: number) {
+	const { chainId, account: connectedAccount } = useConnection();
+	const accountToQuery = account ?? connectedAccount;
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId);
+	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
+	const { data: balances, ...delegated } = useQuery(["balances", tokens, accountToQuery, blockNumberToQuery], async () => {
+		const balances = await getBalances(chainId!, tokens, accountToQuery!, blockNumberToQuery);
+		return balances;
+	}, {
+		enabled: !!chainId && !!accountToQuery && tokens.length > 0,
+		// We already re-fetch when the block change, so we don't need to re-fetch on an interval.
+		refetchInterval: false
+	});
+	return {
+		balances,
 		...delegated
 	}
 }
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -13,7 +13,7 @@ import { getBalance } from 'utils';
 export function useBalance(token: string, account?: string, blockNumber?: number) {
 	const { chainId, account: connectedAccount } = useConnection();
 	const accountToQuery = account ?? connectedAccount;
-	const latestBlockNumber = useBlockNumber(chainId ?? 1);
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId ?? 1);
 	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
 	const { data: balance, ...delegated } = useQuery(["balance", accountToQuery, blockNumberToQuery], async () => {
 		const balance = await getBalance(chainId!, token, accountToQuery!, blockNumberToQuery);
--- a/src/hooks/useBlockNumber.ts
+++ b/src/hooks/useBlockNumber.ts
@@ -1,29 +1,23 @@
-import { useLayoutEffect, useEffect, useState } from 'react'
+import { useQuery } from 'react-query';
 import { PROVIDERS, ChainId } from 'utils';
-
 
 
 /**
  * 
  * @param chainId The chain Id of the chain to poll for new blocks.
- * @returns The latest block number of the chain, if fetched.
+ * @returns The balance of the account and the UseQueryResult object
  */
-export function useBlockNumber(chainId: ChainId): number | undefined {
-	const [blockNumber, setBlockNumber] = useState<number | undefined>();
-	useLayoutEffect(() => {
+export function useBlockNumber(chainId: ChainId) {
+	const { data: blockNumber, ...delegated } = useQuery(["block", chainId], async () => {
 		const provider = PROVIDERS[chainId]();
-		if (!blockNumber) {
-			provider.getBlockNumber().then(setBlockNumber);
-		}
-	}, [blockNumber, chainId]);
-	useEffect(() => {
-		const provider = PROVIDERS[chainId]();
-		provider.on('block', (blockNumber) => {
-			setBlockNumber(blockNumber);
-		});
-		return () => {
-			provider.removeAllListeners('block');
-		}
-	}, [chainId])
-	return blockNumber;
+		const blockNumber = await provider.getBlockNumber();
+		return blockNumber;
+	}, {
+		// refetch every 5 seconds
+		refetchInterval: 5 * 1000,
+	});
+	return {
+		blockNumber,
+		...delegated
+	}
 }
--- a/src/hooks/useBlockNumber.ts
+++ b/src/hooks/useBlockNumber.ts
@@ -7,17 +7,26 @@ import { PROVIDERS, ChainId } from 'utils';
  * @param chainId The chain Id of the chain to poll for new blocks.
  * @returns The balance of the account and the UseQueryResult object
  */
-export function useBlockNumber(chainId: ChainId) {
+export function useBlockNumber(chainId?: ChainId) {
 	const { data: blockNumber, ...delegated } = useQuery(["block", chainId], async () => {
-		const provider = PROVIDERS[chainId]();
-		const blockNumber = await provider.getBlockNumber();
-		return blockNumber;
+		return await getBlockNumber(chainId!);
 	}, {
 		// refetch every 5 seconds
 		refetchInterval: 5 * 1000,
+		enabled: !!chainId,
 	});
 	return {
 		blockNumber,
 		...delegated
 	}
+}
+
+/**
+ * 
+ * @param chainId The chain Id of the chain to fetch the latest block from.
+ * @returns A promise resolving to the latest block number.
+ */
+async function getBlockNumber(chainId: ChainId): Promise<number> {
+	const provider = PROVIDERS[chainId]();
+	return provider.getBlockNumber();
 }
--- a/src/hooks/useBridgeFees.ts
+++ b/src/hooks/useBridgeFees.ts
@@ -1,0 +1,31 @@
+import { useQuery } from 'react-query';
+import { ethers } from 'ethers';
+import { getBridgeFees, PROVIDERS } from 'utils';
+import { useConnection } from 'state/hooks';
+import { useBlockNumber } from './useBlockNumber';
+
+
+
+/**
+ * 
+ * @param amount - The amount to check bridge fees for.
+ * @param tokenSymbol - The token symbol to check bridge fees for. 
+ * @returns The bridge fees for the given amount and token symbol and the UseQueryResult object. 
+ */
+export function useBridgeFees(amount: ethers.BigNumber, tokenSymbol: string) {
+	const { chainId } = useConnection();
+	const { blockNumber } = useBlockNumber(chainId);
+	const { data: fees, ...delegated } = useQuery(['bridgeFees', blockNumber, chainId], async () => {
+		const provider = await PROVIDERS[chainId!]();
+		const blockTimestamp = await provider.getBlock(blockNumber!).then(block => block.timestamp);
+		return getBridgeFees({ amount, tokenSymbol, blockTimestamp });
+	}, {
+		enabled: !!blockNumber && !!chainId,
+	});
+	return {
+		fees,
+		...delegated
+	}
+}
+
+
--- a/src/utils/balance.ts
+++ b/src/utils/balance.ts
@@ -18,3 +18,7 @@ export function getBalance(chainId: ChainId, token: string, account: string, blo
 	const contract = clients.erc20.connect(token, provider);
 	return contract.balanceOf(account, { blockTag: blockNumber ?? "latest" })
 }
+
+export function getBalances(chainId: ChainId, tokens: string[], account: string, blockNumber?: number): Promise<ethers.BigNumber[]> {
+	return Promise.all(tokens.map(token => getBalance(chainId, token, account, blockNumber)));
+}
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -134,6 +134,53 @@ export async function getLpFee(
   return result;
 }
 
+
+type GetBridgeFeesArgs = {
+  amount: ethers.BigNumber;
+  tokenSymbol: string;
+  blockTimestamp: number;
+};
+
+type GetBridgeFeesResult = BridgeFees & {
+  isAmountTooLow: boolean;
+  isLiquidityInsufficient: boolean;
+};
+
+class FailedFeeCalculationError extends Error {
+  constructor() {
+    super();
+    this.name = this.constructor.name;
+    this.message = `Failed to calculate bridge fees`;
+  }
+}
+/**
+ * 
+ * @param amount - amount to bridge
+ * @param tokenSymbol - symbol of the token to bridge
+ * @param blockTimestamp - timestamp of the block to use for calculating fees on
+ * @returns Returns the `slowRelayFee`, `instantRelayFee`, and `lpFee` fees for bridging the given amount of tokens, along with an `isAmountTooLow` flag indicating whether the amount is too low to bridge and an `isLiquidityInsufficient` flag indicating whether the liquidity is insufficient. 
+ */
+export async function getBridgeFees({ amount, tokenSymbol, blockTimestamp }: GetBridgeFeesArgs): Promise<GetBridgeFeesResult> {
+  const { instantRelayFee, slowRelayFee, isAmountTooLow } =
+    await getRelayFees(tokenSymbol, amount);
+
+  const { isLiquidityInsufficient, ...lpFee } = await getLpFee(
+    tokenSymbol,
+    amount,
+    blockTimestamp
+  );
+
+  return {
+    instantRelayFee,
+    slowRelayFee,
+    lpFee,
+    isAmountTooLow,
+    isLiquidityInsufficient,
+  };
+}
+
+
+
 export const getEstimatedDepositTime = (chainId: ChainId) => {
   switch (chainId) {
     case ChainId.OPTIMISM:


src/hooks/useBalance.ts
```diff
@@ -27,4 +27,29 @@ export function useBalance(token: string, account?: string, blockNumber?: number
...
+ * @param tokens - The tokens to fetch the balance of.
+ * @param account - The account to query the balance of 
+ * @param blockNumber - The preferred block number to execute the query. Note, past blocks require an archive node. 
+ */
+export function useBalances(tokens: string[], account?: string, blockNumber?: number) {
+	const { chainId, account: connectedAccount } = useConnection();
+	const accountToQuery = account ?? connectedAccount;
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId);
+	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
+	const { data: balances, ...delegated } = useQuery(["balances", tokens, accountToQuery, blockNumberToQuery], async () => {
```
daywiss, line 43: i feel like these calls could be made a lot simpler by removing the requirement of blockNumber and setting RQ to throttle underlying calls to once every X seconds, ie invalidate cache every X seconds.

src/hooks/useBlockNumber.ts
```diff
@@ -1,29 +1,32 @@
...
  * @param chainId The chain Id of the chain to poll for new blocks.
- * @returns The latest block number of the chain, if fetched.
+ * @returns The balance of the account and the UseQueryResult object
+ */
+export function useBlockNumber(chainId?: ChainId) {
+	const { data: blockNumber, ...delegated } = useQuery(["block", chainId], async () => {
+		return await getBlockNumber(chainId!);
+	}, {
+		// refetch every 5 seconds
+		refetchInterval: 5 * 1000,
```
daywiss, line 15: would be great if this coudl be configurable, or even configurable per chain since each one will have different block times

src/hooks/useBalance.ts
mrice32, line 43: What does delegated mean?

src/utils/balance.ts
```diff
@@ -18,3 +18,7 @@ export function getBalance(chainId: ChainId, token: string, account: string, blo
 	const contract = clients.erc20.connect(token, provider);
 	return contract.balanceOf(account, { blockTag: blockNumber ?? "latest" })
 }
+
+export function getBalances(chainId: ChainId, tokens: string[], account: string, blockNumber?: number): Promise<ethers.BigNumber[]> {
+	return Promise.all(tokens.map(token => getBalance(chainId, token, account, blockNumber)));
+}
```
mrice32, line 24: nit: we don't have to do this in this PR, but could we add a TODO to use multicall for this? This would allow us to grab all balances in a single node call. https://github.com/makerdao/multicall

src/hooks/useBlockNumber.ts
```diff
@@ -1,29 +1,32 @@
...
-			setBlockNumber(blockNumber);
-		});
-		return () => {
-			provider.removeAllListeners('block');
-		}
-	}, [chainId])
-	return blockNumber;
+async function getBlockNumber(chainId: ChainId): Promise<number> {
+	const provider = PROVIDERS[chainId]();
+	return provider.getBlockNumber();
```
daywiss, line 31: u might as well just get latest block and save timestamp and number since u have to use timestamp

src/hooks/useBridgeFees.ts
```diff
@@ -0,0 +1,31 @@
...
+
+
+/**
+ * 
+ * @param amount - The amount to check bridge fees for.
+ * @param tokenSymbol - The token symbol to check bridge fees for. 
+ * @returns The bridge fees for the given amount and token symbol and the UseQueryResult object. 
+ */
+export function useBridgeFees(amount: ethers.BigNumber, tokenSymbol: string) {
+	const { chainId } = useConnection();
```
daywiss, line 16: i feel like this is one of those things that we might want to have as a parameter, since fees arent necessary based off of what chain you are connected to, but rather what your source/destination chains are.

src/hooks/useBridgeFees.ts
```diff
@@ -0,0 +1,31 @@
...
+	const { blockNumber } = useBlockNumber(chainId);
+	const { data: fees, ...delegated } = useQuery(['bridgeFees', blockNumber, chainId], async () => {
+		const provider = await PROVIDERS[chainId!]();
+		const blockTimestamp = await provider.getBlock(blockNumber!).then(block => block.timestamp);
+		return getBridgeFees({ amount, tokenSymbol, blockTimestamp });
```
daywiss, line 21: i dont remember what timestamp we are supposed to be using, the source chain timestamp or the destination chain timestamp? i think this is right though, its based on looking up the timestamp of the source chain to find the block on the destination chain, but just want to give this a closer look

src/hooks/useBalance.ts
```diff
@@ -13,9 +13,9 @@ import { getBalance } from 'utils';
 export function useBalance(token: string, account?: string, blockNumber?: number) {
 	const { chainId, account: connectedAccount } = useConnection();
 	const accountToQuery = account ?? connectedAccount;
-	const latestBlockNumber = useBlockNumber(chainId ?? 1);
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId);
 	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
-	const { data: balance, ...delegated } = useQuery(["balance", accountToQuery, blockNumberToQuery], async () => {
+	const { data: balance, ...delegated } = useQuery(["balance", token, accountToQuery, blockNumberToQuery], async () => {
```
daywiss, line 18: we might want to consider doing some kind of mapping to these query keys. doesnt necessarily need to be in this PR but before this gets super complicated might be worth thinking about

src/hooks/useBalance.ts
daywiss: i guess now that i think about it, this is meant more to have a way to auto update the balance on an interval

src/hooks/useBlockNumber.ts
```diff
@@ -1,29 +1,32 @@
...
 
 
-
 /**
  * 
  * @param chainId The chain Id of the chain to poll for new blocks.
- * @returns The latest block number of the chain, if fetched.
+ * @returns The balance of the account and the UseQueryResult object
+ */
+export function useBlockNumber(chainId?: ChainId) {
```
mrice32, line 10: Why block number? Why not block? From the block, you can get both the timestamp and the block number, meaning you won't have to make subsequent calls to get the block timestamp.

I think `const latestBlock = await provider.getBlock("latest")` would work for this.

mrice32 (commented):
Looks great! A few comments

src/hooks/useBalance.ts
Gamaranto: I like to call it "delegated" to signal that they're not used in the hook and only passed down, could be called "others" or any other name

src/utils/balance.ts
Gamaranto: Yes! I wanted to do it with multicall but didn't bother fetching all the addresses on all the chains, could also be done with the geth graphql endpoint by reading storage directly (not sure infura supports it)

src/hooks/useBalance.ts
Gamaranto: My reasoning was to have all queries execute against a fixed block number (so they are all consistent with each other), and poll blocks in the background (all queries will re-run anyway once the block number changes)

src/hooks/useBlockNumber.ts
Gamaranto: Agree! I think it makes sense to add this on a chain config, wdyt?

src/hooks/useBalance.ts
Gamaranto: I think its actually worth to block this and standardize it before things get out of hand, what would you think of having a map of queries (balance, balances etc...) to function calculating query given the args passed? Open to other ideas as well

src/hooks/useBridgeFees.ts
Gamaranto: Oh this is a great point, my implementation is plain wrong since if you're connected to an unsupported chain the fee calc will fail

src/hooks/useBlockNumber.ts
daywiss: yea it makes sense there

src/hooks/useBalance.ts
daywiss: i agree with this. i think because these are pure, you can just have a single file export the functions named identically to the top level key, and then parameters for the rest of the values. 
```js
// exported from some file, maybe "react-query-keys.ts" or something
export function balance(token,account,blockNumber){
  return ['balance',token,account,blockNumber]
}
```

src/hooks/useBalance.ts
daywiss: i see. im not sure i agree that imposing that constraint is necessary, as it has a cost to infura usage. you could imagine doing the same thing with just a timestamp that runs on an interval, any inconsitency with the balances woudl eventually be corrected.  the only part of the system that actually needs a block query is the fee calculation, which can be allowed to be stale for up to 30 seconds or something. 

anyways the only reason why i push on this is because we have really high infura usage in v1, so if there are any constraints we can relax to help this ( ie using timers rather than block polling) i think we should consider.

src/hooks/useBalance.ts
```diff
@@ -27,4 +27,29 @@ export function useBalance(token: string, account?: string, blockNumber?: number
...
+export function useBalances(tokens: string[], account?: string, blockNumber?: number) {
+	const { chainId, account: connectedAccount } = useConnection();
+	const accountToQuery = account ?? connectedAccount;
+	const { blockNumber: latestBlockNumber } = useBlockNumber(chainId);
+	const blockNumberToQuery = blockNumber ?? latestBlockNumber;
+	const { data: balances, ...delegated } = useQuery(balancesQueryKey(chainId!, tokens, accountToQuery!, blockNumberToQuery), async () => {
+		const balances = await getBalances(chainId!, tokens, accountToQuery!, blockNumberToQuery);
+		return balances;
+	}, {
+		enabled: !!chainId && !!accountToQuery && tokens.length > 0,
```
Tulun, line 47: I would probably either extract this enabled variable out and/or also comment as to why these particular variables are part of this call.

As it's actually a big deal this call doesn't fire unexpectedly, as react-query tends to want to overfetch.

src/hooks/useBridgeFees.ts
Tulun: +1

src/hooks/useBridgeFees.ts
Tulun: Also -- shouldn't this callback function be pulled out as a pure function? That's part of the pattern discussed.

src/utils/query-keys.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export function balancesQueryKey(chainId: ChainId, tokens: string[], account: string, blockNumber?: number) {
+	return ["balances", chainId, tokens, account, blockNumber];
+}
+/**
+ * @param chainId
+ * @param tokenSymbol
+ * @param blockNumber
+ * @returns An array of query keys for react-query `useQuery` hook.
+ */
```
Tulun, line 32: Make sure to complete this documentation, otherwise I probably wouldn't add it in this form.

Gamaranto: do you have any suggestions on how to improve it?

src/hooks/useBlockNumber.ts
Gamaranto: changed it to fetch the whole block

src/hooks/useBridgeFees.ts
Gamaranto: I changed `useBlockNumber` to be `useBlock` so now its a one liner, also yes its using the source chain destination right now, but lets check this again before release

src/hooks/useBlockNumber.ts
Gamaranto: good catch, it fetches the whole block now

src/hooks/useBlockNumber.ts
mrice32: Awesome, yeah, I think fetching the entire block right here so we don't have to do it elsewhere is preferable.

src/utils/query-keys.ts
mrice32: I think he's saying add descriptions to the parameters.
```
 * @param tokenSymbol symbol for the token whose fees this key is used to look up.
```

Tulun: > I think he's saying add descriptions to the parameters.
> 
> ```
>  * @param tokenSymbol symbol for the token whose fees this key is used to look up.
> ```

Yeah, this is a lot better. Basically it seems unnecessary to just say @param tokenSymbol or @param blockNumber -- give a more in depth description, or I can just read the variable and type in the function declaration.

src/hooks/useBlock.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/**
+ * @param chainId The chain Id of the chain to fetch the latest block from.
+ * @param blockHashOrBlockTag The block hash or block tag to fetch, defaults to "latest".
+ * @returns A promise resolving to the requested block.
+ */
+async function getBlock(chainId: ChainId, blockHashOrBlockTag?: ethers.providers.BlockTag): Promise<ethers.providers.Block> {
+	const provider = PROVIDERS[chainId]();
+	const hashOrTag = blockHashOrBlockTag ?? 'latest';
```
mrice32, line 31: nit: prefer to do defaulting in the function definition as it makes the function a bit simpler and more readable IMO:

```ts
getBlock(chainId: ChainId, blockHashOrBlockTag?: ethers.providers.BlockTag = 'latest')
```

src/hooks/useBridgeFees.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @remarks This hook **SHOULD NOT** be used to calculate the fees for a L1 to L2 transfer, as those use canonical bridges instead.
+ * @param amount - The amount to check bridge fees for.
+ * @param tokenSymbol - The token symbol to check bridge fees for. 
+ * @returns The bridge fees for the given amount and token symbol and the UseQueryResult object. 
+ */
+export function useBridgeFees(amount: ethers.BigNumber, tokenSymbol: string) {
+	const { chainId } = useConnection();
+	const { block } = useBlock(chainId);
+
+	const { data: fees, ...delegated } = useQuery(bridgeFeesQueryKey(chainId!, tokenSymbol, block?.number), async () => {
```
mrice32, line 20: nit: same comment as above. We expect this function to be called without the chainId being defined, so the chainId! assertion causes an invalid cast sometimes. Prefer to only call the `bridgeFeesQueryKey` function when chainId is defined.

src/utils/balance.ts
mrice32: Thanks!!

src/utils/query-keys.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Generates query keys for react-query `useQuery` hook, used in the `useBalances` hook.
+ * @param chainId 
+ * @param tokens 
+ * @param account 
+ * @param blockNumber 
+ * @returns An array of query keys for react-query `useQuery` hook.
+ */
+export function balancesQueryKey(chainId: ChainId, tokens: string[], account: string, blockNumber?: number) {
```
mrice32, line 33: What happens when a value in this list is undefined? Does it leave it out of the key computation entirely? Or does it put "undefined" into the string key?

Also, why should we ever allow blockNumber to be undefined? Shouldn't it _always_ be the block number that it was queried on (even if that's the latest)?

src/hooks/useBalance.ts
```diff
@@ -27,4 +27,31 @@ export function useBalance(token: string, account?: string, blockNumber?: number
...
+ * 
+ * @param tokens - The tokens to fetch the balance of.
+ * @param account - The account to query the balance of 
+ * @param blockNumber - The preferred block number to execute the query. Note, past blocks require an archive node. 
+ */
+export function useBalances(tokens: string[], account?: string, blockNumber?: number) {
+	const { chainId, account: connectedAccount } = useConnection();
+	const accountToQuery = account ?? connectedAccount;
+	const { block: latestBlock } = useBlock(chainId);
+	const blockNumberToQuery = blockNumber ?? latestBlock?.number;
```
mrice32, line 42: Can the latest block not have its `.number` defined? When would we expect this to happen? Should we proceed with the query if our blockNumberToQuery is undefined? What's the desired behavior?

mrice32 (commented):
Looks great! Mostly minor comments on style, defaulting behavior, and query keys.

src/hooks/useBalance.ts
```diff
@@ -27,4 +27,31 @@ export function useBalance(token: string, account?: string, blockNumber?: number
...
+	// To fetch balances, we need a list of tokens, an account to get balances of, and a specified chainId.
+	const enabledQuery = !!chainId && !!accountToQuery && tokens.length > 0
+	const { data: balances, ...delegated } = useQuery(balancesQueryKey(chainId!, tokens, accountToQuery!, blockNumberToQuery), async () => {
```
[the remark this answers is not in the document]
daywiss: you could also allow the balancesQueryKey to take in undefined values for params, and just return this disabled string if not all values exist. This way you can kind of encapsulate all this logic in each function rather than bringing it out into the hook.  I agree though generally, it would be nice to be able to avoid the type coersions

src/hooks/useBlock.ts
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * @param chainId The chain Id of the chain to poll for new blocks.
+ * @returns The latest block, and the useQueryResult object.
+ */
+export function useBlock(chainId?: ChainId) {
+	const { data: block, ...delegated } = useQuery(["block", chainId], async () => {
+		return await getBlock(chainId!);
+	}, {
+		// refetch based on the chain polling interval
+		refetchInterval: CHAINS[chainId ?? 1]?.pollingInterval ?? 5 * 1000,
```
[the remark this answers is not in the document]
daywiss: side note here, not to hijack matts point, but i really dislike the use of the default chain. this is a source of bugs. we've run into issues where there should have been errors, but things operate as if its ok because some value we thought was defined, was undefined and logic was defaulting to mainnet.  The opinion of what defaults to use should be set at the very highest level of the app, not the lowest level.

daywiss (commented):
i know there are a lot of comments, but this is looking good.

src/hooks/useBalance.ts
```diff
@@ -27,4 +27,31 @@ export function useBalance(token: string, account?: string, blockNumber?: number
 		balance,
 		...delegated
 	}
+}
+
+/**
+ * 
+ * @param tokens - The tokens to fetch the balance of.
+ * @param account - The account to query the balance of 
+ * @param blockNumber - The preferred block number to execute the query. Note, past blocks require an archive node.
```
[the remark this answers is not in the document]
Gamaranto: sounds good, I changed it to this:
`The block number to execute the query on, if not specified, defaults to the latest block. Note, past blocks require an archive node.`

src/hooks/useBlock.ts
Gamaranto: good idea, changed it

src/hooks/useBlock.ts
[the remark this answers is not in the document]
mrice32: I agree. I think defaults often hide errors.

src/hooks/useBalance.ts
mrice32: Ah, you're totally right. I misread the syntax. I agree that blockNumberToQuery should be defined or the query should be disabled. +1

src/utils/query-keys.ts
Gamaranto: I think it would put `undefined` in the string key, regarding `blockNumber`, it will always be undefined in the short period between when we start fetching it and when it is saved into state, in which case the query key has to handle it and the query should not run (it should be checked in the `enabled` flag)

src/hooks/useBlock.ts
```diff
@@ -0,0 +1,33 @@
...
+import { useQuery } from 'react-query';
+import { PROVIDERS, ChainId, CHAINS } from 'utils';
+
+/**
+ * 
+ * @param chainId The chain Id of the chain to poll for new blocks.
+ * @returns The latest block, and the useQueryResult object.
+ */
+export function useBlock(chainId?: ChainId) {
+	const { data: block, ...delegated } = useQuery(["block", chainId], async () => {
```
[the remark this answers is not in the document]
Gamaranto: yes, moved it

Gamaranto:
Regarding the query keys, I think the cleanest pattern is to make the key generating function as strict as possible, and handle the pass in a `DISABLED_QUERY_NAME` string as the key otherwise. 
I initially tried something like what you suggested @daywiss,  but I found it easier to be strict on the key factory and have TS warn you about the undefined variables vs having the function accept possibly undefined variables (and possibly re-implement the `enabledQuery` check).

src/hooks/useBlock.ts
```diff
@@ -0,0 +1,37 @@
...
+import { PROVIDERS, ChainId, CHAINS, latestBlockQueryKey } from 'utils';
+
+
+const DEFAULT_BLOCK_POLLING_INTERVAL = 5 * 1000
+/**
+ * Fetches the latest block from a given chain Id on an interval. 
+ * @param chainId The chain Id of the chain to poll for new blocks.
+ * @returns The latest block, and the useQueryResult object.
+ */
+export function useBlock(chainId?: ChainId) {
```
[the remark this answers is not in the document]
Tulun: In fact isn't ChainId always just a list the enum list of of ChainIds we've whitelisted?

mrice32:
> Regarding the query keys, I think the cleanest pattern is to make the key generating function as strict as possible, and handle the pass in a `DISABLED_QUERY_NAME` string as the key otherwise. I initially tried something like what you suggested @daywiss, but I found it easier to be strict on the key factory and have TS warn you about the undefined variables vs having the function accept possibly undefined variables (and possibly re-implement the `enabledQuery` check).

Agree with this. Seems simpler this way.

src/hooks/useBlock.ts
```diff
@@ -0,0 +1,37 @@
...
+	const enabledQuery = !!chainId;
+	const queryKey = enabledQuery ? latestBlockQueryKey(chainId) : "DISABLED_BLOCK_QUERY";
+	const refetchInterval = enabledQuery ? (CHAINS[chainId].pollingInterval ?? DEFAULT_BLOCK_POLLING_INTERVAL) : undefined;
```
mrice32, line 15: What happens when you pass undefined into useQuery? (I know enabled will be false, but I want to just make sure the library handles this robustly)

src/utils/constants.ts
```diff
@@ -730,4 +731,4 @@ export const CHAINS_SELECTION: ChainsSelection = [
 ];
 
 export const MAX_APPROVAL_AMOUNT = ethers.constants.MaxUint256;
-export const ETH_ADDRESS = ethers.constants.AddressZero;
\ No newline at end of file
+export const ETH_ADDRESS = ethers.constants.AddressZero;
```
mrice32, line 734: nit: I know you didn't add this, but I think this is really confusing, since this is really _what the SDK considers the ETH address to be_. However, you could have chains that actually have an ETH address (like optimism). I would suggest naming this something like ZERO_ADDRESS or SDK_ETH_ADDRESS or SDK_ETH_ADDRESS_PLACEHOLDER.

src/hooks/useBalance.ts
```diff
@@ -1,30 +1,60 @@
...
  * @param token - The token to fetch the balance of.
- * @param account - The account to query the balance of
- * @param blockNumber - The preferred block number to execute the query. Note, past blocks require an archive node.
+ * @param account - The account to query the balance of.
+ * @param blockNumber - The block number to execute the query on, if not specified, defaults to the latest block. Note, past blocks require an archive node.
  * @remarks Passing the zero address as token will return the ETH balance. Passing no account will return the balance of the connected account.
  * @returns The balance of the account and the UseQueryResult object
  */
 export function useBalance(token: string, account?: string, blockNumber?: number) {
 	const { chainId, account: connectedAccount } = useConnection();
```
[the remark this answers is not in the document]
mrice32: Agree.

mrice32 (approved):
A few more minor comments, but mostly LGTM!

src/hooks/useBalance.ts
```diff
@@ -27,4 +27,31 @@ export function useBalance(token: string, account?: string, blockNumber?: number
...
+	const { block: latestBlock } = useBlock(chainId);
+	const blockNumberToQuery = blockNumber ?? latestBlock?.number;
+	// To fetch balances, we need a list of tokens, an account to get balances of, and a specified chainId.
+	const enabledQuery = !!chainId && !!accountToQuery && tokens.length > 0
+	const { data: balances, ...delegated } = useQuery(balancesQueryKey(chainId!, tokens, accountToQuery!, blockNumberToQuery), async () => {
+		const balances = await getBalances(chainId!, tokens, accountToQuery!, blockNumberToQuery);
+		return balances;
+	}, {
+		enabled: enabledQuery,
+		// We already re-fetch when the block change, so we don't need to re-fetch on an interval.
```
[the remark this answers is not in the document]
mrice32: bump on this

src/hooks/useBalance.ts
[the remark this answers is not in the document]
Gamaranto: I think it would be confusing user wise to be connected to a network and have balances shown from another network (especially since the wallet shows your ETH balance on the network you're connected to), this is a product decision however, so I will add the possibility of passing a `chainId` defaulting to the network the user is connected to if not passed

src/hooks/useBlock.ts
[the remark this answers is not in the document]
Gamaranto: I could make `chainId` required here, but then each other hook / component consuming this hook would have to default it / choose a sensible fallback, and doing that in many places is potentially dangerous / annoying (have to keep the values in sync).
For how the dapp is currently structured, chainId is the __user__ chainId, which will always be undefined, rather than a "developer passed" chainId, we can def change this to be the other way around, where the user chainId is almost ignored and its up to the developer to pass a chainId to each hook, I'm don't have a strong opinion on either of those, but I would say we should think about this a bit, since its a significant ergonomic change in how the hook is used

src/hooks/useBlock.ts
Gamaranto: I think its the same thing as if you did not pass anything, that is it will use the library default, which is `false`.

src/utils/constants.ts
Gamaranto: Good point, I will make a note of this and fix it in a PR about constants

src/hooks/useBalance.ts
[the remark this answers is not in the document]
daywiss: Its true, but the context of where this is shown matters. This function has no context of what view its in, it could be the navbar where it makes sense to show eth on the chain u are connected to, or it could be in the send form, where its showing balances per currency in a dropdown of the chain you have selected from. Its a small assumption that limits usefulness of the hook. Again tho not a big deal now, hopefully not to hard to change later if we need to.

daywiss (approved):
thanks for your patience with this, great work.  my only nit is the `useBridgeFees` function being bound to your connected chain. Might want to change that before we forget, but i think we will be going back in that function any way in the future.

PR #3, opened 2022-03-07
feature: View for Transactions table

Tulun:
story: https://app.shortcut.com/uma-project/story/4741/create-view-of-transaction-history-for-an-account

Logged out view:
Screen Shot 2022-03-07 at 2 41 00 PM

Logged in view:
Screen Shot 2022-03-07 at 2 41 07 PM


src/components/Table/Table.styles.tsx
```diff
@@ -0,0 +1,104 @@
+import styled from "@emotion/styled";
+import { CellSize } from "./Table";
+
+const BREAKPOINTS = {
+  tabletMin: 550,
+  laptopMin: 1100,
+  desktopMin: 1500,
+};
+
+const QUERIES = {
```
Gamaranto, line 10: Any reason to not import those from the constants file?

src/views/Transactions/TransactionTable.styles.tsx
```diff
@@ -0,0 +1,62 @@
...
+  Cell,
+} from "components/Table/Table.styles";
+
+export const Wrapper = styled.div`
+  margin-top: 2rem;
+`;
+
+export const Title = styled.h2`
+  color: var(--color-white);
+  font-size: 1.25rem;
```
Gamaranto, line 16: I think we should use `${20 / 16}rem` to keep things consistent with other parts of the codebase

src/views/Transactions/Transactions.tsx
```diff
@@ -1,9 +1,33 @@
-import { Wrapper, Title, ConnectButton } from "./Transactions.styles";
+import { Wrapper, Title, ConnectButton, Account } from "./Transactions.styles";
+import useTransactionsView from "./useTransactionsView";
+import TransactionsTable from "./TransactionsTable";
+import { shortenAddressLong } from "utils/format";
+
 const Transactions = () => {
+  const { provider, isConnected, initOnboard, account } = useTransactionsView();
+
+  const handleClick = () => {
```
Gamaranto, line 9: nit: let's move this inside the hook

src/utils/format.ts
```diff
@@ -14,11 +14,16 @@ export function shortenAddress(address: string): string {
   return `${address.substr(0, 4)}...${address.substr(-4)}`;
 }
 
-export function shortenAddressLong(address: string): string {
+export function shortenAddressLong(address: string, numDots = 3): string {
```
daywiss, line 17: curious why the number of dots is configurable rather than the length of the address.  seems like you could combine shortenAddress and shortenAddressLong into a single function with a configurable length.  You could also just specify the delimiter you want to use, ie it defaults to "...", but the caller could enter "......" instead. Makes it a bit more generalized

src/views/Transactions/TransactionsTable.tsx
```diff
@@ -0,0 +1,281 @@
...
+          >
+            {shortenTransactionHash(
+              "0x0000000000000000000000000000000000000000"
+            )}
+          </TableLink>
+        ),
+      },
+    ],
+  },
+];
```
daywiss, line 237: why didnt you create stub data?  Would make a lot more sense as props to the table, exposed from the hook or something like that, rather than mixing it with code, wdyt?

src/views/Transactions/Transactions.tsx
```diff
@@ -1,9 +1,33 @@
...
+    }
+  };
+
   return (
     <Wrapper>
-      <Title>Transactions</Title>
-      <ConnectButton>Connect wallet</ConnectButton>
+      <Title>
+        Transactions
+        {isConnected && account && (
```
daywiss, line 19: just want to comment on this, nothing u need to change, but we should still keep in mind keeping an interface between view and hooks. It might be overkill in this case, but having a purely prop driven component would have been acceptable here, while deferring hook implementation to later.

src/utils/format.ts
Tulun: Sure, that's fair. 
It's largely because the existing function assumed "..." but we have an instance in this view where we want "....." instead. This was a way of avoiding a redundant function.
I could make the change you suggest, but it would break current compatibility wherever this is called. Which is fine, it would just mean I have to change it throughout the app. 

This implementation, with the optional "numDots" argument was my workaround to not change all called instances.

src/views/Transactions/TransactionsTable.tsx
Tulun: Ease of development. Was easier for me to work with stubbed data in a single file.

What you suggest will happen when I integrate with the hook ofc, I can do it now if you'd like now that the stubbed data is done.

src/views/Transactions/Transactions.tsx
Tulun: Where would you pass the props down from though? This is the parent component for the view -- only thing higher would be the router, which seems to be an odd place to place placeholder code.

src/components/Table/Table.styles.tsx
Tulun: This component was built from the component lib which had its dependencies baked in (like this constant)), I'll refactor it to just use the values we have built in app.

src/components/Table/Table.styles.tsx
```diff
@@ -0,0 +1,104 @@
...
+  &:first-of-type {
+    margin-left: 24px;
+  }
+  &:not(:first-of-type) {
+    min-width: 150px;
+  }
+`;
+
+export const Body = styled.div`
+  /* padding: 15px 0; */
```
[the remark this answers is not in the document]
Tulun: Done.

src/views/Transactions/Transactions.tsx
daywiss: you would need to have a view component and a stateful component that wraps the view

src/views/Transactions/Transactions.tsx
daywiss: i think he means to expose the handleClick from the hook, const {handleClick} = useTransactionsView(), which i think makes sense to me

Tulun: David and I I hashed this out, this function isn't even necessary -- just passing init directly.

src/utils/format.ts
Tulun: Created a more generic shortenAddress and shortenString function.

src/utils/format.ts
```diff
@@ -7,18 +7,46 @@ export function isValidString(s: string | null | undefined | ""): s is string {
...
+ */
+export function shortenString(
+  str: string,
+  delimiter: string,
+  numChars: number
+) {
+  return `${str.substring(0, numChars)}${delimiter}${str.substring(
+    str.length - numChars,
+    str.length
+  )}`;
```
mrice32, line 45: What if the string is already shorter than 2 * numChars? Should we not shorten in that case?

src/views/Transactions/useTransactionsView.tsx
```diff
@@ -0,0 +1,248 @@
...
+              "0x0000000000000000000000000000000000000000"
+            )}
+          </TableLink>
+        ),
+      },
+    ],
+  },
+];
+
+export default function useTransactionsView() {
```
mrice32, line 233: What's the purpose of this hook?

headerCells seems like a visual component (jsx) being exported through a hook.

The rest of this (especially transactions) seems like something that makes sense to do via hooks.

Why can't this file just be a component that takes props and then we have a separate hook file that does the rest of this logic (and would probably call sub-hooks for things like transactions, etc)?

Maybe this doesn't work, but I was hoping that the view logic would be 100% components and whenever a component needed data, it would have an accompanying hook that it would call to provide it. In this pattern, you would generally never see JSX passed through a hook. Does that work? Does that make sense?

src/views/Transactions/useTransactionsView.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const [transactions] = useState([]);
```
mrice32, line 234: nit: might be nice to add a type and/or sample data for transactions so it's clear what's expected when building out the hook that grabs the transactions from the blockchain.

mrice32 (commented):
Looks great! A few minor comments and one on the structure of the hook

src/views/Transactions/useTransactionsView.tsx
Tulun: What if the data provided needs to be an element?

The issue is that we are dealing with a DOM. The cell of a Table is either a string or an element -- if we want to store cells as piece of state data (which we have to, as it's dynamic), I don't see how we avoid creating JSX in a hook given it is part of the requirements of the feature.

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -112,7 +112,9 @@ const AddressSelection: React.FC = () => {
                       ? "Mainnet"
                       : selectedItem?.name}
                   </ToggleChainName>
-                  {toAddress && <Address>{shortenAddress(toAddress)}</Address>}
+                  {toAddress && (
+                    <Address>{shortenAddress(toAddress, "...", 4)}</Address>
```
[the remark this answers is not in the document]
Tulun: We decided to combine shortenAddress and shortenAddressLong into 1 function call and to make it more flexible, as we had 3 separate instances of "shortening" that did similar, but slightly different things.

src/views/Transactions/useTransactionsView.tsx
Tulun: I have no strong opinions here. I figure I'd wait until I actually do the react-query integration, but I'm honestly not using this data at the time and it can be deleted.

src/utils/format.ts
```diff
@@ -7,18 +7,46 @@ export function isValidString(s: string | null | undefined | ""): s is string {
...
-  if (!isValidString(address)) {
-    return "";
-  }
-  return `${address.substr(0, 10)}...${address.substr(-10)}`;
+/**
+ *
+ * @param str string to shorten
+ * @param delimiter string to put between your split string, eg: "...", "---"
+ * @param numChars number of characters to keep on each part of the string
+ * @returns
```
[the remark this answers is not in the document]
Tulun: Fixed.

src/views/Transactions/useTransactionsView.tsx
```diff
@@ -0,0 +1,245 @@
+import { TableLogo, TableLink } from "./TransactionTable.styles";
```
mrice32, line 1: As discussed IRL, @Tulun and I agreed that this is the structure we want:

TransactionsView.tsx:

```
function transformTransactions(transactions: Transaction[]) {
  const rows = transactions.map(transaction => {
    // some logic here
  });

   const header = transactions.map(transaction => {
    // some logic here
  });

  return { rows, header };
}

function TransactionTableComponent(...) {

  const { transactions,  } = useTransactionsView();
  
  const { rows, header } = transformTransactions(transactions);
  
  return <MyTableComponent rows=rows header=header>;

}
```


useTransactionsView.tsx:
```
function useTransactionsView() {
  // Stub
  const { transactions } = useTransactions();
  const { provider, chainId, isConnected, account } = useConnection();
  const { init } = onboard;

  return {
    provider,
    chainId,
    isConnected,
    account,
    transactions
  };
}
```

daywiss: Great solution

Tulun: I feel like we can break this task into its own feature, created the story for this.

https://app.shortcut.com/uma-project/story/4863/create-interface-and-convert-function-for-transactions

daywiss: im ok with doing this as a new pr

daywiss (approved):

src/views/Transactions/Transactions.tsx
```diff
@@ -1,9 +1,29 @@
-import { Wrapper, Title, ConnectButton } from "./Transactions.styles";
+import { Wrapper, Title, ConnectButton, Account } from "./Transactions.styles";
+import useTransactionsView from "./useTransactionsView";
+import TransactionsTable from "./TransactionsTable";
+import { shortenAddress } from "utils/format";
+import formatTransactions from "./formatTransactions";
+
 const Transactions = () => {
+  const { isConnected, initOnboard, account } = useTransactionsView();
+  const { rows, headers } = formatTransactions();
```
mrice32, line 9: Two questions:
1. Why does formatTransactions need its own file? It seems like it makes sense as a helper function in a component file rather than being some sort of external utility. Especially since it's tightly coupled with the table, itself.
2. It seems a bit odd to do this transformation in this file? Any reason why it shouldn't be done inside the table component since that logic is more tightly coupled to the table itself than the outside view? Note: I'm not super opinionated on where the hook is called since I could see wanting to consolidate into fewer hooks, but the transformation of that raw data into a format that the Table component understands seems like it should be a concern of the TransactionsTable component, not this one.

mrice32 (commented):
One remaining comment!

src/views/Transactions/Transactions.tsx
Tulun: 1.  Because components should not compose of a bunch separate functions. A component file should not consist of a a single view and then a bunch of unrelated, pure functions imo -- being in its own file keeps it clean. It would be jarring to me to come to a random Component in a new repo and find out it had a dozen of pure functions in it too.

2. We could move it down to the component itself sure.

Tulun: Tbh, we should actually consider attaching logic like this in a useCallback inside the component itself, because this transform file technically will re-run on every re-render of the component, not just when the Transactions object will change.

but that would break some of the design philosophy.

mrice32: Sounds good! I think it's fine to put in its own file.

One minor point to make on this, though. I don't think this is a utility function. This function is highly coupled with the component where it's being called (it would never be used elsewhere). I would argue that with that level of coupling, you would probably want the function and the view to be in the same file as it might be confusing to jump into a file with _just_ that function in it. The context to understand what that function is doing is the view where it's used -- it can't really be understood without that.

I think it totally makes sense to separate it into its own file as a way of managing file length/space, but I think we'd want to be wary of tossing into a file containing other, unrelated functions, as it might be difficult to understand where that function fits into the project.

src/views/Transactions/useTransactionTableValues.tsx
```diff
@@ -0,0 +1,66 @@
...
+    setOngoingRows(pending);
+    setFilledRows(filled);
+  }, []);
+
+  return {
+    headers,
+    ongoingRows,
+    filledRows,
+  };
+}
```
mrice32, line 23: I saw your comment about wanting to cache the result of the formatTransactions function.

I think that totally makes sense, but I don't think this design is in-line with our design philosophy, because the formatTransactions function that is meant to tell the Table how to _display_ the transaction data is running inside a hook. And our custom hooks shouldn't be performing any view-related logic.

I also think we should try to avoid having useEffects that cache intermediate values if at all possible. I think it could make state transitions for this table pretty difficult to reason about since we now have a multi-stage state update happening as a result of a single update to the transactions data.

First of all, I don't think we need to prematurely optimize this. If we find that the page feels sluggish because of this transformation function, we can so something about that, but a pure function that's just looping over an object shouldn't be too intensive.

Second, if we do want to optimize this now, I think there are two strategies we should use to avoid useEffect or coupling our view logic and our custom hook logic:
1. We could memoize the entire TransactionsTable component so it won't rerun its render cycle unless its props change.
2. We could just memoize the call to this formatTransactions function where we only rerun it if we see a new `transactions` object has been passed to us.

I think both of those would require only minimal changes to the version of the code before this useEffect was added. I also just think memoizing is a bit easier to reason about than useEffects!

Thoughts?

src/views/Transactions/Transactions.tsx
```diff
@@ -1,9 +1,38 @@
-import { Wrapper, Title, ConnectButton } from "./Transactions.styles";
+import { Wrapper, Title, ConnectButton, Account } from "./Transactions.styles";
+import useTransactionsView from "./useTransactionsView";
+import TransactionsTable from "./TransactionsTable";
+import { shortenAddress } from "utils/format";
+import useTransactionTableValues from "./useTransactionTableValues";
+
 const Transactions = () => {
+  const { isConnected, initOnboard, account } = useTransactionsView();
+  const { ongoingRows, filledRows, headers } = useTransactionTableValues();
```
Gamaranto, line 9: I know this comment is super annoying (and think this code is fine), but any reason we couldn't collapse those 2 hooks together? It would follow our rule of "1 component = 1 hook"

Tulun: Matt doesn't want JSX in the main hook, which is required for this other hook.

src/views/Transactions/useTransactionTableValues.tsx
Tulun: I wouldn't call this early optimization -- this will have to change as soon as we have dynamic data.

First suggestion we'd refactor immediately.

Second suggestion can work, but keep in mind you can't have state variables wrapped in useMemo or useCallback. You'd still need to have these values setState.

So a useCallback implementation would still look like a useEffect.

mrice32: Why? Is dynamic data harder to transform?

Don't quite understand why we'd need to refactor if we memoized the component, since all the dynamic data would be passed down from the parent.

I don't think we need state variables at all, right? We'd just pass in the raw tranasactions array like so:
```
const transformed = useMemo(() => transformTransactions(transactions), [transactions]);
```

In that way, the transformation would be cached until a new transactions array is passed in.

src/views/Transactions/Transactions.tsx
```diff
@@ -1,9 +1,44 @@
...
+import formatTransactionData from "./formatTransactionData";
+
 const Transactions = () => {
+  const { isConnected, initOnboard, account } = useTransactionsView();
+
+  // Will take a Transaction Model argument
+  const { ongoingTx, filledTx, headers } = useMemo(
+    () => formatTransactionData(),
+    []
+  );
```
mrice32, line 15: nit: can we move this call down into the TransactionsTable component since the transformation logic is specific to how that table works?

Note: I think we could still grab the (future) `transactions` array from the hook at this level, I just think that array should ideally be passed to the TransactionsTable via props.

mrice32 (approved):
Looks awesome!!

src/views/Transactions/Transactions.tsx
Tulun: The filtering logic would have to be moved into the <Transactions /> view, as the Table below needs to be compatible with both pending and filled transactions.

So yes if filtering is done above, and you pass in the transaction data each table cares about via props.

mrice32: Ah, I see. Makes sense!

mrice32 (approved):

PR #5, opened 2022-03-11
feat(across-v2): refactor sendForm

Gamaranto:


src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -0,0 +1,69 @@
...
+      setAddress(toAddress);
+    }
+  }, [toAddress]);
+
+  const toggle = () => {
+    // modal is closing, reset address to the current toAddress
+    if (!isConnected) return;
+    if (open) setAddress(toAddress || address);
+    setOpen((oldOpen) => !oldOpen);
+  };
```
Tulun, line 37: Nit: I might name this prevOpen, or even prevValue.

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -0,0 +1,69 @@
...
+  // keep the address in sync with the form address
+  useEffect(() => {
+    if (toAddress) {
+      setAddress(toAddress);
+    }
+  }, [toAddress]);
+
+  const toggle = () => {
+    // modal is closing, reset address to the current toAddress
+    if (!isConnected) return;
```
Tulun, line 34: This function seems to do a lot more than just toggle the modal -- maybe it's better these functions were compsed somehow?

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const { fees } = useBridgeFees(amount, fromChain, selectedItem!.symbol);
+  const [inputAmount, setInputAmount] = React.useState<string>(
+    selectedItem && amount.gt("0")
+      ? formatUnits(amount, selectedItem.decimals)
+      : ""
+  );
+
+  useEffect(() => {
+    if (formStatus === "idle") {
+      setInputAmount("");
```
Tulun, line 64: I'd suggest making these strings constants exported from somewhere in case they change (or possible an enum?)

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -0,0 +1,147 @@
...
+        setAmount(adjustedBalance);
+        setInputAmount(formatUnits(adjustedBalance, selectedItem.decimals));
+      } else {
+        setAmount(ethers.constants.Zero);
+        setInputAmount(
+          formatUnits(ethers.constants.Zero, selectedItem.decimals)
+        );
+      }
+    }
+  };
```
Tulun, line 107: Both of these functions might be worth memoizing in useCallback.

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -0,0 +1,147 @@
...
+    error = new InsufficientBalanceError();
+  }
+  const errorMsg = error
+    ? error.message
+    : fees?.isAmountTooLow
+    ? "Bridge fee is high for this amount. Send a larger amount."
+    : fees?.isLiquidityInsufficient
+    ? `Insufficient liquidity for ${selectedItem?.symbol}.`
+    : undefined;
+
```
Tulun, line 126: Really unwieldy ternary here. This would be super awkward to test. I'd suggest creating a function to handle this more cleanly (switch statement, nested if, whatever).

src/components/SendAction/useSendAction.ts
```diff
@@ -0,0 +1,72 @@
...
+      const tx = await approve();
+      console.log(tx);
+    } else {
+      const tx = send();
+      console.log(tx);
+    }
+  };
+
+  const buttonDisabled = status !== "ready";
+
```
Tulun, line 45: Similar to above -- we should avoid magic strings.

src/components/SendAction/useSendAction.ts
```diff
@@ -0,0 +1,72 @@
...
+  let buttonMsg: string = "Send";
+  if (status === "ready") {
+    buttonMsg = hasToApprove ? "Approve" : "Send";
+  } else if (status === "validating") {
+    buttonMsg = "Loading...";
+  } else if (status === "error") {
+    buttonMsg = "Send";
+  }
```
Tulun, line 53: I'd probably extrapolate this to a function.

src/hooks/useBalance.ts
```diff
@@ -11,54 +17,104 @@ import { getBalance, getBalances, balanceQueryKey, balancesQueryKey, ChainId } f
...
+  const accountToQuery = account ?? connectedAccount;
+  const { block: latestBlock } = useBlock(chainId);
+  const blockNumberToQuery = blockNumber ?? latestBlock?.number;
+  // To fetch balances, we need a list of tokens, an account to get balances of, and a specified chainId.
+  const enabledQuery =
+    !!chainIdToQuery &&
+    !!accountToQuery &&
+    tokens.length > 0 &&
+    !!blockNumberToQuery;
+  const queryKey = enabledQuery
```
Tulun, line 88: Any particular reason we need to bit flip these? Are these not falsey?

src/hooks/useSendForm.tsx
```diff
@@ -0,0 +1,362 @@
...
+ * @remarks This hook must be used within the {@link SendFormProvider} component.
+ * @returns Returns a {@link SendFormManagerContext} object
+ */
+export function useSendForm() {
+  const context = useContext(SendFormContext);
+  if (!context) {
+    throw new Error("UseSendForm must be used within a <SendFormProvider>");
+  }
+  return context;
+}
```
Tulun, line 362: I thought the point of react-query was to avoid creating new contexts and global states? Is this not the case?

@daywiss

src/hooks/useSendForm.tsx
```diff
@@ -0,0 +1,362 @@
...
+  setToken: (token: string) => void;
+  setAmount: (amount: ethers.BigNumber) => void;
+  setToAddress: (toAddress: string) => void;
+  setError: (error: ParsingError | undefined) => void;
+};
+
+function useSendFormManager(): SendFormManagerContext {
+  const [state, dispatch] = useReducer(formReducer, initialFormState);
+  const { account: connectedAccount, chainId } = useConnection();
+  console.log(state);
```
mrice32, line 252: Is this supposed to be logged?

mrice32 (approved):
Looks awesome! This is a lot of code to refactor, and I recognize that a lot of it is just moved. Instead of trying to nitpick all of it, I'm happy to leave most as is.

A few things I think we can probably do differently going forward (but would be interested to hear your thoughts:
- This may be controversial, but there are a lot of long inline functions. These seem confusing (and may be inefficient). For some of the longer functions, it could make sense to declare them as standalone functions that take all their dependencies in as parameters and are wrapped in useMemo or useCallback depending on whether a value is needed or you intend to create a callback. This explicitness could be a forcing function to making functions be better encapsulated and have fewer dependencies.
- If the above isn't a desirable way to split logic, it may be better to try and split these hooks up more into smaller sub-hooks.
- Move logic that is pure, synchronous and doesn't impact state back into the view. I do think the hook can aggregate multiple complex pieces of data into a more concise output, but if the hook is just taking a boolean or two and converting it into the text to show in the view, we might be able to do that in the view. Maybe a good rule of thumb is that the hook shouldn't be producing or formatting display values, it should be returning raw numbers/values, objects, and booleans. I think the view/hook line is pretty blurry though, and we might need to think a bit more to determine how to balance them out so one doesn't end up containing too much code.

src/components/AddressSelection/useAddressSelection.ts
Gamaranto: good suggestion, will change it

src/Routes.tsx
```diff
@@ -1,38 +1,33 @@
...
+    location.pathname === "/" &&
     (error instanceof UnsupportedChainIdError ||
       chainId !== sendState.currentlySelectedFromChain.chainId);
   const wrongNetworkPool =
     provider &&
-    (error instanceof UnsupportedChainIdError ||
-      chainId !== DEFAULT_TO_CHAIN_ID);
+    !!chainId &&
+    location.pathname === "/pool" &&
+    (error instanceof UnsupportedChainIdError || chainId !== ChainId.MAINNET);
```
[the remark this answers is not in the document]
Gamaranto: I intend this to change in a next PR so I did not refactor it for now (it can be done though)

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // keep the address in sync with the form address
+  useEffect(() => {
+    if (toAddress) {
+      setAddress(toAddress);
+    }
+  }, [toAddress]);
+
+  const toggle = () => {
+    // modal is closing, reset address to the current toAddress
```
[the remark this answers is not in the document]
Gamaranto: good catch, moved it

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (!isConnected) return;
+    if (open) setAddress(toAddress || address);
+    setOpen((oldOpen) => !oldOpen);
+  };
+  const clearInput = () => {
+    setAddress("");
+  };
+
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
Gamaranto: changed it to `handleAddressChange` to be more explicit

src/hooks/useSendForm.tsx
Gamaranto: No, was just useful when debugging, removed it

src/components/ChainSelection/useChainSelection.ts
```diff
@@ -0,0 +1,54 @@
...
+      init();
+    } else if (wrongNetworkSend) {
+      switchChain(provider, fromChain);
+    }
+  };
+
+  const downshiftState = useSelect({
+    items: CHAINS_SELECTION,
+    defaultSelectedItem: CHAINS_SELECTION.find((x) => x.chainId === fromChain),
+    selectedItem: CHAINS_SELECTION.find((x) => x.chainId === fromChain),
```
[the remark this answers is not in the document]
Gamaranto: I think there are 2 main use cases to wrapping things in a `useMemo`:
1. You need referential equality (useful when doing hooks other people will consume, as a `useState` value also gives referential equality)
2. You want to memoize a result, to avoid running the computation again, prob everything that has at least linear cost goes here

Note that when doing 2, `useMemo` is not free as it runs a shallow comparison between deps, so for simple computations, you can make them more expensive with useMemo. I'd say we should not consider it unless we can see a notable difference in performance, and then go back and add it

Gamaranto:
> Looks awesome! This is a lot of code to refactor, and I recognize that a lot of it is just moved. Instead of trying to nitpick all of it, I'm happy to leave most as is just so we can move forward without getting bogged down in rewriting all of this logic.
> 
> A few things I think we can probably do differently going forward (but would be interested to hear your thoughts:
> 
> * This may be controversial, but there are a lot of long inline functions. These seem confusing (and may be inefficient). For some of the longer functions, it could make sense to declare them as standalone functions that take all their dependencies in as parameters and are wrapped in useMemo or useCallback depending on whether a value is needed or you intend to create a callback. This explicitness could be a forcing function to making functions be better encapsulated and have fewer dependencies.
> * If the above isn't a desirable way to split logic, it may be better to try and split these hooks up more into smaller sub-hooks.
> * Move logic that is pure, synchronous and doesn't impact state back into the view. I do think the hook can aggregate multiple complex pieces of data into a more concise output, but if the hook is just taking a boolean or two and converting it into the text to show in the view, we might be able to do that in the view. Maybe a good rule of thumb is that the hook shouldn't be producing or formatting display values, it should be returning raw numbers/values, objects, and booleans. I think the view/hook line is pretty blurry though, and we might need to think a bit more to determine how to balance them out so one doesn't end up containing too much code.

I agree fully with 2, 3. not really sure about what you mean in 1. What is one long inline function you were thinking of? just so I understand better

src/hooks/useBalance.ts
Gamaranto: iirc I added the `!!` so that the end value will always be a boolean, otherwise the type is `boolean | number`, but yes the check would work correctly either way

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -0,0 +1,147 @@
...
+      : ""
+  );
+
+  useEffect(() => {
+    if (formStatus === "idle") {
+      setInputAmount("");
+    }
+  }, [formStatus]);
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
Gamaranto: Good callout, changed it to `handleInputChange`

src/components/CoinSelection/useCoinSelection.tsx
Gamaranto: Good catch, I'm going to remove / refactor it in a next PR (I will lift the state of the SendAction up, so we can share the errors instead of re-deriving them), but you're 100% right and the next iteration won't have it like that, I'll strive to avoid complex ternaries and instead use functions

src/components/CoinSelection/useCoinSelection.tsx
Gamaranto: sounds good, will change

src/utils/bridge.ts
```diff
@@ -233,3 +239,256 @@ export const optimismErc20Pairs = () => {
...
   return getTokenPairMapping(ChainId.MAINNET, ChainId.BOBA, ["WETH", "ETH"]);
 };
+
+/**
+ * Returns the list of tokens that can be sent from chain A to chain B, by computing their tokenList intersection and taking care of additional chain specific quirks.
+ * @param chainA - the chain to bridge from, that is, the chain that tokens are sent from.
+ * @param chainB - the destination chain, that is, where tokens will be sent.
+ * @returns Returns a list of tokens that can be sent from chain A to chain B.
+ */
+export function filterTokensByDestinationChain(
```
daywiss, line 249: not a huge fan of this logic, hard to reason about, would much rather prefer a static data structure to look up in, but not a blocker

src/hooks/useSendForm.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+export type FormStatus = "idle" | "touched" | "valid" | "error";
+
+type FormState = {
+  status: FormStatus;
+  token: string;
+  amount: ethers.BigNumber;
+  toChain: ChainId;
+  fromChain: ChainId;
+  toAddress?: string;
```
daywiss, line 30: should this be required? if not explicitly provided, it should default to the users address right?

daywiss (approved):
approving because i dont think i have any blockers

src/components/ChainSelection/useChainSelection.ts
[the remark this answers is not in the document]
Gamaranto: Agree, we could just use the ChainIds, and then do the look up in the `CHAINS` object in the view

src/utils/bridge.ts
```diff
@@ -233,3 +239,256 @@ export const optimismErc20Pairs = () => {
...
+
+    case ChainId.BOBA:
+      return { approve: sendBobaApproval, deposit: sendBobaDeposit };
+
+    case ChainId.ARBITRUM:
+      return { approve: sendArbitrumApproval, deposit: sendArbitrumDeposit };
+    default:
+      throw new Error(`Unsupported bridge ${toChain}`);
+  }
+}
```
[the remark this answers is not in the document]
Gamaranto: Agree, I think the `BaseDeposit` type is a good base type to expand from, but the `AcrossDeposit` def might not be, not really sure how to handle a more generalized case so open to suggestions

src/hooks/useSendForm.tsx
```diff
@@ -0,0 +1,362 @@
...
+    case "valid":
+      return { ...state, status: "error", error };
+    case "error":
+      return state;
+    default:
+      throw new Error("unreachable");
+  }
+}
+
+function formReducer(state: FormState, action: Action) {
```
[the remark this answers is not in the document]
Gamaranto: Yes, I tried to keep it simple initially (just one reducer), but I found it really hard to reason about, and found separating the events helped

Gamaranto:
Regarding doing the form with RQ, the way to do it would be to set up reducers and then invalidate the cache for the query keys used, so it might look like this:
```javascript
 const setAmount = (amount) => {
    //  fetch the current form state
    const formState = queryClient.getQueryData(["form"]);
    // do some validation...
    // now invalidate the old form state
   client.invalidateQueries(["form"])
   // set the new form state 
   const newFormState = {...formState, amount, ...otherFields}
   client.setQueryData(["form"], newFormState)
```

The annoying thing is that when reading this state, the function passed to the `useQuery` hook must be async, this would be work:
```javascript
// this would be ok I think, the fn does nothing since the query cache is prefilled / invalidated manually anyway 
const data = useQuery(["form"], async () => {
  return initialFormState;
}, {staleTime: Infinity})
```

But would be hard to reason about.
Regarding context, I want to move all of this state up and share it directly as props to all the elements of the form, so I should be able to get rid of it, but for now I think it is the right tool

src/utils/bridge.ts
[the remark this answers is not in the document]
daywiss: yea dont have a solution. naively i would join the params for each type so the caller could provide either, then error at runtime if they provided the wrong kind

src/hooks/useSendForm.tsx
daywiss: good point

src/utils/bridge.ts
daywiss: yea what i had proposed was a nested object: `Record<chainId,Record<chainId,Config>>`  All posible from/to configurations are explicitly enumerated. The config contains config for that pair, so it would contain available tokens you can select from.   Any to/from not included in the object would not be selectable in the form.

daywiss:
> Regarding doing the form with RQ, the way to do it would be to set up reducers and then invalidate the cache for the query keys used, so it might look like this:

would require more thought, but a lot of logic could exist outside of react, the only thing its doing is taking function calls and values and reducing them to an object.  That doesnt change the fact u still need some kind of global store though, but perhaps if you thought of it like an api, you could use react query to make specific calls, passing data in and getting the output of the form every call thats made.  this is just spitballing though.

Tulun (approved):
I think my biggest concern is a larger conversation outside the scope of this PR -- That ultimately there is a disconnect between global state here because there is *no* global state anymore.
And the solution here would just extend us to potentially have dozens of contexts to solve this problem if there is more state that cannot live in RQ.

We've divided out the state into buckets and we just hope they aren't interconnected. But this seems to defeat the purpose of shifting into RQ in the first place. 

@mrice32 @daywiss

PR #6, opened 2022-03-11
feat: Create TransactionModel and convert data to Rows & cells

Tulun:
1) Create the Transaction model that will be formatted from the raw event data.
2) Use that Model to create the rows and cells for Table.

https://app.shortcut.com/uma-project/story/4863/create-interface-and-convert-function-for-transactions

### DIFF
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -3,7 +3,7 @@ import { Wrapper, Title, ConnectButton, Account } from "./Transactions.styles";
 import useTransactionsView from "./useTransactionsView";
 import TransactionsTable from "./TransactionsTable";
 import { shortenAddress } from "utils/format";
-import formatTransactionData from "./formatTransactionData";
+import createTransactionTableJSX from "./createTransactionTableJSX";
 
 const Transactions = () => {
   const { isConnected, initOnboard, account, transactions } =
@@ -11,7 +11,7 @@ const Transactions = () => {
 
   // Will take a Transaction Model argument
   const { ongoingTx, filledTx, headers } = useMemo(
-    () => formatTransactionData(transactions),
+    () => createTransactionTableJSX(transactions),
     [transactions]
   );
 
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -6,12 +6,13 @@ import { shortenAddress } from "utils/format";
 import formatTransactionData from "./formatTransactionData";
 
 const Transactions = () => {
-  const { isConnected, initOnboard, account } = useTransactionsView();
+  const { isConnected, initOnboard, account, transactions } =
+    useTransactionsView();
 
   // Will take a Transaction Model argument
   const { ongoingTx, filledTx, headers } = useMemo(
-    () => formatTransactionData(),
-    []
+    () => formatTransactionData(transactions),
+    [transactions]
   );
 
   return (
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -4,14 +4,20 @@ import useTransactionsView from "./useTransactionsView";
 import TransactionsTable from "./TransactionsTable";
 import { shortenAddress } from "utils/format";
 import createTransactionTableJSX from "./createTransactionTableJSX";
+import { headers } from "./createTransactionTableJSX";
 
 const Transactions = () => {
   const { isConnected, initOnboard, account, transactions } =
     useTransactionsView();
 
-  // Will take a Transaction Model argument
-  const { ongoingTx, filledTx, headers } = useMemo(
-    () => createTransactionTableJSX(transactions),
+  const ongoingTx = useMemo(
+    () => createTransactionTableJSX(transactions.filter((x) => x.filled < 100)),
+    [transactions]
+  );
+
+  const filledTx = useMemo(
+    () =>
+      createTransactionTableJSX(transactions.filter((x) => x.filled >= 100)),
     [transactions]
   );
 
--- a/src/views/Transactions/createTransactionModel.ts
+++ b/src/views/Transactions/createTransactionModel.ts
@@ -6,7 +6,7 @@ export interface Transaction {
   filled: number;
   fromChain: ChainId;
   toChain: ChainId;
-  symbol: string;
+  assetContractAddress: string;
   amount: ethers.BigNumber;
   txHash: string;
 }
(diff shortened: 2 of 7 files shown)


src/views/Transactions/formatTransactionData.tsx
```diff
@@ -23,241 +23,93 @@ export default function formatTransactionsData() {
...
-              )}
-            </TableLink>
-          ),
-        },
-      ],
-    },
-  ];
+function formatRows(transactions: Transaction[]) {
+  const rows: IRow[] = [];
+  transactions.forEach((tx) => {
```
daywiss, line 28: why do u use `forEach` here rather than `.map`?

src/views/Transactions/formatTransactionData.tsx
```diff
@@ -1,20 +1,20 @@
...
 // Stub of function.
 // Will take View Model Transaction as arg
-export default function formatTransactionsData() {
-  const rows = formatRows();
-  const filledTx = rows.filter((x, i) => i !== 0);
-  const ongoingTx = rows.filter((x, i) => i === 0);
+export default function formatTransactionsData(transactions: Transaction[]) {
+  const rows = formatRows(transactions);
+  const filledTx = rows.filter((x, i) => x.cells[1].value === "Filled");
+  const ongoingTx = rows.filter((x, i) => x.cells[1].value === "Pending");
```
daywiss, line 17: personally i would invert this logic. pull these filters out to a higher level, and have this format function just take in an array of txs to return the table jsx.

src/views/Transactions/Transactions.tsx
```diff
@@ -6,12 +6,13 @@ import { shortenAddress } from "utils/format";
...
-  const { isConnected, initOnboard, account } = useTransactionsView();
+  const { isConnected, initOnboard, account, transactions } =
+    useTransactionsView();
 
   // Will take a Transaction Model argument
   const { ongoingTx, filledTx, headers } = useMemo(
-    () => formatTransactionData(),
-    []
+    () => formatTransactionData(transactions),
+    [transactions]
```
daywiss, line 15: I dont think this memo does anything other than add more overhead. Assuming you get a transaction array from some client, any time theres a change to any data inside that array of transactions, there will be a new array. The new array will cause this function to be re-called and the entire table to re-render. This will happen with or without the memo function here, so in general prefer to remove code and overhead than add it.

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,15 +1,18 @@
 import { useConnection } from "state/hooks";
 import { onboard } from "utils";
+import createTransactionModel from "./createTransactionModel";
 
 export default function useTransactionsView() {
   const { provider, chainId, isConnected, account } = useConnection();
   const { init } = onboard;
+  const transactions = createTransactionModel();
```
daywiss, line 8: i would use this hook to filter the different types of transactions and return `filledTransactions` and `ongoingTransactions`. You can then use `formatTransactionData` on those arrays. 

about the memoization you have, i dont actually think you need it, but it would be preferred to keep it in here. The reason i think why its not here, is because it requires an input parameter to memoize, and you dont have any here. anyways see my comment on why i think its not needed.

src/views/Transactions/Transactions.tsx
Tulun: If I remove useMemo, this array would be recreated on every re-render.

src/views/Transactions/formatTransactionData.tsx
Tulun: It can be done either way.

src/views/Transactions/formatTransactionData.tsx
amateima: just a comment, no need to change here. I'd rename this function to smth more specific like getTransactionsJSX or smth. `formatData` sounds a little bit vague, it can be formatted in many ways

amateima (approved):
From my perspective it can be merged after other comments are addressed. 
One very small nit from me: inside JSX map callbacks, I'd replace curly braces with parenthesis, so you can delete the `return` keyword and reduce the number of indentation levels

src/views/Transactions/formatTransactionData.tsx
amateima: +1 for David's comment

src/views/Transactions/createTransactionModel.ts
```diff
@@ -0,0 +1,48 @@
...
+  filled: number;
+  fromChain: ChainId;
+  toChain: ChainId;
+  assetContractAddress: string;
+  amount: ethers.BigNumber;
+  txHash: string;
+}
+
+// Will take a raw TX argument and transform into ViewModel.
+export default function createTransactionModel(): Transaction[] {
```
mrice32, line 15: nit: I'm confused at the comment and this function.

What does raw TX mean? And I don't really understand what a ViewModel means.

In my mind, there are three layers:
1. Pure functions (can be async!) that contain business logic.
2. Hooks to call, cache, and orchestrate business logic functions.
3. Views that take the output of a primary hook and their props to render JSX.

What you have now is in-line with this pattern outside of a few small details:
1. createTransactionsModel seems like an odd name for a function that's just grabbing the list of bridge transactions from the blockchain? Would think something more like getTransactions might be better suited for that.
2. The way this stub is being called now is pretty different from how it will work in the future. What you have is definitely the right way to go, but just want to make sure we're aligned there. This will be an async business logic function that's called via a useEffect or react-query and will have its result cached. It will be responsible for grabbing a few different events from different chains and piecing them together to get a normalized list of bridging transactions. The only reason I mention this is that your comment suggests that there will be a createTransactionModel function that transforms some "raw Tx argument". I don't really think that's the case.

Once again, no changes required, the comment just confused me and I wanted to make sure we were on the same page!

src/views/Transactions/formatTransactionData.tsx
```diff
@@ -1,20 +1,20 @@
...
 import { shortenTransactionHash } from "utils/format";
 import { ICell, IRow } from "components/Table/Table";
-
-import arbLogo from "assets/arbitrum-logo.svg";
-import umaLogo from "assets/UMA-round.svg";
-import ethLogo from "assets/ethereum-logo.svg";
+import { Transaction } from "./createTransactionModel";
+import { CHAINS, TOKENS_LIST } from "utils/constants";
 
 // Stub of function.
```
mrice32, line 12: I think we can probably remove this comment since this is no longer a stub.

src/views/Transactions/formatTransactionData.tsx
mrice32: nit: I think maps are preferred when each input element is used to create a new element in an output array (as is the case here).

src/views/Transactions/useTransactionsView.ts
mrice32: Either way, isn't it specifically tied to this view? Doing the filtering at a higher level allows the lower level functions to be generic. Doing it at a lower level (as is done now) implicitly makes everything above the filtering specific to that type of filtering.

If you want this hook and the createTransactionModel function to both be generic, then I think the only thing you can do is put the filtering in the view or in the table component.

mrice32 (commented):
Looks good! A few comments.

src/views/Transactions/useTransactionsView.ts
daywiss: yea i guess this is a difference in philosophy around what the "one hook" is actually for. In my mind the hook is specific to the view component its supporting, and i didnt think it was meant to be reused in other components, since they would design their own view specific hooks.  You could argue that these hooks are meant to be used more generically, but i would need some convincing.  thoughts? @amateima @Gamaranto @mrice32 @Tulun

src/views/Transactions/createTransactionModel.ts
Tulun: It's just a comment to say this is a stub.

The "raw tx" in this case will be whatever is returned from react-query and the normalized for the ViewModel.

The model is the raw data -- IE in this case event data or whatever else -- but we aren't consuming that at this point.

src/views/Transactions/formatTransactionData.tsx
Tulun: Fixed

src/views/Transactions/useTransactionsView.ts
Tulun: I think it largely depends how we want to write these functions. Probably somewhere down the line something is more generic, though, as you could see this model being reused.

src/views/Transactions/formatTransactionData.tsx
Tulun: Fixed.

src/views/Transactions/formatTransactionData.tsx
Tulun: Fixed.

src/views/Transactions/createTransactionTableJSX.tsx
```diff
@@ -0,0 +1,148 @@
...
+          target="_blank"
+          rel="noreferrer"
+        >
+          {shortenTransactionHash(tx.txHash)}
+        </TableLink>
+      ),
+    };
+    row.cells.push(txHash);
+
+    return row;
```
daywiss, line 103: this is just a small thing but you can make this slightly more concise and readable imo:
```js
function formatTransactionRows(transactions: Transaction[]): IRow[] {
 // delete const row initialized at top of map
 ...
 return {
   cells : [timestamp, status, filled, fromChain, toChain, symbol, amount, txHash]
 }
}
```

src/views/Transactions/Transactions.tsx
daywiss: i see, i guess i was assuming the majority of renders would be coming from the transaction array itself changing. but ok that could justify it

src/views/Transactions/useTransactionsView.ts
daywiss: i think the generic thing in this case is the transaction client you are using, which could be re-used in other hooks. theres not a big need to make this particular hook generic, though we can leave that option open

daywiss (approved):

mrice32 (approved):
LGTM!

PR #7, opened 2022-03-14
feat: Sidebar menu

Tulun:
Sidebar feature.
https://app.shortcut.com/uma-project/story/4750/create-view-for-hamburger-menu

Screen Shot 2022-03-14 at 4 29 53 PM
Screen Shot 2022-03-14 at 4 29 46 PM

Note: Jesper asked me to change the sidebar slightly, so I've adjusted it.

Requested changes:
not connected
connected-tweak


src/components/Sidebar/Sidebar.styles.tsx
```diff
@@ -0,0 +1,111 @@
+import { ProSidebar, SidebarHeader, Menu, MenuItem } from "react-pro-sidebar";
+
+import styled from "@emotion/styled";
+import "react-pro-sidebar/dist/css/styles.css";
+import { SecondaryButton } from "../Buttons";
+
+export const Overlay = styled.div`
+  z-index: 500;
```
Gamaranto, line 8: Any reason for 500 here? I think we should use always use the lowest value possible (like `1` or `2`) to make it easy to see what goes on top of what

src/components/Sidebar/Sidebar.styles.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const StyledSidebar = styled(ProSidebar)`
+  position: absolute;
+  transition: all 0.5s ease-in-out;
+  right: -450px;
+  top: 0;
+  width: 450px;
+  transform: translateX(50px);
+  &.open {
+    transform: translateX(-450px);
```
Gamaranto, line 37: this could be changed to `translateX(-100%)` so that it will always follow the width of the sidebar if we were to change it in the future

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,130 @@
...
+const { init, reset } = onboard;
+
+interface Props {
+  openSidebar: boolean;
+  setOpenSidebar: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+const Sidebar: FC<Props> = ({ openSidebar, setOpenSidebar }) => {
+  const { account, isConnected, chainId } = useConnection();
+  const location = useLocation();
```
Gamaranto, line 26: let's group those two into one hook like `useSidebar`

src/utils/constants.ts
```diff
@@ -36,7 +36,12 @@ export const COLORS = {
     150: "0deg 0% 96%",
     200: "220deg 2% 72%",
     300: "240deg 4% 27%",
+    // #2d2e33
     500: "230deg 6% 19%",
+    // #4d4c53
+    600: "249deg 4% 31%",
```
Gamaranto, line 42: I think using 600 and 700 here is misleading, as they break the scale assumption (from lighter colors to darker ones), we should order this scale based on the last number of the hsl, in descending order (so 100 is lightest, 900 darkest)

src/components/Sidebar/Sidebar.styles.tsx
amateima: I don't have a problem with using high numbers, but if we use `z-index` in multiple places (overlays, popups, tooltips, notifications, etc) we can specify these values somewhere in the readme or in a css comment

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,130 @@
...
+          )}
+          <ConnectButton
+            onClick={() => {
+              if (isConnected) {
+                return reset();
+              } else {
+                return init();
+              }
+            }}
+          >
```
amateima, line 49:
```suggestion
          <ConnectButton onClick={() => isConnected ? reset() : init() }>
```
The ternary operator would save some lines

amateima (approved):

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,130 @@
...
+              target="_blank"
+              rel="noreferrer"
+              onClick={() => setOpenSidebar(false)}
+            >
+              Support (Discord)
+            </a>
+          </StyledMenuItem>
+          <StyledMenuItem>
+            <a
+              href="https://github.com/across-protocol/frontend-v2/"
```
daywiss, line 96: you might just want to link to the across-protocol org

daywiss (commented):
lgtm outside of some nits. I agree with the other reviewers, so make sure to address the outstanding issues.  Also im not sure if this was meant to have a drop shadow but currently it looks a bit flat, otherwise works well.

src/utils/constants.ts
Tulun: Fair enough

src/components/Sidebar/Sidebar.tsx
Tulun: Good call!

src/components/Sidebar/Sidebar.styles.tsx
```diff
@@ -0,0 +1,111 @@
...
+  right: 0;
+  bottom: 0;
+  background-color: var(--color-gray-lighter);
+  display: -webkit-box;
+  display: -webkit-flex;
+  display: -ms-flexbox;
+  display: flex;
+  -webkit-box-pack: center;
+  -ms-flex-pack: center;
+  -webkit-justify-content: center;
```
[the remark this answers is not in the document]
Tulun: Fixed.

src/components/Sidebar/Sidebar.styles.tsx
Tulun: Fixed!

src/components/Sidebar/Sidebar.tsx
Tulun: Fixed.

Tulun:
> lgtm outside of some nits. I agree with the other reviewers, so make sure to address the outstanding issues. Also im not sure if this was meant to have a drop shadow but currently it looks a bit flat, otherwise works well.

There was an issue with the overlay, should be fixed.

src/components/Header/Header.tsx
```diff
@@ -31,7 +31,11 @@ const MOBILE_ONLY_LINKS = [
...
   { href: "https://github.com/across-protocol", name: "Github" },
 ];
-const Header: React.FC = () => {
+
+interface Props {
+  setOpenSidebar: React.Dispatch<React.SetStateAction<boolean>>;
+}
+const Header: React.FC<Props> = ({ setOpenSidebar }) => {
   const location = useLocation();
   const [isMenuOpen, setMenuOpen] = useState(false);
```
mrice32, line 40: Do we still need tis menu state, logic, and useEffect? Or does the sidebar logic you added handle all of that now?

daywiss (approved):

PR #8, opened 2022-03-16
fix(across-v2): refactor confirmation screen

Gamaranto:
This PR makes the confirmation screen work with the new state management of the sendForm


src/views/Confirmation/Confirmation.tsx
```diff
@@ -130,15 +142,13 @@ const Confirmation: React.FC = () => {
...
                     }`}
                     target="_blank"
                     rel="noopener noreferrer"
                   >
-                    <span>{deposit.toAddress}</span>
-                    <span>
-                      {shortenAddress(deposit.toAddress ?? "", "...", 10)}
-                    </span>
+                    <span>{deposit.to}</span>
+                    <span>{shortenAddress(deposit.to ?? "", "...", 10)}</span>
```
daywiss, line 151: this probably should be hidden if deposit.to doesnt exist, but you might want to double check the code in `shortenAddress` to see what happens if u provide empty string

src/utils/constants.ts
```diff
@@ -746,3 +746,4 @@ export const MAX_APPROVAL_AMOUNT = ethers.constants.MaxUint256;
 // FIXME: https://github.com/across-protocol/frontend-v2/pull/2#discussion_r822100493
 export const ETH_ADDRESS = ethers.constants.AddressZero;
 export const FEE_ESTIMATION = ".004";
+export const CONFIRMATIONS = 1;
```
daywiss, line 749: something got lost here, i thought this was an env in v1

daywiss (commented):
looks good for the most part, just some minor comments, also:
1. run prettier
2. screenshots would be nice since we have to pay to see this

src/views/Confirmation/Confirmation.tsx
Gamaranto: I think `deposit.to` will always be defined, isn't it the address the funds were sent to?

src/utils/constants.ts
Gamaranto: Sorry, will change it to be an env again

src/components/SendAction/useSendAction.ts
```diff
@@ -32,14 +35,35 @@ export default function useSendAction() {
...
     if (hasToApprove) {
       const tx = await approve();
-      console.log(tx);
+      return tx;
     } else {
-      const tx = send();
-      console.log(tx);
+      // We save the fees here, in case they change between here and when we save the deposit.
+      const feesUsed = fees;
+      const tx = await send();
```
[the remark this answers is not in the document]
Gamaranto: Really good call out, what I had in mind was invalidating all the queries related to the transactions page (comment on line `62`) so that they get refetched immediately, there we could also prefill the cache with the tx we have synchronously

src/components/SendAction/useSendAction.ts
```diff
@@ -32,14 +35,35 @@ export default function useSendAction() {
...
+      // NOTE: This check is redundant, as if `status` is `ready`, all of those are defined.
+      if (tx && toAddress && account && feesUsed) {
+        tx.wait(CONFIRMATIONS).then(tx => {
+          onDeposit({
```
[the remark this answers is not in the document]
Gamaranto: Agree, will change it

src/components/SendAction/useSendAction.ts
```diff
@@ -32,14 +35,35 @@ export default function useSendAction() {
...
+            txHash: tx.transactionHash,
+            amount,
+            token,
+            fromChain,
+            toChain,
+            to: toAddress,
+            from: account,
+            fees: feesUsed,
+          })
+        });
```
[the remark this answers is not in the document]
Gamaranto: good call, not sure, I guess we could log the error for now

src/views/Confirmation/Confirmation.tsx
daywiss: oh ok, wel it just looks like the code expects it to not exist sometimes

src/components/SendAction/useSendAction.ts
```diff
@@ -32,14 +35,35 @@ export default function useSendAction() {
   }, [amount, fees, fromChain]);
 
   const handleActionClick = async () => {
+    if (status !== "ready") {
+      return;
+    }
     if (hasToApprove) {
       const tx = await approve();
-      console.log(tx);
+      return tx;
```
mrice32, line 43: Why do we return tx if this return value is ignored by the onClick caller?

src/components/SendAction/SendAction.tsx
```diff
@@ -12,8 +12,12 @@ import {
 
 import InformationDialog from "components/InformationDialog";
 import useSendAction from "./useSendAction";
+import type { Deposit } from "views/Confirmation";
 
-const SendAction: React.FC = () => {
+type Props = {
+  onDeposit: (deposit: Deposit) => void;
+};
+const SendAction: React.FC<Props> = ({ onDeposit }) => {
```
Tulun, line 20: Prefer interfaces to types for objects.

src/components/SendAction/useSendAction.ts
```diff
@@ -32,14 +35,35 @@ export default function useSendAction() {
   }, [amount, fees, fromChain]);
 
   const handleActionClick = async () => {
+    if (status !== "ready") {
+      return;
+    }
     if (hasToApprove) {
```
Tulun, line 41: What does the user see if the button is clickable but it is not "ready"? Is it disabled and this just a catch?

mrice32 (commented):
Looks good! A few comments

src/components/SendAction/useSendAction.ts
```diff
@@ -13,16 +15,17 @@ type TokenInfo =
   }
   | Token;
 
-export default function useSendAction() {
+export default function useSendAction(onDeposit: (deposit: Deposit) => void) {
```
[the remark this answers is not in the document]
Tulun: +1. I've always found SendAction to be very unclear.

src/components/SendAction/SendAction.tsx
Gamaranto: why?

src/views/Send.tsx
```diff
@@ -1,24 +1,37 @@
...
-  CoinSelection,
-  AddressSelection,
-  SendAction,
-} from "components";
-import { SendFormProvider } from "hooks";
+import React, { useState } from "react";
+import SendForm from "components/SendForm";
+import Confirmation, { Deposit } from "./Confirmation";
+
+function useSend() {
```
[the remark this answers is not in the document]
Gamaranto: yes we can do that, I left it here because it was small, but will move it

src/components/SendAction/useSendAction.ts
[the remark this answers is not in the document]
Gamaranto: Naming things is hard 😆  
- `useSend` is meant to be the hook for the `Send` component
- `useSendAction` is meant to be the hook for the `SendAction` component

What should I rename those components to? ( I think we should keep the convention `Component` and `useComponent` for the component hooks)
I like `useSendTab` for the current `useSend` (we should also prob rename it to BridgeTab to match the headers) but need some ideas for the replacement of `useSendAction`

src/components/SendAction/useSendAction.ts
Gamaranto: Mh, I thought that we might use it in some other way like saving the transaction etc... but will remove for now

src/views/Send.tsx
```diff
@@ -1,24 +1,37 @@
...
+    setDeposit(deposit);
+    setShow(true);
+  };
+  return {
+    showConfirmationScreen,
+    deposit,
+    onDeposit,
+    onCloseConfirmationScreen,
+  };
+}
```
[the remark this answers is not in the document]
Gamaranto: will do

src/components/SendAction/useSendAction.ts
Gamaranto: yes, it would be disabled as `status` is not `ready`, but thought to add this for additional safety

src/views/Confirmation/Confirmation.tsx
Gamaranto: Removed the defaulting, the types were wrong, but we only save deposits after sending a transaction, so this will always be defined

src/components/SendAction/useSendAction.ts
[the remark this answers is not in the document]
daywiss: after some thought on this, i think its probably ok to not worry about it until we knwo more about how its going to work. potentially we have a client which has a function where you can save this data, or we may not even care about it at all

daywiss (approved):

src/components/SendAction/SendAction.tsx
Tulun: "If you would like a heuristic, use interface until you need to use features from type."

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

src/components/SendAction/useSendAction.ts
Tulun: Got it. I might note this in a comment.

mrice32 (approved):

PR #9, opened 2022-03-16
fix: CSS fixes from design QA; remove wallet modal

Tulun:
1) Wallet info opens hamburger menu as well
2) Minor fixes for 13.3" laptop (1280 x 800) with overflow-x table
3) Minor CSS fixes with background and box shadow.

Fixes from feedback:
Skärmavbild 2022-03-16 kl  11 02 36

Skärmavbild 2022-03-16 kl  14 52 21

Skärmavbild 2022-03-16 kl  14 51 25


daywiss (approved):
lgtm

daywiss:
just wanted to mention, the table scrolling-x should probably not scroll the header as well

amateima (approved):

src/components/Wallet/Wallet.tsx
```diff
@@ -59,28 +43,16 @@ const Wallet: FC = () => {
...
-            Disconnect
-          </WalletModalDisconnect>
-        </WalletModal>
-      )}
-    </div>
+    <Wrapper onClick={() => setOpenSidebar(true)}>
+      <Info>
+        <div>
+          {formatEther(balance ?? "0")}{" "}
+          {CHAINS[chainId ?? 1].nativeCurrency.symbol}
```
Gamaranto, line 50: nit: I remember in another PR someone (i think @mrice32 or @daywiss ) pointed out to avoid defaulting if possible, there are no issues here (as the chainId will be defined if the wallet shows up), but I think we should do something like adding a check:
```javascript
if(!chainId) {
    return null
   }
   return (
     // jsx here...
     )
   
   
   ```

File: src/components/Wallet/Wallet.tsx [hunk 2]
```diff
@@ -14,32 +14,16 @@ import {
...
-  useEffect(() => {
-    if (!isConnected && isOpen) setIsOpen(false);
-  }, [isConnected, isOpen]);
-
-  const disconnectWallet = () => {
-    setIsOpen(false);
-    reset();
-  };
 
   const { data: balance } = useETHBalance(
```
Gamaranto, line 28: we can use the new react-query hooks here, in this case it would be the `useBalance` hook
And it would become:
```javascript
import {useBalance} from "hooks";
// inside the component here
const {balance} = useBalance(ETH_ADDRESS, chainId, account);
```

Gamaranto (commented):
just 2 small comments, otherwise lgtm

src/components/Wallet/Wallet.tsx
Tulun: I think this is outside the scope of the PR. This was just deleting the view in Wallet -- this largely just CSS fixes, we can do this later.

PR #10, opened 2022-03-18
feat: Hamburger mobile

Tulun:
Mobile for hamburger menu.

Screen Shot 2022-03-18 at 11 16 55 AM


daywiss (commented):
somethings not right here, i check in the mobile view and wasnt able to close the menu

EDIT: i tried again and could see the X, it was somehow missing the first time i tested

PR #11, opened 2022-03-18
fix(across-v2): make error banner work with new logic

Gamaranto:
This PR refactors the error banner at the top of the app to work with the new state management logic.
Note: 
I tried a different implementation with React Portals where you could render the `SuperHeader` banner inside the `Send` view (in the React tree) but render on top of the header in the real html tree. 
I think it had a nicer API:
```javascript
function SomeSendComp() {
   const {fromChain} = useSendForm();
   return (
     <>
      <SuperHeader>{logic to show errors}</SuperHeader>
      <OtherComp>...
      )
```

But it was really slow, so I abandoned that approach for this one, were we sync the banner with `useEffects` , I don't love this solution, so open to ideas on how to improve on this

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -1,10 +1,10 @@
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { Switch, Route, useLocation } from "react-router-dom";
 import { Send, Pool, About, Transactions } from "views";
 import { Header, SuperHeader } from "components";
 import { useConnection } from "state/hooks";
-import { CHAINS, switchChain, WrongNetworkError } from "utils";
-import { useError, usePrevious } from "hooks";
+import { WrongNetworkError } from "utils";
+import { useError } from "hooks";
 import styled from "@emotion/styled";
 import Sidebar from "components/Sidebar";
 
@@ -14,16 +14,6 @@ function useRoutes() {
   const location = useLocation();
   const { error, removeError } = useError();
 
-  // reset wrong chain error on route change
-  const prevLocation = usePrevious(location);
-  useEffect(() => {
-    if (
-      error instanceof WrongNetworkError &&
-      prevLocation.pathname !== location.pathname
-    ) {
-      removeError();
-    }
-  }, [removeError, error, location.pathname, prevLocation.pathname]);
   return {
     openSidebar,
     setOpenSidebar,
@@ -35,25 +25,14 @@ function useRoutes() {
 }
 // Need this component for useLocation hook
 const Routes: React.FC = () => {
-  const { openSidebar, setOpenSidebar, error, removeError, provider } =
-    useRoutes();
-  const showError = provider && error instanceof WrongNetworkError;
+  const { openSidebar, setOpenSidebar, error, removeError } = useRoutes();
+
   return (
     <>
       {error && !(error instanceof WrongNetworkError) && (
         <SuperHeader>
           <div>{error.message}</div>
           <RemoveErrorSpan onClick={() => removeError()}>X</RemoveErrorSpan>
-        </SuperHeader>
-      )}
-      {showError && (
-        <SuperHeader>
-          <div>
-            You are on an incorrect network. Please{" "}
-            <button onClick={() => switchChain(provider, error.correctChainId)}>
-              switch to {CHAINS[error.correctChainId].name}
-            </button>
-          </div>
         </SuperHeader>
       )}
       <Header openSidebar={openSidebar} setOpenSidebar={setOpenSidebar} />
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -1,68 +1,57 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { Switch, Route, useLocation } from "react-router-dom";
 import { Send, Pool, About, Transactions } from "views";
 import { Header, SuperHeader } from "components";
 import { useConnection } from "state/hooks";
-import { CHAINS, UnsupportedChainIdError, switchChain, ChainId } from "utils";
-
-import { useAppSelector } from "state/hooks";
-import { useError } from "hooks";
+import { CHAINS, switchChain, WrongNetworkError } from "utils";
+import { useError, usePrevious } from "hooks";
 import styled from "@emotion/styled";
 import Sidebar from "components/Sidebar";
 
+function useRoutes() {
+  const [openSidebar, setOpenSidebar] = useState(false);
+  const { provider } = useConnection();
+  const location = useLocation();
+  const { error, removeError } = useError();
+
+  // reset wrong chain error on route change
+  const prevLocation = usePrevious(location);
+  useEffect(() => {
+    if (
+      error instanceof WrongNetworkError &&
+      prevLocation.pathname !== location.pathname
+    ) {
+      removeError();
+    }
+  }, [removeError, error, location.pathname, prevLocation.pathname]);
+  return {
+    openSidebar,
+    setOpenSidebar,
+    provider,
+    error,
+    removeError,
+    location,
+  };
+}
 // Need this component for useLocation hook
 const Routes: React.FC = () => {
-  const [openSidebar, setOpenSidebar] = useState(false);
-  const { provider, chainId, error } = useConnection();
-  const location = useLocation();
-  const sendState = useAppSelector((state) => state.send);
-  const { error: globalError, removeError } = useError();
-
-  const wrongNetworkSend =
-    provider &&
-    chainId &&
-    location.pathname === "/" &&
-    (error instanceof UnsupportedChainIdError ||
-      chainId !== sendState.currentlySelectedFromChain.chainId);
-  const wrongNetworkPool =
-    provider &&
-    !!chainId &&
-    location.pathname === "/pool" &&
-    (error instanceof UnsupportedChainIdError || chainId !== ChainId.MAINNET);
-
+  const { openSidebar, setOpenSidebar, error, removeError, provider } =
+    useRoutes();
+  const showError = provider && error instanceof WrongNetworkError;
   return (
     <>
-      {globalError && (
+      {error && !(error instanceof WrongNetworkError) && (
         <SuperHeader>
-          <div>{globalError}</div>
+          <div>{error.message}</div>
           <RemoveErrorSpan onClick={() => removeError()}>X</RemoveErrorSpan>
         </SuperHeader>
       )}
-      {wrongNetworkSend && location.pathname === "/" && (
+      {showError && (
         <SuperHeader>
           <div>
             You are on an incorrect network. Please{" "}
-            <button
-              onClick={() =>
-                switchChain(
-                  provider,
-                  sendState.currentlySelectedFromChain.chainId
-                )
-              }
-            >
-              switch to{" "}
-              {CHAINS[sendState.currentlySelectedFromChain.chainId].name}
-            </button>
-          </div>
-        </SuperHeader>
-      )}
-
-      {wrongNetworkPool && (
-        <SuperHeader>
-          <div>
-            You are on an incorrect network. Please{" "}
-            <button onClick={() => switchChain(provider, ChainId.MAINNET)}>
-              switch to {CHAINS[ChainId.MAINNET].name}
+            <button onClick={() => switchChain(provider, error.correctChainId)}>
+              switch to {CHAINS[error.correctChainId].name}
             </button>
           </div>
         </SuperHeader>
--- a/src/components/SendForm/SendForm.tsx
+++ b/src/components/SendForm/SendForm.tsx
@@ -6,22 +6,38 @@ import {
   AddressSelection,
   SendAction,
 } from "components";
-import { SendFormProvider } from "hooks";
+import { useSendForm } from "hooks";
 import type { Deposit } from "views/Confirmation";
+import SuperHeader from "components/SuperHeader";
+import { useConnection } from "state/hooks";
+import { CHAINS, switchChain } from "utils";
 
 type Props = {
   onDepositConfirmed: (deposit: Deposit) => void;
 };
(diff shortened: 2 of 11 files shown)


src/Routes.tsx
```diff
@@ -1,68 +1,57 @@
...
+  // reset wrong chain error on route change
+  const prevLocation = usePrevious(location);
+  useEffect(() => {
+    if (
+      error instanceof WrongNetworkError &&
+      prevLocation.pathname !== location.pathname
+    ) {
+      removeError();
+    }
+  }, [removeError, error, location.pathname, prevLocation.pathname]);
```
mrice32, line 26: This is really interesting. Why is this part of the app responsible for unsetting the error?

I'm assuming there's some component, like useSendFormManager that cares about the chainId being set according to a rule and adds this error. Can't that component remove the error when it unmounts by returning a function? ex: `return () => removeError();`

src/hooks/useSendForm.tsx
```diff
@@ -258,6 +259,14 @@ type SendFormManagerContext = FormState & {
...
   const [state, dispatch] = useReducer(formReducer, initialFormState);
   const { account: connectedAccount, chainId } = useConnection();
+  const { addError } = useError();
+
+  // Dispatch a global error if the user is connected but not to the fromChain
+  useEffect(() => {
+    if (chainId && chainId !== state.fromChain) {
+      addError(new WrongNetworkError(state.fromChain));
+    }
+  });
```
mrice32, line 269: No dependencies?

Shouldn't this depend on state.fromChain and chainId?

src/views/Pool.tsx
```diff
@@ -53,9 +54,14 @@ const Pool: FC = () => {
...
-      chainId !== DEFAULT_TO_CHAIN_ID);
-
+    (error instanceof UnsupportedChainIdError || chainId !== ChainId.MAINNET);
+  const { addError } = useError();
+  // if the user is connected but not to mainnet, dispatch a global error
+  useEffect(() => {
+    if (chainId && chainId !== ChainId.MAINNET) {
+      addError(new WrongNetworkError(ChainId.MAINNET));
+    }
+  });
```
mrice32, line 64: Same question about dependencies here.

mrice32 (commented):
Looks good! A few minor questions.

src/hooks/useSendForm.tsx
```diff
@@ -15,8 +15,9 @@ import {
   ParsingError,
   InsufficientBalanceError,
   ETH_ADDRESS,
```
Tulun, line 17: Isn't this entire file a context, minus the bottom 4-5 lines? Why is this a hook?

src/Routes.tsx
Tulun: It would have to be an effect Matt. If changing the route removes the error, it seems reasonable it could live in the useRouter hook.

We could have generic hook that has no dependencies that does this cleanup if we want, but it sort of makes sense we could do this in one place than telling each view to remove a global banner.

Gamaranto: @mrice32 I hadn't thought of that, changed it.
One thing to note though, effects cleanups run on every render, not only when a component unmounts. 
The cycle is render --> effect cleanup **_from the previous render_** --> effect , so the errors gets unset and then set immediately after, so its fine.
Note: This also gives a warning:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

Because we're now calling `removeError` in the cleanup function, but I think its fine, as the `ErrorContext` we're updating is not specific to one component, so updates are not wasted

src/hooks/useSendForm.tsx
Gamaranto: I wanted to re-run it in case location changed too, added deps back in the last commit

src/Routes.tsx
```diff
@@ -1,68 +1,57 @@
...
+    error,
+    removeError,
+    location,
+  };
+}
+// Need this component for useLocation hook
+const Routes: React.FC = () => {
+  const { openSidebar, setOpenSidebar, error, removeError, provider } =
+    useRoutes();
+  const showError = provider && error instanceof WrongNetworkError;
```
daywiss, line 40: this is misleading to call this `showError` when really its `showWrongNetworkError`.  Why not treat all errors as distinct and switch on the type if an error exists?  Vs what it seems like now, you treat wrong network error as a special case and all other errors as the same.

src/views/Pool.tsx
```diff
@@ -53,9 +54,18 @@ const Pool: FC = () => {
...
+  // // if the user is connected but not to mainnet, dispatch a global error
+  useEffect(() => {
+    const isWrongNetworkError = globalError instanceof WrongNetworkError;
+    if (!isWrongNetworkError && chainId && chainId !== ChainId.MAINNET) {
+      addError(new WrongNetworkError(ChainId.MAINNET));
+    }
+    if (isWrongNetworkError && chainId === ChainId.MAINNET) {
+      removeError();
+    }
+  }, [chainId, addError, removeError, globalError]);
```
daywiss, line 68: this logic is duplicated, and is also quite complicated to get right. it seems like there should be some abstraction here to prevent someone from having to duplicate this code for every "tab".

src/Routes.tsx
daywiss: there are multiple places in the app removing the error, this seems like a smell. but i understand why you do this, because each tab is responsible for setting its own network error

daywiss (commented):
@Gamaranto i think the reason why you arent completely happy with this solution is because it tries to be abstract, but internals of the implementation leak into inappropriate places in the code.  If you want to type errors, then treat those types as a way to hide business logic from the rest of the app.  I'm just spitballing here but I could imagine something like this:

```js
// inside a higher level hook, "useErrorManager"
const {setError,removeError,error,} = useError()
const {state} = useAppState() // this is just pseudo code for way to get data about the app

//check state for all errors you need to set
let nextError:Error | undefined;
// error manager handles generating errors
if(state.desiredChain != state.requiredChain){
   nextError = new NetworkError(state.requiredChain);
}else if (some other error condition){
  nextError = new Error('some other error')
}
// error manager handles removing the errors
if(nextError){
  setError(nextErrror)
}else if(error){
  removeError()
}

return { error }
```

Theres still an issue with rendering of the error, but if you standardize an interface for the banner, you could, rather than return an error from the manager, return JSX which you can just render in the banners place. 

```js
if(state.desiredChain != state.requiredChain){
   component = <div>
            You are on an incorrect network. Please{" "}
            <button onClick={() => switchChain(provider, error.correctChainId)}>
              switch to {CHAINS[error.correctChainId].name}
            </button>
          </div>
}
...
return component
```

something like that.  There may be technical reasons why you cant do this, but it would make a lot mroe sense to me to have this logic grouped into some kind of management system rather than spread around the app. WDYT?

Gamaranto:
@daywiss @Tulun @mrice32 I disliked all the synchronization done with useEffects, so I went ahead and used portals instead, from the [React docs](https://reactjs.org/docs/portals.html)

> Portals provide a first-class way to render children into a DOM node that exists outside the DOM hierarchy of the parent component.

By rendering the `SuperHeader` in a Portal, we can use it as a normal component everywhere in the code (and so access local state), but it will always render on top of our existing header.
This also gives each page the ability to re-style its instance of the `SuperHeader`, or add arbitrary logic to it. This also completely removes the need for synchronization (which we should usually try to avoid, as its easy to mess up) since the condition to show the banner and the children inside it can be derived from local state

daywiss (commented):
interesting approach, but im running into issues during testing.

1. open page, set metamask to arbitrum
2. go to bridge, reload page
3. connect wallet, see that optimism is the default from network ( expected it to be arbitrum)
4. go to pool tab, see there is no wrong network error ( expected it to warn to go to mainnet)

Gamaranto:
> interesting approach, but im running into issues during testing.
> 
> 1. open page, set metamask to arbitrum
> 2. go to bridge, reload page
> 3. connect wallet, see that optimism is the default from network ( expected it to be arbitrum)
> 4. go to pool tab, see there is no wrong network error ( expected it to warn to go to mainnet)

Thanks for catching this, it was caused by the cleanup of the `useLayoutEffect`, removed it, and now everything should work fine

src/Routes.tsx
daywiss: > I just have the sense that the logic of setting and removing errors being localized is simpler to understand than each action being done in different files across the app.

I came to the same conclusion, what i was proposing was doing the setting and unsetting in a component called `ErrorManager`, and implementation could be hidden from the rest of the app, and only consume the error.

src/Routes.tsx
```diff
@@ -3,70 +3,38 @@ import { Switch, Route, useLocation } from "react-router-dom";
...
+  };
+}
+// Need this component for useLocation hook
+const Routes: React.FC = () => {
+  const { openSidebar, setOpenSidebar, error, removeError } = useRoutes();
 
   return (
     <>
-      {globalError && (
+      {error && !(error instanceof WrongNetworkError) && (
```
daywiss, line 32: do we need this banner here? i dont remember if we have any errors other than "wrong network".  i think theres a problem that if you happen to be on the wrong network and set a non wrong network error, you will potentially see 2 banners.

edit: i guess because this is a "portal" you will only see 1 banner, but its actually unclear to me what will happen if 2 separate compoenents try to render this at the same time

src/Routes.tsx
```diff
@@ -3,70 +3,38 @@ import { Switch, Route, useLocation } from "react-router-dom";
...
-// Need this component for useLocation hook
-const Routes: React.FC = () => {
+function useRoutes() {
   const [openSidebar, setOpenSidebar] = useState(false);
-  const { provider, chainId, error } = useConnection();
+  const { provider } = useConnection();
   const location = useLocation();
-  const sendState = useAppSelector((state) => state.send);
-  const { error: globalError, removeError } = useError();
+  const { error, removeError } = useError();
```
daywiss, line 15: this is the only place where we use this now, each tab shows/hides their own banner independently.  do we even need this error hook anymore, or the error type since its not really being used?

src/components/SuperHeader/SuperHeader.tsx
```diff
@@ -26,4 +28,25 @@ const SuperHeader = styled.div`
   }
 `;
 
+/**
+ * React component that renders its children in a super header on top of the page.
+ */
+const SuperHeader: React.FC = ({ children }) => {
```
daywiss, line 34: very interesting

daywiss (commented):
this is a lot cleaner using the portal, but we still kind of suffer from the same problem ( less so than before), that each component needs to have logic to calculate if the banner should be shown. this has potential race conditions.  theres also not a clear owner of the process of setting an error. 

im willing to overlook these issues, but it just seems like having some kind of global state representation would be really useful for this type of feature. 

I think since you are using this method theres room to cleanup some stale code. The error types and error hook are not really used anymore.

src/components/SendForm/SendForm.tsx
```diff
@@ -6,22 +6,38 @@ import {
...
+import { useConnection } from "state/hooks";
+import { CHAINS, switchChain } from "utils";
 
 type Props = {
   onDepositConfirmed: (deposit: Deposit) => void;
 };
 const SendForm: React.FC<Props> = ({ onDepositConfirmed }) => {
+  const { fromChain } = useSendForm();
+  const { chainId, provider } = useConnection();
+  const showError = !!chainId && provider && chainId !== fromChain;
```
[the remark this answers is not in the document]
Gamaranto: good call, will include in a hook and change the name

src/Routes.tsx
Gamaranto: I think was initially set for stuff like network errors, I think we can remove it though

src/Routes.tsx
Gamaranto: I think we can get rid of it

Gamaranto:
> this is a lot cleaner using the portal, but we still kind of suffer from the same problem ( less so than before), that each component needs to have logic to calculate if the banner should be shown. this has potential race conditions. theres also not a clear owner of the process of setting an error.
> 
> im willing to overlook these issues, but it just seems like having some kind of global state representation would be really useful for this type of feature.
> 
> I think since you are using this method theres room to cleanup some stale code. The error types and error hook are not really used anymore.

I think the error is not global, but rather "scoped" to a page, so its necessary that each page has its own logic for setting the error.
I completely agree with you on your second point, that is, if 2 components try to render the banner at the same time, then the last one wins, which is an annoying race condition. 
One way to solve this (if we run into it) could be to have a context which stores a queue of errors and orders them based on priority, and then render only the first one in the queue, wdyt?

src/Routes.tsx
Gamaranto: actually, it seems used in the pool tab, will keep it in now to keep PR size down, and will remove all dead code in a future PR

src/Routes.tsx
Gamaranto: actually, it seems used in the pool tab, will keep it in now to keep PR size down, and will remove all dead code in a future PR

daywiss: ok

daywiss:
> One way to solve this (if we run into it) could be to have a context which stores a queue of errors and orders them based on priority, and then render only the first one in the queue, wdyt?

I think its not really necessary in this PR. but if you think we would benefit from a standard way of setting/showing errors then that would make sense. In this implementation, we only really care that theres an "owner" of the banner, and whoever owns it sets/removes the banner, you dont even need a notion of errors until theres some kind of "global" system where any component could dispatch an error.  Currently it seems like the "owner" of the banner is the current tab ( if you remove error logic from router), which is fine for now but i can see that being a potential coordination problem in the future, that we could address then.

daywiss (approved):

mrice32 (approved):

PR #16, opened 2022-03-25
Fix: Tx history css qa

Tulun:
Adjustments ALA: https://app.shortcut.com/uma-project/story/4743/create-view-of-transaction-history-with-an-ongoing-account-for-an-account

LMK if this works


daywiss (approved):

Tommy1231232 (changes requested):
@Tulun looks like we still need to reduce the spacing between columns

Screen Shot 2022-03-28 at 7 55 23 AM

Tulun:
> @Tulun looks like we still need to reduce the spacing between columns
> 
> Screen Shot 2022-03-28 at 7 55 23 AM
Screen Shot 2022-03-29 at 11 58 13 AM

LMK if this is ok

Tommy1231232 (approved):

PR #17, opened 2022-03-28
fix(across-v2): constants refactor

Gamaranto:
**Motivation** 
This PR partially refactors the constants file in order to make it more understandable and easy to reason about
**Details**

1. Testnet chains have been removed for now. Screening them out complicates the code a lot (type safety mainly), and I don't have a great solution for it. We haven't really used them, so I think this is ok, but happy to hear other opinions
2. Dropdowns now take flat chain Ids, and the view then displays things using the `CHAINS` object.
3. The type safety of the `ADDRESS` object has been improved, and I've added a `CANONICAL_BRIDGE` object for the addresses of the canonical bridges of each L2. I hardcoded them for now (this is hacky since they can change), but this is still an improvement over what we had before, as we were using a mock address on MAINNET and checking for allowance against that address, but then approving the canonical bridge address. See here: https://github.com/across-protocol/frontend-v2/blob/506e50fbd55c33d8bbd1a4a5d2e97eb933d48972/src/hooks/useBridge.ts#L51

Regarding changing the `CHAINS` data structure, I've tried a few, wasn't particularly satisfied with each one, but didn't want to stall progress on this PR, so I omitted it for now


src/utils/constants.ts
```diff
@@ -390,7 +297,9 @@ export const TOKENS_LIST: Record<ChainId, TokenList> = {
 
 export type ChainInfo = {
   name: string;
+  fullName?: string;
   chainId: ChainId;
+  ETHAddress: string;
```
daywiss, line 302: what does this value mean on other chains?

src/utils/chains.ts
```diff
@@ -17,21 +17,24 @@ export async function switchChain(
         await provider.send("wallet_addEthereumChain", [
           {
             chainId: ethers.utils.hexValue(chainId),
-            chainName: CHAINS[chainId].name,
+            chainName: CHAINS[chainId].fullName ?? CHAINS[chainId].name,
             rpcUrls: [CHAINS[chainId].rpcUrl],
             blockExplorerUrls: [CHAINS[chainId].explorerUrl],
             nativeCurrency: CHAINS[chainId].nativeCurrency,
```
daywiss, line 23: i used this structure as an inspiration for how the chains configuration object should be shaped, since its somewhat standard

src/utils/constants.ts
```diff
@@ -520,32 +378,35 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
       symbol: "ETH",
       decimals: 18,
     },
+    ETHAddress: ethers.constants.AddressZero,
   },
 };
+/** Chains as they appear in dropdowns */
+export const CHAINS_SELECTION = [
```
daywiss, line 385: is this the chain dropdown whitelist? if so we could potentially keep in the definitions for testnets and use this as the source for whats shown in form

src/utils/constants.ts
```diff
@@ -602,18 +439,8 @@ export function getChainName(chainId: ChainId): string {
...
-    case ChainId.OPTIMISM:
-      return CHAINS[ChainId.OPTIMISM].name;
-    case ChainId.KOVAN_OPTIMISM:
-      return CHAINS[ChainId.KOVAN_OPTIMISM].name;
     case ChainId.ARBITRUM:
       return CHAINS[ChainId.ARBITRUM].name;
-    case ChainId.ARBITRUM_RINKEBY:
-      return CHAINS[ChainId.ARBITRUM_RINKEBY].name;
     default:
       return "unkwown";
```
daywiss, line 445: this is misspelled

src/components/ChainSelection/ChainSelection.tsx
```diff
@@ -29,16 +29,20 @@ const ChainSelection: React.FC = () => {
...
     <Section>
       <Wrapper>
         <SectionTitle>From</SectionTitle>
         <InputGroup>
           <RoundBox as="label" {...getLabelProps()}>
             <ToggleButton type="button" {...getToggleButtonProps()}>
-              <Logo src={selectedItem?.logoURI} alt={selectedItem?.name} />
-              <ToggleChainName>{selectedItem?.name}</ToggleChainName>
+              <Logo
+                src={CHAINS[selectedItem].logoURI}
```
daywiss, line 42: this tends to be indexed a lot, would make sense if this was indexed to a var and used, but maybe u wanted to avoid having to add it to a hook

src/utils/constants.ts
```diff
@@ -655,99 +482,9 @@ export function onboardBaseConfig(): Initialization {
 
 // this client requires multicall2 be accessible on the chain. This is the address for mainnet.
 export const multicallTwoAddress = "0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696";
-export interface IChainSelection {
```
daywiss, line 658: was this redundant or something

daywiss (commented):
was hoping we could get to a more configuration based rather than constants based system, as it seems inconvenient that we cant easily remove/add chains based on context (dev, staging, prod).  not a huge deal but we will need to figure out how we are going to test things off of mainnet.  wondering if we should spend more time to work envs into this pr so we can have some configurability? @Gamaranto @mrice32 

also sent francesco a link to a package i tried out for typesript validations, which is relevant: https://github.com/ianstormtaylor/superstruct

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -79,17 +84,15 @@ const AddressSelection: React.FC = () => {
...
                     >
-                      <Logo src={t.logoURI} alt={t.name} />
-                      <div>{t.name}</div>
+                      <Logo src={CHAINS[t].logoURI} alt={CHAINS[t].name} />
+                      <div>{CHAINS[t].name}</div>
                       <span className="layer-type">
-                        {t.chainId !== ChainId.MAINNET ? "L2" : "L1"}
+                        {t !== ChainId.MAINNET ? "L2" : "L1"}
                       </span>
                     </Item>
```
Tulun, line 97: Any reason this isn't abstracted into a styled-component?

Tulun (approved):
LGTM.

mrice32 (approved):
LGTM!

src/components/ChainSelection/ChainSelection.tsx
```diff
@@ -29,16 +29,20 @@ const ChainSelection: React.FC = () => {
     buttonText,
     fromChain,
   } = useChainSelection();
-
+  const selectedItem = maybeSelectedItem ?? DEFAULT_FROM_CHAIN_ID;
+  console.log(selectedItem);
```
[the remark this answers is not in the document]
Gamaranto: thanks removed

src/utils/constants.ts
Gamaranto: This is supposed to be the ETH_ADDRESS on that chain (or the base asset, but its always ETH for now), for example Optimism has `0x42000....` as address for ETH, for chains where this is not a thing, the zero address can be used

src/utils/constants.ts
Gamaranto: yes, let's actually go ahead and do that (i'll reintroduce the testnets)

src/utils/constants.ts
Gamaranto: sorry, corrected

src/utils/constants.ts
Gamaranto: yes, it was a duplicate and was necessary because we had testnets in the `CHAINS` object and could not use it directly, it has now been replaced and you only define the chainIds

src/components/AddressSelection/AddressSelection.tsx
Gamaranto: not sure, but nice catch will change this

src/components/ChainSelection/ChainSelection.tsx
Gamaranto: good idea, indexed it in the body of the function

daywiss (approved):

PR #19, opened 2022-03-30
feat(across-v2): integrate bridge tab with v2

Gamaranto:
This PR integrates the fe with the v2 contracts.


src/utils/bridge.ts
```diff
@@ -413,34 +442,24 @@ type AcrossApprovalArgs = BaseApprovalArgs & {
...
-    timestamp,
-    slowRelayFeePct,
-    instantRelayFeePct,
-    referrer,
-  }: AcrossDepositArgs
+  { fromChain, token, amount, toAddress: recipient, toChain: destinationChainId, relayerFeePct, timestamp: quoteTimestamp, referrer }: AcrossDepositArgs
 ): Promise<ethers.providers.TransactionResponse> {
   if (!isL2(fromChain)) {
     throw new Error(
       "Across does not support mainnet deposits. The canonical bridge should be used instead."
```
daywiss, line 449: just curious if this is actually still the case with v2? @mrice32

daywiss (approved):
looks good other than a few questions and comments!

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -49,7 +49,7 @@ export default function useAddressSelection() {
     }
   };
 
-  const isL1toL2 = fromChain === ChainId.MAINNET;
+  const isL1toL2 = !isL2(fromChain);
```
mrice32, line 52: This may be premature, but I think it might make the implementation simpler if we remove all special casing around L2 vs L1. I think mainnet can be treated just like any other chain. You can send from or to mainnet and mainnet has a spoke pool just like everywhere else.

Note: Jesper said he thinks it's fine to drop all labels related to L2 and L1 from the UI, as I think Ethereum vs Optimism is clear enough.

Thoughts?

src/utils/bridge.ts
```diff
@@ -263,6 +255,40 @@ export function filterTokensByDestinationChain(
...
+    return false;
+  }
+  // can't bridge between testnets and prod chains
+  if ((isTestnet(chainA) && !isTestnet(chainB)) || (!isTestnet(chainA) && isTestnet(chainB))) {
+    return false;
+  }
+  // can't bridge between L1s 
+  if (!isL2(chainA) && !isL2(chainB)) {
+    return false;
+  }
```
mrice32, line 276: This is kind of an odd check? What are we considering an L1? And if there were two distinct L1s, why couldn't we bridge between them?

I think the L2/L1 language is a little confusing. I think for the purposes of the dapp, we should think of all chains as equivalent, except that one chain is special in that it holds the pools (this would be Ethereum in prod, but could be something else on testnet).

Thoughts? Do you think this will require a lot of restructuring?

src/utils/bridge.ts
```diff
@@ -263,6 +255,40 @@ export function filterTokensByDestinationChain(
...
+  // check if they have at least one token in common
+  return filterTokensByDestinationChain(chainA, chainB).length > 0;
+}
+
+/**
+ * 
+ * @param fromChain the chain to bridge from, that is, the chain that tokens are sent from.
+ * @returns The list of chains that can be bridged to from the given `fromChain`.
+ */
+export function getReacheableChains(fromChain: ChainId): ChainId[] {
```
mrice32, line 286: +1!

src/utils/constants.ts
```diff
@@ -535,30 +535,34 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
...
-  },
-  [ChainId.ARBITRUM_RINKEBY]: {
-    BRIDGE: "0x6999526e507Cc3b03b180BbE05E1Ff938259A874",
-  },
-  [ChainId.KOVAN_OPTIMISM]: {
-    BRIDGE: "0x2271a5E74eA8A29764ab10523575b41AA52455f0",
-  },
+
+/** FIXME:  use the actual spoke pool addresses!!!! */
+export const SPOKES_ADDRESSES: Record<
```
mrice32, line 557: nit: I think this should either be SPOKE_ADDRESSES or SPOKE_POOL_ADDRESSES.

src/utils/math.ts
```diff
@@ -6,21 +6,13 @@ export function max(a: BigNumberish, b: BigNumberish) {
...
-  slowRelayFee: Pick<BridgeFees["slowRelayFee"], "total">;
-} & {
-  lpFee: Pick<BridgeFees["lpFee"], "total">;
-};
-export function receiveAmount(amount: BigNumberish, fees?: PartialBridgeFees) {
-  const amountBn = BigNumber.from(amount);
-  if (amountBn.lte(0)) return BigNumber.from(0);
-  if (!fees) return amountBn;
+
+export function receiveAmount(amount: BigNumber, fees?: BridgeFees) {
```
mrice32, line 10: Can you add a comment about when this is supposed to be used?

When would we have no fees or an amount < 0?

mrice32 (commented):
Looks awesome! A few comments

src/utils/bridge.ts
```diff
@@ -263,6 +255,40 @@ export function filterTokensByDestinationChain(
   return TOKENS_LIST[chainA].filter(filterByToChain);
 }
 
+/**
+  * Checks if its possible to bridge from chain A to chain B.
+  * @param chainA  the chain to bridge from, that is, the chain that tokens are sent from.
+  * @param chainB  the destination chain, that is, where tokens will be sent.
+  * @returns Returns `true` if it is possible to bridge from chain A to chain B, `false` otherwise.
+*/
+function canBridge(chainA: ChainId, chainB: ChainId): boolean {
```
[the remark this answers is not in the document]
Tulun: I might add typing to the docs.

src/utils/bridge.ts
```diff
@@ -263,6 +255,40 @@ export function filterTokensByDestinationChain(
...
+  // can't bridge to itself
+  if (chainA === chainB) {
+    return false;
+  }
+  // can't bridge between testnets and prod chains
+  if ((isTestnet(chainA) && !isTestnet(chainB)) || (!isTestnet(chainA) && isTestnet(chainB))) {
```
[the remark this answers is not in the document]
Tulun: I also think adding extra code in prod for testing seems a bit off -- the option in our UI to send to / from a testnet should never be in prod in the first place (either under a configuration or feature flags.

Tulun (commented):
Generally looks good to me, just address Matt's and David's concerns

src/utils/constants.ts
Gamaranto: good catch, changed it

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -77,55 +80,25 @@ const AddressSelection: React.FC = () => {
...
               {isOpen &&
-                toChain !== ChainId.MAINNET &&
                 CHAINS_SELECTION.map((t, index) => {
                   return (
                     <Item
                       className={
-                        t === toChain || t === ChainId.MAINNET ? "disabled" : ""
+                        t === toChain ||
+                        t === fromChain ||
+                        !reacheableChains.includes(t)
```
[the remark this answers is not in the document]
Gamaranto: yes, good catch

src/utils/bridge.ts
```diff
@@ -263,6 +255,40 @@ export function filterTokensByDestinationChain(
...
+    return false;
+  }
+  // can't bridge between L1s 
+  if (!isL2(chainA) && !isL2(chainB)) {
+    return false;
+  }
+  // check if they have at least one token in common
+  return filterTokensByDestinationChain(chainA, chainB).length > 0;
```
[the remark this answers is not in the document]
Gamaranto: I had a 2 misunderstandings:
1. that you couldn't bridge between testnets and other chains protocol wise (wrong)
2. that we wanted to keep using canonical bridges for v2

I have cleared those out and now this is the only check left

src/state/chainApi.ts
```diff
@@ -101,42 +94,12 @@ const api = createApi({
...
-      },
-    }),
   }),
 });
 
 export const {
   useBalancesQuery: useBalances,
   useEthBalanceQuery: useETHBalance,
   useAllowanceQuery: useAllowance,
-  useBridgeFeesQuery: useBridgeFees,
```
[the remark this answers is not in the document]
Gamaranto: sgtm!

Tulun (approved):

src/utils/math.ts
mrice32: This makes sense, but if it hasn't been fetched, then why do we return amount? Shouldn't we return 0, null, or undefined to indicate that we can't compute a valid result right now?

As a side note: since this isn't a hook, we can call it optionally. Do you think it would be preferable to have the caller short circuit if the values that are needed to call this function aren't available? That may make this function's behavior clearer under different circumstances. Note: I'm not super strongly opinionated here, though.

src/utils/math.ts
```diff
@@ -6,22 +6,8 @@ export function max(a: BigNumberish, b: BigNumberish) {
...
-  if (!fees) return amountBn;
-  return max(
-    amountBn
-      .sub(fees.instantRelayFee.total)
-      .sub(fees.slowRelayFee.total)
-      .sub(fees.lpFee.total),
-    0
-  );
+export function receiveAmount(amount: BigNumber, fees?: BridgeFees) {
+  if (amount.eq(0)) return BigNumber.from(0);
```
mrice32, line 10: Why do we need to check for this? Won't the statement at the end return 0 if its result is 0 or negative anyway?

mrice32 (approved):

src/utils/math.ts
Gamaranto: good catch, fixed it, also made it only accept `fees` if defined, and handled the defaulting logic where its consumed

PR #20, opened 2022-03-31
feat: integrate v2 pool client into pool tab

daywiss:

Integrates sdk pool client. Minor changes were needed to pool logic. 
Before this is merged, it will automatically run on kovan, which is how I tested it. so you should be good to actually run through the pool tab UI.

### DIFF
--- a/src/components/SendAction/useSendAction.ts
+++ b/src/components/SendAction/useSendAction.ts
@@ -1,23 +1,18 @@
 import { useState } from "react";
 import { useSendForm, useBridgeFees, useBridge } from "hooks";
-import {
-  onboard,
-  TOKENS_LIST,
-  Token,
-  CONFIRMATIONS,
-} from "utils";
+import { onboard, TOKENS_LIST, Token, CONFIRMATIONS } from "utils";
 import { Deposit } from "views/Confirmation";
 import { useConnection } from "state/hooks";
 
 type TokenInfo =
   | {
-    address: string;
-    symbol: "WETH";
-    name: "Wrapped Ether";
-    decimals: 18;
-    logoURI: string;
-    bridgePool: string;
-  }
+      address: string;
+      symbol: "WETH";
+      name: "Wrapped Ether";
+      decimals: 18;
+      logoURI: string;
+      bridgePool: string;
+    }
   | Token;
 
 export default function useSendAction(
--- a/src/state/connection.ts
+++ b/src/state/connection.ts
@@ -1,5 +1,6 @@
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { ethers } from "ethers";
+import { POOL_CHAINID } from "utils";
 import {
   ChainId,
   UnsupportedChainIdError,
@@ -25,7 +26,7 @@ type ErrorUpdate = Required<Pick<State, "error">>;
 const initialState: State = {
   notify: Notify({
     dappId: process.env.REACT_APP_PUBLIC_ONBOARD_API_KEY, // [String] The API key created by step one above
-    networkId: 1, // [Integer] The Ethereum network ID your Dapp uses.
+    networkId: POOL_CHAINID,
     desktopPosition: "topRight",
   }),
 };
--- a/src/state/connection.ts
+++ b/src/state/connection.ts
@@ -1,6 +1,6 @@
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { ethers } from "ethers";
-import { POOL_CHAINID } from "utils";
+import { HUBPOOL_CHAINID } from "utils";
 import {
   ChainId,
   UnsupportedChainIdError,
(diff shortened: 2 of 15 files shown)


src/utils/constants.ts
```diff
@@ -306,6 +312,28 @@ export const SPOKE_ADDRESSES: Record<ChainId, string> = {
...
+  [ChainId.BOBA]: ethers.constants.AddressZero,
+  [ChainId.ARBITRUM]: ethers.constants.AddressZero,
+  [ChainId.RINKEBY]: ethers.constants.AddressZero,
+  [ChainId.KOVAN]: getAddress("0xD449Af45a032Df413b497A709EeD3E8C112EbcE3"),
+  [ChainId.KOVAN_OPTIMISM]: ethers.constants.AddressZero,
+  [ChainId.ARBITRUM_RINKEBY]: ethers.constants.AddressZero,
+};
+// Update once addresses are known
+export const RATEMODEL_ADDRESSES: Record<ChainId, string> = {
+  [ChainId.MAINNET]: ethers.constants.AddressZero,
```
daywiss, line 328: app will crash if we end up going to production without setting these, the config generator will check that no addresses going into pool client are address zero

src/components/SendAction/useSendAction.ts
```diff
@@ -1,23 +1,18 @@
...
-    bridgePool: string;
-  }
+      address: string;
+      symbol: "WETH";
+      name: "Wrapped Ether";
+      decimals: 18;
+      logoURI: string;
+      bridgePool: string;
+    }
   | Token;
```
Tulun, line 16: What's the difference between these two?

Gamaranto: +1, we should add a task to standardize linting, we keep having those weird formatting diffs in PRs

src/state/connection.ts
```diff
@@ -25,7 +26,7 @@ type ErrorUpdate = Required<Pick<State, "error">>;
 const initialState: State = {
   notify: Notify({
     dappId: process.env.REACT_APP_PUBLIC_ONBOARD_API_KEY, // [String] The API key created by step one above
-    networkId: 1, // [Integer] The Ethereum network ID your Dapp uses.
+    networkId: POOL_CHAINID,
```
Gamaranto, line 29: nit: should we name this something like `HUB_CHAINID` ? I think it would be ideal if we only ever mentioned HUB or POOL, but not both

daywiss: Good point, ok

src/components/SendAction/useSendAction.ts
daywiss: No idea! I can double check it

src/utils/constants.ts
```diff
@@ -306,6 +312,28 @@ export const SPOKE_ADDRESSES: Record<ChainId, string> = {
     "0x3BED21dAe767e4Df894B31b14aD32369cE4bad8b"
   ),
 };
+// Update once addresses are known
+export const HUBPOOL_ADDRESSES: Record<ChainId, string> = {
+  [ChainId.MAINNET]: ethers.constants.AddressZero,
```
mrice32, line 317: nit: why do we put the tokens in a JSON, but have these written into the code?

src/views/Pool.tsx
```diff
@@ -26,50 +27,69 @@ import { SuperHeader } from "components";
...
-    chainId: ChainId.MAINNET,
+    chainId,
     account,
     tokenAddress: token.address,
   });
 
   const wrongNetwork =
     provider &&
-    (error instanceof UnsupportedChainIdError || chainId !== ChainId.MAINNET);
+    (error instanceof UnsupportedChainIdError || chainId !== activeChainId);
```
mrice32, line 70: Why did this change? I thought we would still expect the pool tab to only support mainnet transactions/connections, right?

src/views/Pool.tsx
```diff
@@ -26,50 +27,69 @@ import { SuperHeader } from "components";
...
-    poolClient.updatePool(token.bridgePool).then((res) => {
-      setLoadingPoolState(false);
-    });
+    poolClient
+      .updatePool(token.address)
+      .then((res) => {
+        setLoadingPoolState(false);
+      })
+      .catch((err) => {
+        console.error("Unable to load pool info", err);
```
mrice32, line 82: Is this errors, should we remove the loading state? Not a suggestion, genuine question.

mrice32 (approved):
LGTM! This is awesome, @daywiss! Only minor questions/comments

amateima (approved):

src/views/Pool.tsx
daywiss: good point

src/utils/constants.ts
daywiss: was following the pattern on the spoke addresses, though i think you are right, this could use some further refactoring. lets see what @Gamaranto thinks

src/components/SendAction/useSendAction.ts
daywiss: i ran this again and this is how the linter wrote it

src/views/Pool.tsx
daywiss: the chain id is set automatically through the `isProduction()` function

PR #24, opened 2022-04-06
fix: add loading animation on deposit

amateima:
Signed-off-by: amateima <amatei@umaproject.org>

This PR adds an animation to the deposit button and also disables it while the deposit transaction is in pending state

Screenshot 2022-04-06 at 13 27 29

### DIFF
--- a/src/components/SendAction/useSendAction.ts
+++ b/src/components/SendAction/useSendAction.ts
@@ -67,17 +67,14 @@ export default function useSendAction(
             .catch(console.error);
           // TODO: we should invalidate and refetch any queries of the transaction tab, so when a user switches to it, they see the new transaction immediately.
         }
-
-        setTxPending(false);
         return tx;
       }
-    } catch (error) {
+    } finally {
       setTxPending(false);
-      throw error;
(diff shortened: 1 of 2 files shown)


src/components/SendAction/useSendAction.ts
```diff
@@ -38,32 +39,41 @@ export default function useSendAction(
...
+            .catch(console.error);
+          // TODO: we should invalidate and refetch any queries of the transaction tab, so when a user switches to it, they see the new transaction immediately.
+        }
+
+        setTxPending(false);
+        return tx;
       }
-      return tx;
+    } catch (error) {
+      setTxPending(false);
```
daywiss, line 75: you could add a finally block and have a single `setTxPending(false)` in it

daywiss (approved):

src/components/SendAction/useSendAction.ts
Tulun: +1.

Tulun (approved):
Add finally and you should be gtg.

PR #25, opened 2022-04-06
fix(across-v2): take timestamp from hubpool

Gamaranto:
This PR fixes a timestamp issue, `quoteTimestamps` were taken as the block timestamp of the sending chain, but now this is fixed and `getCurrentTime` on the `HubPool` is used.

### DIFF
--- a/src/hooks/useBridge.ts
+++ b/src/hooks/useBridge.ts
@@ -80,16 +80,21 @@ export function useBridge() {
     if (!signer || status !== "ready" || !toAddress || !block || !fees) {
       return;
     }
-    return sendAcrossDeposit(signer, {
-      toAddress,
-      amount,
-      token,
-      fromChain,
-      toChain,
-      relayerFeePct: fees.relayerFee.pct,
-      timestamp: await hubPool.getCurrentTime(),
-      referrer,
-    });
+    try {
+      return sendAcrossDeposit(signer, {
+        toAddress,
+        amount,
+        token,
+        fromChain,
+        toChain,
+        relayerFeePct: fees.relayerFee.pct,
+        timestamp: await hubPool.getCurrentTime(),
+        referrer,
+      });
+    } catch (e) {
+      console.error("Something went wrong when depositing.")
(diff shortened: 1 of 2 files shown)


src/utils/bridge.ts
```diff
@@ -30,6 +32,28 @@ export function getSpokePool(
...
+    case ChainId.KOVAN_OPTIMISM:
+      return ChainId.KOVAN;
+    default:
+      return ChainId.MAINNET
+  }
+}
+
+export function getHubPool(fromChain: ChainId, signer?: ethers.Signer): HubPool {
+  const hubPoolChainId = getHubPoolChainId(fromChain);
+  const maybeAddress = HUBPOOL_ADDRESSES[hubPoolChainId];
```
daywiss, line 49: you should error if this address is address zero, or add that check in `isValidAddress`. we should be immediately alerted to unset addresses

mrice32: Good point

mrice32 (approved):
LGTM!

src/utils/bridge.ts
Gamaranto: Since we use the zero address as an eth address, and it is a valid address, I propose we use an empty string for pool/spokes addresses that are not yet set, and check against valid addresses

src/hooks/useBridge.ts
```diff
@@ -84,7 +87,7 @@ export function useBridge() {
       fromChain,
       toChain,
       relayerFeePct: fees.relayerFee.pct,
-      timestamp: block.timestamp,
+      timestamp: await hubPool.getCurrentTime(),
       referrer,
```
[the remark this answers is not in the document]
Gamaranto: will add, good idea!

src/utils/bridge.ts
daywiss: i mean specifically for a hubpool address, that should never be address zero, so we should check that

daywiss (approved):

PR #27, opened 2022-04-07
feat: integrate TX history logic

Tulun:
Draft of implementation.


src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,48 @@
...
   const [openFilledRow, setOpenFilledRow] = useState<number>(-1);
   const [openOngoingRow, setOpenOngoingRow] = useState<number>(-1);
   const [currentPage, setCurrentPage] = useState(0);
+  const [rawFilledTx, setRawFilledTx] = useState<Transfer[]>([]);
+  const [rawOngoingTx, setRawOngoingTx] = useState<Transfer[]>([]);
+
+  useEffect(() => {
+    let interval: ReturnType<typeof setInterval>;
+    if (account) {
+      startFetchingTransfers(account).then((res) => {
```
daywiss, line 22: id suggest separating out this call from the interval. probably have a separate useEffect which listens to the account changing, clears the transaction state and a second useEffect which just continually polls the filled/pending transfers as long as account is defined

Tulun: Yeah I'd like to better test different accounts with real transaction data if possible.

src/utils/format.ts
```diff
@@ -99,3 +99,11 @@ export function tagAddress(dataHex: string, address: string) {
...
   return tagHex(dataHex, address);
 }
+
+export function capitalizeFirstLetter(str: string) {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
+export const formatNumberTwoSigDigits = new Intl.NumberFormat("en-US", {
+  maximumSignificantDigits: 2,
+}).format;
```
daywiss, line 109: im surprised this will work consistently without using `.bind`

src/views/Transactions/useTransactionsView.ts
```diff
@@ -28,7 +60,20 @@ export default function useTransactionsView() {
     setOpenOngoingRow,
     currentPage,
     setCurrentPage,
+    txHistoryClient,
+    rawFilledTx,
+    rawOngoingTx,
   };
 }
 
+async function startFetchingTransfers(account: string) {
```
daywiss, line 69: this is a bit wierd, its  just a wrapper that nulls and logs an error

daywiss (commented):
looking good

src/views/Transactions/useTransactionsView.ts
Tulun: You can't put async function directly into a useEffect.

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
...
   const { width } = useWindowSize();
   const [openFilledRow, setOpenFilledRow] = useState<number>(-1);
   const [openOngoingRow, setOpenOngoingRow] = useState<number>(-1);
   const [currentPage, setCurrentPage] = useState(0);
+  const [rawFilledTx, setRawFilledTx] = useState<Transfer[]>([]);
+  const [rawOngoingTx, setRawOngoingTx] = useState<Transfer[]>([]);
+
+  // Start the tracking / stopping of the TX in the client.
+  useEffect(() => {
+    if (account) {
```
mrice32, line 21: What happens on account _change_? Does if the account changes without this being unmounted, does that account get tracked forever?

If you change and change back, will this throw an error because you've tracked the same account twice?

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { useConnection } from "state/hooks";
 import { onboard } from "utils";
-import createTransactionModel from "./TransactionsTable/createTransactionModel";
 import useWindowSize from "hooks/useWindowsSize";
+import txHistoryClient from "state/transferHistory";
+import { Transfer } from "@across-protocol/sdk-v2/dist/transfers-history/model";
```
mrice32, line 6: nit: prefer importing from `@across-protocol/sdk-v2` and drilling down in the object that that returns.

src/state/transferHistory.ts
```diff
@@ -0,0 +1,42 @@
+import { transfersHistory } from "@across-protocol/sdk-v2";
+import { createTxHistoryClient } from "utils/constants";
+export enum TransfersHistoryEvent {
+  TransfersUpdated = "TransfersUpdated",
+}
+const { TransfersHistoryClient } = transfersHistory;
+
+const chainConfigs = createTxHistoryClient();
```
mrice32, line 8: nit: why are the chain configs returned by a function called createTxHistoryClient? I would expect that that function would return the client itself?

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
...
+  const [rawOngoingTx, setRawOngoingTx] = useState<Transfer[]>([]);
+
+  // Start the tracking / stopping of the TX in the client.
+  useEffect(() => {
+    if (account) {
+      txHistoryClient.startFetchingTransfers(account).catch((err) => {
+        console.error(
+          "Error in startFetchingTransfers call in txHistoryClient",
+          err
+        );
```
mrice32, line 26: Why do we have a special catch just for this? Do we expect this to error often or is there something different about this error than, for instance, stopFetchingTransfers?

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
...
+  // Create interval to update transactions.
+  useEffect(() => {
+    let interval: ReturnType<typeof setInterval>;
+    if (account) {
+      interval = setInterval(() => {
+        const nextFilledTx = txHistoryClient.getFilledTransfers(account);
+        const nextOngoingTx = txHistoryClient.getPendingTransfers(account);
+        setRawFilledTx(nextFilledTx);
+        setRawOngoingTx(nextOngoingTx);
+      }, 5000);
```
mrice32, line 45: nit: should this be a constant?

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
...
+      }
+    };
+  }, [account]);
+
+  // Create interval to update transactions.
+  useEffect(() => {
+    let interval: ReturnType<typeof setInterval>;
+    if (account) {
+      interval = setInterval(() => {
+        const nextFilledTx = txHistoryClient.getFilledTransfers(account);
```
mrice32, line 41: cc @amateima @daywiss 

Something I want to make a note of here:

We have a client that does internal polling to determine the transfers and then we have UI logic to _poll that client_ on a second interval since the client can't write directly to app state.

I think this is a code smell.

I think this might be an example of why clients that manage their own request timings aren't well suited for our new frontend philosophy or really any frontend that doesn't allow these clients to directly write to the frontend state.

A few styles of clients that might work better:
1. A client where updates are triggered via a function call. After the update is called, data can be retrieved synchronously. I think this would basically be this current client without polling, but with an update function.
2. A client that's essentially a caching object. Async functions are called on the client to ask for specific pieces of data. New data is requested, but only data after the last call/update is pulled. Basically an object with async functions that do caching of past queries to make future queries easier.
3. No client. Just pure async functions with no caching.

Do y'all agree with this assessment? Or do y'all think we should manage the client interaction differently and I'm misunderstanding the tradeoff space here?

mrice32 (commented):
Looks great! Some comments, but most are very minor.

src/views/Transactions/useTransactionsView.ts
```diff
@@ -1,25 +1,62 @@
...
+        txHistoryClient.stopFetchingTransfers(account);
+      }
+    };
+  }, [account]);
+
+  // Create interval to update transactions.
+  useEffect(() => {
+    let interval: ReturnType<typeof setInterval>;
+    if (account) {
+      interval = setInterval(() => {
```
mrice32, line 40: Is this somewhere where we could use react-query? Wouldn't it automate a lot of this logic for us?

Couldn't it poll, cache the state, and terminate the polling when one of the inputs changes?

src/views/Transactions/TransactionsTable/TransactionsTableWithPagination.tsx
```diff
@@ -41,9 +41,11 @@ const TransactionsTableWithPagination: FC<Props> = ({
   return (
     <>
       <TransactionsTable rows={paginatedRows} headers={headers} title={title} />
-      <PaginationWrapper>
-        <Pagination onPageChange={setCurrentPage} {...paginateState} />
-      </PaginationWrapper>
+      {paginatedRows.length > 25 && (
```
daywiss, line 44: you could do somehting like this 
```
      {paginateState.totalPages > 1 ? (
        <PageNavigation onPageChange={setCurrentPage} {...paginateState} />
      ) : null}
```

src/views/Transactions/useTransactionsView.ts
daywiss: good q for @amateima

src/views/Transactions/useTransactionsView.ts
daywiss: I'm not sure why theres a code smell. the way the client is designed its compatible with react query. You could set it up to poll/cache results from the calls.

Why do you say those designs will work better, is that because you think they technically will be better, the interfaces will be better or it conforms more to the design philosophy?

src/state/transferHistory.ts
```diff
@@ -0,0 +1,42 @@
...
+const client = new TransfersHistoryClient({
+  chains: chainConfigs,
+  // optional
+  pollingIntervalSeconds: 60000,
+});
+// optional
+client.setLogLevel("debug");
+
+export default client;
+// await client.startFetchingTransfers(<depositor_addr>);
```
Gamaranto, line 19: stray comments (I think)

src/utils/bridge.ts
```diff
@@ -40,19 +45,25 @@ function getHubPoolChainId(sendingChain: ChainId): ChainId {
...
+): HubPool {
   const hubPoolChainId = getHubPoolChainId(fromChain);
   const maybeAddress = HUBPOOL_ADDRESSES[hubPoolChainId];
-  if (!isValidAddress(maybeAddress) || maybeAddress === ethers.constants.AddressZero) {
-    throw new Error(
-      `No HubPool supported on ${CHAINS[hubPoolChainId].name} with chainId: ${hubPoolChainId}`
-    );
-  }
-  return HubPool__factory.connect(maybeAddress, signer ?? PROVIDERS[hubPoolChainId]());
+  // if (!isValidAddress(maybeAddress) || maybeAddress === ethers.constants.AddressZero) {
```
Gamaranto, line 58: lets keep the check but add a "fake" value for the hub pool on mainnet imho

src/utils/constants.ts
```diff
@@ -487,3 +487,72 @@ export function makePoolClientConfig(chainId: ChainId): acrossSdk.pool.Config {
 // default to kovan when testing
 export const HUBPOOL_CHAINID = isProduction() ? ChainId.MAINNET : ChainId.KOVAN;
 export const HUBPOOL_CONFIG = makePoolClientConfig(HUBPOOL_CHAINID);
+
+export const PROVIDER_URLS: Record<ChainId, string> = {
```
Gamaranto, line 491: this seems redundant with the `PROVIDERS` object in the `constants` file

src/utils/format.ts
Gamaranto: I agree with the comments, but I suggest we keep the object intact and then invoke the method on the object directly:
```javascript
export const formatNumberTwoSigDigits = new Intl.NumberFormat("en-US", {
  maximumSignificantDigits: 2,
});
// when consuming it in another file
formatNumberTwoSigDigits.format(yourNumber)
```

Which is the same way its used in the [MDN examples](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format)

src/views/Transactions/TransactionsTable/MobileTransactionsTableWithPagination.tsx
```diff
@@ -45,7 +45,9 @@ const MobileTransactionsTableWithPagination: FC<Props> = ({
         title={title}
         openIndex={openIndex}
       />
-      <Pagination onPageChange={setCurrentPage} {...paginateState} />
+      {paginatedRows.length > 25 && (
```
Gamaranto, line 48: lets extract this to a variable, just to make it easier to find / change without having to look at the jsx, something like:
```javascript
const showPagination = paginatedRows.length > 25
```

in the function body

src/views/Transactions/TransactionsTable/createMobileTransactionTableJSX.tsx
```diff
@@ -22,29 +22,41 @@ export interface IMobileRow extends IRow {
   onClick?: () => void;
 }
 
+/*  
+Transfer:
```
Gamaranto, line 26: stray comment

src/views/Transactions/TransactionsTable/createMobileTransactionTableJSX.tsx
```diff
@@ -56,25 +68,29 @@ function formatTransactionRows(
...
     const toChain = (
       <>
         <TableLogo src={toLogo} alt={`${toChainName}_logo`} /> {toChainName}
       </>
     );
 
-    const token = TOKENS_LIST[1].find(
-      (x) => x.address === tx.assetContractAddress
-    );
+    const token = TOKENS_LIST[1].find((x) => x.address === tx.assetAddr);
```
Gamaranto, line 92: why is this always Ethereum Mainnet?

src/views/Transactions/TransactionsTable/createTransactionTableJSX.tsx
```diff
@@ -1,37 +1,58 @@
...
+  capitalizeFirstLetter,
+  formatNumberTwoSigDigits,
+} from "utils/format";
 import { ICell, IRow } from "components/Table/Table";
-import { Transaction } from "./createTransactionModel";
 import { CHAINS, TOKENS_LIST } from "utils/constants";
+import { Transfer } from "@across-protocol/sdk-v2/dist/transfers-history/model";
+import { ChainId } from "utils";
+/* 
+amount: BigNumber {_hex: '0x058d15e176280000', _isBigNumber: true}
```
Gamaranto, line 14: stray comment

src/views/Transactions/useTransactionsView.ts
Gamaranto: My understanding is that the cleanup effect will run after an account change, before the next effect runs  https://github.com/donavon/hook-flow 
So the account will be untracked at each account change, before a subscription to the new account is made

src/state/transferHistory.ts
```diff
@@ -0,0 +1,42 @@
...
+  TransfersUpdated = "TransfersUpdated",
+}
+const { TransfersHistoryClient } = transfersHistory;
+
+const chainConfigs = createTxHistoryClient();
+
+const client = new TransfersHistoryClient({
+  chains: chainConfigs,
+  // optional
+  pollingIntervalSeconds: 60000,
```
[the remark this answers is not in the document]
amateima: it is supposed to be specified in seconds

src/views/Transactions/useTransactionsView.ts
amateima: I wouldn't use `setInterval` for polling the client. I added events in the client to avoid polling and rely only on events notifications

amateima (changes requested):

src/utils/bridge.ts
```diff
@@ -40,19 +45,25 @@ function getHubPoolChainId(sendingChain: ChainId): ChainId {
...
+  //   throw new Error(
+  //     `No HubPool supported on ${CHAINS[hubPoolChainId].name} with chainId: ${hubPoolChainId}`
+  //   );
+  // }
```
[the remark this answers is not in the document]
Tulun: This was a temp fix -- re-added it back in.

src/utils/constants.ts
```diff
@@ -487,3 +487,72 @@ export function makePoolClientConfig(chainId: ChainId): acrossSdk.pool.Config {
...
+// Chains currently in SDK v2
+/*  
+const SPOKE_CHAINS = {
+  [ChainId.ARBITRUM_RINKEBY]: { lowerBoundBlockNumber: 9828565 },
+  [ChainId.KOVAN_OPTIMISM]: { lowerBoundBlockNumber: 0 },
+  [ChainId.RINKEBY]: { lowerBoundBlockNumber: 0 },
+  [ChainId.KOVAN]: { lowerBoundBlockNumber: 0 },
+  [ChainId.MAINNET]: { lowerBoundBlockNumber: 0 },
+};
+*/
```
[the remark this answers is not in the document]
Tulun: Yes documentation.

src/state/transferHistory.ts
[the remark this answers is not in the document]
Tulun: Fixed to 60.

src/views/Transactions/useTransactionsView.ts
Tulun: It's just to expose if an error happens in this particular call -- we can ignore it if you want, but not sure there's any harm being mindful if the SDK bugs out for debugging.

src/state/transferHistory.ts
Tulun: I could combine the chainConfigs in transferHistory -- I just wasn't sure if I was needing to be doing a bunch of formatting logic, as the sdk requires a specific format that's not currently built, but could possibly be combined from existing constants we have here.

src/views/Transactions/useTransactionsView.ts
Tulun: I doubt it would automate anything, as we are waiting for messages for the client. Whether or not we want to store the data in react-query is a different question. Basically, it would just be hard pushing this data into the data store as we aren't using react-query to actually call the API.

src/state/transferHistory.ts
Tulun: Yup!

src/utils/constants.ts
Tulun: Not exactly. The PROVIDERS object doesn't have this information -- I could extend it but I didn't necessarily want to touch the types there. It returns the PROVIDER instance not the URL, which is what the SDK wants.

But I figure it might require more refactoring than what it was worth.

src/views/Transactions/TransactionsTable/createMobileTransactionTableJSX.tsx
Tulun: Should be sourceChain, fixed.

src/views/Transactions/useTransactionsView.ts
Tulun: According to @amateima this is not currently exported -- he'll do it on next release.

src/utils/constants.ts
amateima: @Tulun 
Hey, I can change the sdk to accept the Provider instances instead of the URLs. What do you think? This way we'd have less providers instantiated

src/views/Transactions/TransactionsTable/TransactionsTableWithPagination.tsx
Tulun: Like that better, fixed.

Tulun:
There are a couple bugs in the SDK that @amateima is going to fix -- we chatted about this on slack. He is going to expose the Transfer type and there is an issue in the stopFetching call that prevents it from actually fully stopping the interval.

Should be fixed tomorrow morning.

src/views/Transactions/useTransactionsView.ts
Tulun: Removed interval :)

src/utils/format.ts
Tulun: I'm open to consensus on implementation.

src/utils/bridge.ts
Tulun: Done!

src/utils/constants.ts
Tulun: Sure, this seems reasonable to me.
@mrice32 @daywiss @Gamaranto

daywiss: i think thats fine

src/state/transferHistory.ts
mrice32: Not strongly opinionated! I would just change the name of the function if this is just returning a config!

src/utils/constants.ts
mrice32: SGTM!

src/views/Transactions/useTransactionsView.ts
mrice32: @Gamaranto you're totally correct. I continue to mix this up.

@Tulun sorry for the misleading comment. We should be able to do something simpler here like:

```ts
useEffect(()  => {
  if (account) {
    txHistoryClient.startFetchingTransfers(account).catch((err) => {
      console.error(
          "Error in startFetchingTransfers call in txHistoryClient",
          err
        );
    });
    txHistoryClient.on(
        transfersHistory.TransfersHistoryEvent.TransfersUpdated,
        () => {
          console.log("does this fire", account);
          const nextFilledTx = txHistoryClient.getFilledTransfers(account);
          const nextOngoingTx = txHistoryClient.getPendingTransfers(account);

          setRawFilledTx(nextFilledTx);
          setRawOngoingTx(nextOngoingTx);
        }
      );
    return () => {
      txHistoryClient.stopFetchingTransfers(account);
    }
  }
}, [account]);
```

And we should be able to avoid all the previousAccount logic!

mrice32 (commented):
@Tulun I realize now that I said something wrong in my previous comment. Once that's resolved, I think we should be good to go! Sorry about that!

Tulun:
> [mrice32](https://github.com/mrice32) [2 days ago](https://github.com/across-protocol/frontend-v2/pull/27#discussion_r848819146)
@Gamaranto you're totally correct. I continue to mix this up.

@Tulun sorry for the misleading comment. We should be able to do something simpler here like:

useEffect(()  => {
  if (account) {
    txHistoryClient.startFetchingTransfers(account).catch((err) => {
      console.error(
          "Error in startFetchingTransfers call in txHistoryClient",
          err
        );
    });
    txHistoryClient.on(
        transfersHistory.TransfersHistoryEvent.TransfersUpdated,
        () => {
          console.log("does this fire", account);
          const nextFilledTx = txHistoryClient.getFilledTransfers(account);
          const nextOngoingTx = txHistoryClient.getPendingTransfers(account);

          setRawFilledTx(nextFilledTx);
          setRawOngoingTx(nextOngoingTx);
        }
      );
    return () => {
      txHistoryClient.stopFetchingTransfers(account);
    }
  }
}, [account]);
And we should be able to avoid all the previousAccount logic!


Still experiencing some SDK issues, but I'll try just adding the cleanup func.

src/views/Transactions/useTransactionsView.ts
daywiss: you should update this

daywiss (commented):
there are still some minor outstanding comments but i think its good once resolved

src/utils/constants.ts
Gamaranto: good to me!

Gamaranto (approved):
after other comments are addressed, looks good to me

mrice32 (approved):
LGTM!

src/state/transferHistory.ts
```diff
@@ -0,0 +1,18 @@
...
+  TransfersUpdated = "TransfersUpdated",
+}
+const { TransfersHistoryClient } = transfersHistory;
+
+const chainConfigs = createTxHistoryClient();
+
+const client = new TransfersHistoryClient({
+  chains: chainConfigs,
+  // optional
+  pollingIntervalSeconds: 5,
```
amateima, line 13: I would disable the polling for now until we add a proper Refresh button which will appear/enable when new data is available

amateima (approved):

Tulun:
> there are still some minor outstanding comments but i think its good once resolved

This type doesn't seem exposed yet in the sdk -- non blocking change. gonna kick it down for later.

PR #29, opened 2022-04-08
fix(across-v2): check code before sending

Gamaranto:
Checks there is code at the spoke address to prevent sending/approving to an EOA or the zero address


daywiss (approved):

src/utils/bridge.ts
```diff
@@ -291,6 +291,11 @@ export async function sendAcrossDeposit(
   }: AcrossDepositArgs
 ): Promise<ethers.providers.TransactionResponse> {
   const spokePool = getSpokePool(fromChain);
+  const provider = PROVIDERS[fromChain]();
```
mrice32, line 294: @daywiss @Gamaranto, do y'all think it would be better to build a wrapper for the ethers provider that automatically does this on txn send rather than trying to put these calls everywhere we send a txn?

mrice32 (commented):
One one high level comment!

src/utils/bridge.ts
Gamaranto: yes, I think proxing the sender or the provider would be preferrable, this was more of a super quick fix

mrice32 (approved):

PR #33, opened 2022-04-12
feat: disable deposits and show banner

Tulun:
Screen Shot 2022-04-12 at 12 46 52 PM

To see, add: REACT_APP_DISABLE_DEPOSITS = true to .env .


src/components/PoolForm/AddLiquidityForm.tsx
```diff
@@ -246,13 +253,17 @@ const AddLiquidityForm: FC<Props> = ({
...
+            ((provider && !!disableDeposits) ||
+              !provider ||
+              !!formError ||
+              Number(amount) <= 0) &&
+            isConnected
           }
-          onClick={() =>
-            approveOrPoolTransactionHandler().catch((err) =>
+          onClick={() => {
+            return approveOrPoolTransactionHandler().catch((err) =>
```
daywiss, line 263: why does this need to return

src/utils/constants.ts
```diff
@@ -560,3 +561,4 @@ export function makePoolClientConfig(chainId: ChainId): acrossSdk.pool.Config {
 // FIXME: Switch to Mainnet in prod, when we have a mainnet hub pool
 export const HUBPOOL_CHAINID = isProduction() ? ChainId.KOVAN : ChainId.KOVAN;
 export const HUBPOOL_CONFIG = makePoolClientConfig(HUBPOOL_CHAINID);
+export const disableDeposits = process.env.REACT_APP_DISABLE_DEPOSITS;
```
daywiss, line 564: we havent been good about this, but we should probably add this to readme

daywiss (approved):
minor things but seems good. i would suggest though adding it to readme

src/utils/constants.ts
Tulun: Yeah good call.

src/components/PoolForm/AddLiquidityForm.tsx
Tulun: It doesn't, we just often have a pattern of conditional returns, so I just set it up earlier when I realized this function call (approveOrPoolTransactionHandler) was actually doing 4-5 things.

PR #34, opened 2022-04-13
fix: revert destination address by passing empty value

amateima:
Signed-off-by: amateima <amatei@umaproject.org

Quick PR that allows the user to revert the destination address to the connected wallet  by allowing to pass an empty value

Screenshot 2022-04-13 at 04 19 46

### DIFF
--- a/src/components/AddressSelection/useAddressSelection.ts
+++ b/src/components/AddressSelection/useAddressSelection.ts
@@ -5,7 +5,7 @@ import { useSendForm } from "hooks";
 import { CHAINS_SELECTION, isValidAddress } from "utils";
 
 export default function useAddressSelection() {
-  const { isConnected } = useConnection();
+  const { isConnected, account } = useConnection();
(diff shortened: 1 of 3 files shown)


daywiss (approved):
looks good

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -43,8 +43,12 @@ export default function useAddressSelection() {
...
   const handleSubmit = () => {
-    if (isValid && address) {
-      setToAddress(address);
+    if (isValid) {
+      if (address) {
+        setToAddress(address);
+      } else if (account) {
+        setToAddress(account);
+      }
       toggle();
```
Tulun, line 52: This conditional seems a bit confusing. If the address exists already, you're basically doing an early return, if I understand this correctly.

It might be clearer if it was like:

`if (isValid) {
    if (address) return setToAddress(address);
    if (account) return setToAddress(account);
  }
 `

PR #36, opened 2022-04-14
fix(across-v2): disable send until tx confirms

Gamaranto:
Fixes https://app.shortcut.com/uma-project/story/5238/send-button-is-enabled-while-a-deposit-is-pending

### DIFF
--- a/src/components/SendAction/useSendAction.ts
+++ b/src/components/SendAction/useSendAction.ts
@@ -39,9 +39,7 @@ export default function useSendAction(
       if (hasToApprove) {
         const tx = await approve();
         if (tx) {
-          tx.wait(CONFIRMATIONS).then(() => {
-            setTxPending(false);
-          });
+          tx.wait(CONFIRMATIONS).catch(console.error).finally(() => { setTxPending(false); });
         }
         return tx;
       } else {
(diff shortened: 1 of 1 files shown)


daywiss:
im getting an error when sending
eth/kovan->optimism-kovan
image

I think generally you may want to switch to using .finally() to `setPending` back to false. once this error triggers the button stays pending.

daywiss (commented):
would make it so setPending gets reset back to false on `.finally`

daywiss (approved):

src/components/SendAction/useSendAction.ts
```diff
@@ -59,20 +61,24 @@ export default function useSendAction(
...
+            .finally(() => { setTxPending(false); });
           // TODO: we should invalidate and refetch any queries of the transaction tab, so when a user switches to it, they see the new transaction immediately.
         }
         return tx;
       }
-    } finally {
-      setTxPending(false);
+    } catch (error) {
+      console.error(error);
+      console.error(`Something went wrong sending the transaction`)
```
daywiss, line 72: i think we still need a `setTxPending(false)` here because i keep getting into a state where the sending button doesnt go away

daywiss: basically i tried to send optimism kovan -> kovan, then hit send and closed the metamask dialog. this hit this part of the code and never re-enabled the button

Tulun (approved):

src/components/SendAction/useSendAction.ts
Gamaranto: good catch

PR #37, opened 2022-04-14
feat(across-v2): make order of tokens configurable

Gamaranto:
Makes the order in which the tokens appear in the dropdown configurable

### DIFF
--- a/src/data/token-order.json
+++ b/src/data/token-order.json
@@ -1,0 +1,1 @@
+["ETH", "WETH", "USDC", "DAI", "WBTC", "BOBA", "BADGER", "UMA"]
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -11,6 +11,7 @@ import bobaLogo from "assets/boba-logo.svg";
 import polygonLogo from "assets/polygon-logo.svg";
 import { getAddress } from "./address";
 import rawTokens from "../data/tokens.json";
+import PREFERRED_TOKEN_ORDER from "../data/token-order.json";
 import * as superstruct from "superstruct";
 
 /* Colors and Media Queries section */
@@ -122,6 +123,8 @@ const TokensList = superstruct.array(
 export function isSupportedChainId(chainId: number): chainId is ChainId {
   return chainId in ChainId;
 }
+
+
 function processRawTokens(
   tokens: typeof rawTokens
 ): Record<ChainId, TokenList> {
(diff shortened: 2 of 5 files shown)


src/utils/constants.ts
```diff
@@ -130,10 +133,13 @@ function processRawTokens(
...
     }
-    const tokens = rawTokens.map((token) => ({
-      ...token,
-      logoURI: process.env.PUBLIC_URL + token.logoURI,
-    }));
+    let tokens = [];
+    for (let rawToken of rawTokens) {
+      const token = { ...rawToken, logoURI: process.env.PUBLIC_URL + rawToken.logoURI };
+      tokens[PREFERRED_TOKEN_ORDER.indexOf(token.symbol)] = token;
+    }
```
daywiss, line 140: it would make sense to me to generate this in the constants file at runtime and just export the sorted array so it could be re-used if needed.

daywiss (approved):

src/utils/constants.ts
Tulun: Are we also throwing an error when PUBLIC_URL isn't defined? seems like this could cause an issue. 

We should expose process.env.PUBLIC_URL as a variable as well.

Gamaranto: I think its fine as `PUBLIC_URL` is a variable that cra injects and will always be present https://create-react-app.dev/docs/advanced-configuration/

src/data/token-order.json
```diff
@@ -0,0 +1 @@
+["ETH", "WETH", "USDC", "DAI", "WBTC", "BOBA", "BADGER", "UMA"]
```
mrice32, line 1: What happens if there's a token in here that's not in the tokens.json? What happens if there's a token in the tokens.json that isn't in here?

src/utils/constants.ts
```diff
@@ -130,10 +133,13 @@ function processRawTokens(
     if (!isSupportedChainId(chainId)) {
       throw new Error(`Unsupported chainId: ${chainId}`);
     }
-    const tokens = rawTokens.map((token) => ({
-      ...token,
-      logoURI: process.env.PUBLIC_URL + token.logoURI,
-    }));
+    let tokens = [];
```
mrice32, line 136: Would this not be a slightly better alternative?
```ts
const tokens = PREFERRED_TOKEN_ORDER.map(symbol => {
  const rawToken = rawTokens.find(token => token.symbol === symbol)!;
  // Maybe throw error here if rawToken doesn't exist?
  // if (!rawToken) throw new Error(`${symbol} not found in rawTokens`);
  return { ...rawToken, logoURI: process.env.PUBLIC_URL + rawToken.logoURI };
});
```

The only reason I suggest this is that the edge cases seem easier to manage:
1. If PREFERRED_TOKEN_ORDER doesn't contain a configured token, that token is simply dropped from the output list
2. You can detect and throw if there is no configured token for something in PREFERRED_TOKEN_ORDER.

In the current implementation, 1 would cause something to be placed at the -1 index and 2 would cause indices to be undefined in the output array.

mrice32 (commented):
Looks good! Two minor questions/comments

Tulun (approved):

src/utils/constants.ts
Gamaranto: I like this suggestion , and implemented it with a slight adjustment, instead of throwing if a token in the preferred order is not present (which is fine since not all chains have the all the tokens) I've instead returned undefined and then filter the array

mrice32 (approved):
LGTM!

PR #38, opened 2022-04-17
feat(across-v2): add query params for bridge tab

Gamaranto:
Allows to pre-populate the send tab with the from and to chain from query params:
Example: `http://localhost:3000/bridge?from=288&to=1` would render a send tab from boba to mainnet

### DIFF
--- a/src/hooks/useSendForm.tsx
+++ b/src/hooks/useSendForm.tsx
@@ -17,9 +17,11 @@ import {
   CHAINS,
   filterTokensByDestinationChain,
   getReacheableChains,
+  canBridge,
 } from "utils";
 import { usePrevious } from "hooks";
 import { useConnection } from "state/hooks";
+import { useQueryParams } from "./useQueryParams";
 
 export enum FormStatus {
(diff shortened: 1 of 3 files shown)


daywiss (approved):
looks good to me

src/hooks/useSendForm.tsx
```diff
@@ -267,6 +269,24 @@ type SendFormManagerContext = FormState & {
...
+
+  useEffect(() => {
+    const fromChain = Number(params.from);
+    const toChain = Number(params.to);
+    const areSupportedChains = [fromChain, toChain].every((p) =>
+      CHAINS_SELECTION.includes(Number(p))
+    );
+    if (!areSupportedChains) {
+      return;
+    }
```
Tulun, line 282: I don't think you can do this in an effect -- return in an Effect should be a clean up function.

Gamaranto: It should be fine, if returning `undefined` the clean up function will just not run

PR #42, opened 2022-04-19
feat: add relay and lp fee logic

daywiss:

### DIFF
--- a/src/components/SendAction/useSendAction.ts
+++ b/src/components/SendAction/useSendAction.ts
@@ -39,7 +39,11 @@ export default function useSendAction(
       if (hasToApprove) {
         const tx = await approve();
         if (tx) {
-          tx.wait(CONFIRMATIONS).catch(console.error).finally(() => { setTxPending(false); });
+          tx.wait(CONFIRMATIONS)
+            .catch(console.error)
+            .finally(() => {
+              setTxPending(false);
+            });
         }
         return tx;
       } else {
@@ -62,7 +66,9 @@ export default function useSendAction(
               });
             })
             .catch(console.error)
-            .finally(() => { setTxPending(false); });
+            .finally(() => {
+              setTxPending(false);
+            });
           // TODO: we should invalidate and refetch any queries of the transaction tab, so when a user switches to it, they see the new transaction immediately.
         }
         return tx;
--- a/src/components/SendAction/useSendAction.ts
+++ b/src/components/SendAction/useSendAction.ts
@@ -6,13 +6,13 @@ import { useConnection } from "state/hooks";
 
 type TokenInfo =
   | {
-    address: string;
-    symbol: "WETH";
-    name: "Wrapped Ether";
-    decimals: 18;
-    logoURI: string;
-    bridgePool: string;
-  }
+      address: string;
+      symbol: "WETH";
+      name: "Wrapped Ether";
+      decimals: 18;
+      logoURI: string;
+      bridgePool: string;
+    }
   | Token;
 
 export default function useSendAction(
@@ -26,7 +26,7 @@ export default function useSendAction(
   const tokenInfo = TOKENS_LIST[fromChain].find(
     (t) => t.address === token
   ) as TokenInfo;
-  const { fees } = useBridgeFees(amount, fromChain, tokenInfo?.symbol);
+  const { fees } = useBridgeFees(amount, toChain, tokenInfo?.symbol);
   const { status, hasToApprove, send, approve } = useBridge();
   const { account } = useConnection();
 
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -35,9 +35,9 @@ export function useBalance(
   const queryKey = enabledQuery
     ? balanceQueryKey(chainIdToQuery, token, accountToQuery, blockNumberToQuery)
     : [
-      "DISABLED_BALANCE_QUERY",
-      { chainIdToQuery, token, accountToQuery, blockNumberToQuery },
-    ];
+        "DISABLED_BALANCE_QUERY",
+        { chainIdToQuery, token, accountToQuery, blockNumberToQuery },
+      ];
   const { data: balance, ...delegated } = useQuery(
     queryKey,
     async () => {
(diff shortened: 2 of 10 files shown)


src/hooks/useBridgeFees.ts
```diff
@@ -7,19 +7,19 @@ import { useBlock } from "./useBlock";
...
  */
 export function useBridgeFees(
   amount: ethers.BigNumber,
-  fromChain: ChainId,
+  toChainId: ChainId,
   tokenSymbol?: string
 ) {
-  const { block } = useBlock(fromChain);
-  const enabledQuery = !!fromChain && !!block && !!tokenSymbol && amount.gt(0);
+  const { block } = useBlock(toChainId);
```
daywiss, line 19: hopefully we dont care too much if the block time is coming from the "fromChain" or "toChain"

src/utils/bridge.ts
```diff
@@ -118,19 +121,52 @@ export async function getLpFee(
...
+    rateModelStoreAddress
+  );
+  result.pct = await lpFeeCalculator.getLpFeePct(
+    tokenAddress,
+    hubPoolAddress,
+    amount,
+    blockTime
+  );
   result.total = amount.mul(result.pct).div(parseEther("1"));
+  result.isLiquidityInsufficient = false;
```
daywiss, line 161: no way afaik that we can calculate this?

src/utils/bridge.ts
```diff
@@ -350,3 +402,127 @@ export async function sendAcrossApproval(
   const tokenContract = clients.erc20.connect(token, signer);
   return tokenContract.approve(spokePool.address, amount);
 }
+
+const { exists } = utils;
+const { parseAndReturnRateModelFromString } = across.rateModel;
+const { calculateRealizedLpFeePct } = lpFeeCalculator;
+const { rateModelStore } = clients;
+
+export default class LpFeeCalculator {
```
daywiss, line 411: this is adapted from the protocol/sdk. its slightly different to acomodate new hubpool contract

src/utils/bridge.ts
```diff
@@ -350,3 +402,127 @@ export async function sendAcrossApproval(
...
+    // a JSON object.
+    const rateModel = parseAndReturnRateModelFromString(
+      rateModelForBlockHeight
+    );
+
+    return calculateRealizedLpFeePct(rateModel, currentUt, nextUt);
+  }
+}
+
+class Queries implements relayFeeCalculator.QueryInterface {
```
daywiss, line 471: each chain may need a different version of this class, it implements an interface that gets passed into the relayFeeCalculator class

src/hooks/useBridge.ts
```diff
@@ -134,11 +132,11 @@ type ComputeStatusArgs = {
...
-  | {
-    isLiquidityInsufficient: boolean;
-    isAmountTooLow: boolean;
-  }
-  | undefined;
+    | {
+        isLiquidityInsufficient: boolean;
+        isAmountTooLow: boolean;
+      }
+    | undefined;
```
Tulun, line 139: I know this is a linter, but would this be worth abstracting the type out?

src/hooks/useBridgeFees.ts
Tulun: Wouldn't this matter though? I thought we had issues before when the block times were wildly different from L2 -> L1.

src/utils/bridge.ts
```diff
@@ -350,3 +402,127 @@ export async function sendAcrossApproval(
...
+    return info;
+  }
+  getTokenInfo(tokenSymbol: string) {
+    const info = this.tokens.find((x) => x.symbol === tokenSymbol);
+    assert(info, `No token found in tochain for ${tokenSymbol}`);
+    return info;
+  }
+  async getTokenPrice(tokenSymbol: string): Promise<string | number> {
+    console.log({ tokenSymbol });
+    if (tokenSymbol.toLowerCase() === "eth") return 1;
```
Tulun, line 494: Stray log.

src/utils/bridge.ts
```diff
@@ -350,3 +402,127 @@ export async function sendAcrossApproval(
...
+      hubPoolInstance.callStatic.liquidityUtilizationPostRelay(
+        tokenAddress,
+        amount,
+        { blockTag } as any
+      ),
+      rateModelStoreInstance.callStatic.l1TokenRateModels(tokenAddress, {
+        blockTag,
+      } as any),
+    ]);
+
```
Tulun, line 460: What happens if any of these promises fail?

src/hooks/useBridgeFees.ts
daywiss: theres a pretty big window for variation now, like 10 minutes or something

src/hooks/useBridge.ts
daywiss: sure

src/utils/bridge.ts
daywiss: this gets wrapped in  a try catch in `getBridgeFees` function call. In production the error will propogate up to react query where it will end up just not showing a fee. in development the error will be logged and return some 0 values.

src/components/SendAction/useSendAction.ts
```diff
@@ -6,13 +6,13 @@ import { useConnection } from "state/hooks";
...
-    logoURI: string;
-    bridgePool: string;
-  }
+      address: string;
+      symbol: "WETH";
+      name: "Wrapped Ether";
+      decimals: 18;
+      logoURI: string;
+      bridgePool: string;
+    }
```
daywiss, line 15: i dont understand why the linter is constantly changing this

src/hooks/useBalance.ts
```diff
@@ -88,20 +88,23 @@ export function useBalances(
...
+      ];
   const prevAccount = usePrevious(accountToQuery);
   const prevChain = usePrevious(chainIdToQuery);
   const prevTokens = usePrevious(tokens);
   // Keep the previous data only when blockNumberToQuery changes.
-  const keepPreviousData = prevAccount === accountToQuery && prevChain === chainIdToQuery && JSON.stringify(prevTokens) === JSON.stringify(tokens);
+  const keepPreviousData =
+    prevAccount === accountToQuery &&
+    prevChain === chainIdToQuery &&
+    JSON.stringify(prevTokens) === JSON.stringify(tokens);
```
daywiss, line 107: again this is all linter changes

src/utils/bridge.ts
```diff
@@ -118,19 +121,52 @@ export async function getLpFee(
   amount: ethers.BigNumber,
   blockTime?: number
 ): Promise<Fee & { isLiquidityInsufficient: boolean }> {
+  // eth and weth can be treated the sasme same in this case, but the rate model only currently supports weth address
```
mrice32, line 124:
```suggestion
  // eth and weth can be treated the same in this case, but the rate model only currently supports weth address
```

src/utils/bridge.ts
mrice32: Sorry, I'm confused, why isn't this pulled from the SDK? Wasn't that the original intention?

mrice32 (approved):
Looks great! Only minor comments

Tulun (approved):

src/utils/bridge.ts
daywiss: if thats the case then yea, i just wasnt sure how to do this.  For some reason i thought some of these stats were no longer available in the hubpool contract.  I'll make a new issue for this and get it in a new pr.

src/utils/bridge.ts
daywiss: The problem is that this requires lookups for token symbols to addresses on various chains which are not available in the SDK and defined in the constants file on the frontend.  This makes it really awkward to try and have to do this stuff in the sdk.  Theres no easy way to include this logic in the sdk without duplicating all the constants for tokens defined in the frontend.

PR #43, opened 2022-04-19
feat: empty and loading states for tx history

Tulun:
Empty and loading states for tx history: 
Screen Shot 2022-04-19 at 10 54 02 AM
Screen Shot 2022-04-19 at 11 00 30 AM

### DIFF
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -66,6 +66,7 @@ const Transactions = () => {
     [rawOngoingTx, setOpenOngoingRow]
   );
 
+  const isTxPresent = !filledTx.length && !ongoingTx.length && !initialLoading;
   return (
     <Wrapper>
       {(ongoingTx.length === 0 || !isConnected) && (
@@ -124,7 +125,7 @@ const Transactions = () => {
                 <div>Loading...</div>
               </LoadingWrapper>
             )}
-            {!filledTx.length && !ongoingTx.length && !initialLoading && (
+            {isTxPresent && (
               <NotFoundWrapper>
                 <img src={emptyClouds} alt="empty_state" />
                 No transactions found.
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -37,6 +37,7 @@ const Transactions = () => {
     setCurrentPage,
     rawFilledTx,
     rawOngoingTx,
+    initialLoading,
   } = useTransactionsView();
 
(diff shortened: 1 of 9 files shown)


src/views/Transactions/Transactions.tsx
```diff
@@ -107,6 +114,22 @@ const Transactions = () => {
...
+              <LoadingWrapper>
+                <FontAwesomeIcon
+                  icon={faCircleNotch}
+                  className="fa-spin"
+                  size="2x"
+                />
+                <div>Loading...</div>
+              </LoadingWrapper>
+            )}
+            {!filledTx.length && !ongoingTx.length && !initialLoading && (
```
Gamaranto, line 127: nit: we can abstract this into its own variable like `isTxPresent`

Gamaranto (approved):
just a nit, otherwise lgtm

src/views/Transactions/TransactionsTable/MobileTransactionsTable.tsx
```diff
@@ -30,84 +31,82 @@ const MobileTransactionsTable: FC<Props> = ({
...
-                        <div>{row.txHash}</div>
-                      </AccordionRow>
-                    </AccordionWrapper>
-                  )}
-                </>
-              );
-            })}
-          </StyledBody>
-        </StyledTableWrapper>
+      {rows.length && !initialLoading ? (
```
daywiss, line 38: imo the control of hiding/showing a component should happen in the parent unless its not being hidden fully

src/views/Transactions/TransactionsTable/TransactionsTable.tsx
```diff
@@ -17,55 +17,58 @@ interface Props {
...
-                          </StyledCell>
-                        );
-                      })}
-                    </StyledRow>
-                  );
-                })
-              : null}
-          </StyledBody>
-        </StyledTableWrapper>
+      {rows.length && !initialLoading ? (
```
daywiss, line 31: same here, it seems like the parent would be a better spot to hide the component

daywiss (approved):
looks good, id just address francescos comment

src/views/Transactions/TransactionsTable/TransactionsTable.tsx
Tulun: Fixed!

src/views/Transactions/Transactions.tsx
Tulun: Fair, abstracted.

PR #44, opened 2022-04-19
fix: set eth address back to zero, update user/pool indexing

daywiss:

### DIFF
--- a/src/components/PoolForm/PoolForm.tsx
+++ b/src/components/PoolForm/PoolForm.tsx
@@ -4,6 +4,7 @@ import Tabs from "../Tabs";
 import AddLiquidityForm from "./AddLiquidityForm";
 import RemoveLiquidityForm from "./RemoveLiquidityForm";
 import { QuerySubState } from "@reduxjs/toolkit/dist/query/core/apiState";
+import { poolClient } from "state/poolsApi";
 import {
   Wrapper,
   Info,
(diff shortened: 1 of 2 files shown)


src/components/PoolForm/PoolForm.tsx
```diff
@@ -89,23 +90,24 @@ const PoolForm: FC<Props> = ({
   useEffect(() => {
     if (!signer || !isConnected || symbol !== "ETH") return;
 
-    estimateGasForAddEthLiquidity(signer, tokenAddress)
+    const wethAddress = poolClient.config.wethAddress;
```
daywiss, line 93: we need the weth address configured in the pool in order to call the hubpool contract correctly for eth

src/views/Pool.tsx
```diff
@@ -38,17 +39,27 @@ const Pool: FC = () => {
   const [depositUrl, setDepositUrl] = useState("");
   const [loadingPoolState, setLoadingPoolState] = useState(false);
   const [defaultTab, setDefaultTab] = useState("Add");
-  const pool = useAppSelector((state) => state.pools.pools[token.address]);
+  const pool = useAppSelector(
+    (state) =>
+      state.pools.pools[
+        token.address === AddressZero
+          ? poolClient.config.wethAddress
+          : token.address
```
daywiss, line 47: this is really annoying, but basically every place we have address zero, we have to manually replace with weth address since the pool client cannot handle address zero ( it was more complicated to handle that case in the pool client than in the frontend)

Tulun (approved):

PR #48, opened 2022-04-22
feat: Banner notification

Tulun:
Story details: https://app.shortcut.com/uma-project/story/5110
Screen Shot 2022-04-22 at 2 38 19 PM

### DIFF
--- a/src/views/Pool.tsx
+++ b/src/views/Pool.tsx
@@ -23,6 +23,7 @@ import BouncingDotsLoader from "components/BouncingDotsLoader";
 
 import { BounceType } from "components/BouncingDotsLoader/BouncingDotsLoader";
 import { SuperHeader } from "components";
+import { migrationPoolV2Warning } from "utils";
 
 export type ShowSuccess = "deposit" | "withdraw";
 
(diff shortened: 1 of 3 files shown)


src/views/Pool.tsx
```diff
@@ -114,6 +115,14 @@ const Pool: FC = () => {
       )}
       {!showSuccess ? (
         <Wrapper>
+          {migrationPoolV2Warning && (
+            <MigrationWarning>
+              <div>
+                You still have liquidity on v1 and some more text plus link to
+                these instructions
```
daywiss, line 122: clearly this text isnt right, we should wait until we have the right copy. also since this is going to show for everyone we should change the language a bit to be more of a suggestion to check v1 if you have liquidity.

Tommy1231232: In shortcut the verbiage was: "If you have not migrated liquidity from Across v1 to Across v2, please follow these instructions "

Tommy1231232 (changes requested):
Update verbiage per comment

src/views/Pool.tsx
Tulun: @Tommy1231232 verbiage updated. just lmk when the link is ready.

Tommy1231232: @Tulun I'm not seeing the banner on the pool tab. Not sure why I tried mainnet and kovan

Tulun: @Tommy1231232 : you need to set REACT_APP_MIGRATION_POOL_V2_WARNING in your .env file to "true". 

We can do it on vercel if you really want.

Tommy1231232: @Tulun can you share a screenshot of the view you currently see?

daywiss (approved):
just need a card to remember to add link later

mrice32 (approved):

Tulun:
@mrice32 still waiting on a link.

PR #54, opened 2022-04-29
feat: add ability to discover all token routes from contracts

daywiss:

Adds  a function to discover all supported tokens/spoke pool routes based on a hub pool address. This can be used later to configure the app more concisely and with less errors


src/utils/ethers.ts
```diff
@@ -0,0 +1,26 @@
...
+export type GetEventType<
+  ContractType extends Contract,
+  EventName extends string
+> = ReturnType<
+  ContractType["filters"][EventName] extends Callable
+    ? ContractType["filters"][EventName]
+    : never
+> extends TypedEventFilter<infer T, infer S>
+  ? TypedEvent<T & S extends Result ? T & S : any>
+  : never;
```
daywiss, line 26: this is copied from the sdk. it was failing to build when imported from sdk because of differing ethers versions.

src/utils/tokenRoutes.ts
```diff
@@ -0,0 +1,236 @@
...
+  hubPoolChain: number;
+  fromChain: number;
+  toChain: number;
+  tokenAddress: string;
+  spokeAddress: string;
+  symbol: string;
+}
+export type Routes = Route[];
+
+export class SpokePoolUtils {
```
daywiss, line 42: these utility classes both basically load all events and then provide query functions to get information based on those events.

src/utils/tokenRoutes.ts
```diff
@@ -0,0 +1,236 @@
...
+export interface Route {
+  hubPoolAddress: string;
+  hubPoolChain: number;
+  fromChain: number;
+  toChain: number;
+  tokenAddress: string;
+  spokeAddress: string;
+  symbol: string;
+}
+export type Routes = Route[];
```
daywiss, line 40: these routes are ultimately what we want to get out of this code. it will list all available token routes between enabled chains

PR #56, opened 2022-04-29
feat: add ability to discover all token routes from contracts

daywiss:

Adds a function to discover all supported tokens/spoke pool routes based on a hub pool address. This can be used later to configure the app more concisely and with less errors

example result from kovan
image

### DIFF
--- a/src/utils/ethers.ts
+++ b/src/utils/ethers.ts
@@ -1,4 +1,5 @@
 import { Contract, ethers } from "ethers";
+import type { Event } from "ethers";
 import type {
   TypedEventFilterEthers as TypedEventFilter,
   TypedEventEthers as TypedEvent,
@@ -6,7 +7,7 @@ import type {
 import { Provider } from "@ethersproject/providers";
 import { Signer } from "@ethersproject/abstract-signer";
 
-export { Provider, Signer, Contract, TypedEventFilter, TypedEvent };
+export { Provider, Signer, Contract, TypedEventFilter, TypedEvent, Event };
 
 export type Result = ethers.utils.Result;
 
--- a/src/utils/ethers.ts
+++ b/src/utils/ethers.ts
@@ -1,0 +1,26 @@
+import { Contract, ethers } from "ethers";
+import type {
+  TypedEventFilterEthers as TypedEventFilter,
+  TypedEventEthers as TypedEvent,
+} from "@uma/contracts-frontend";
+import { Provider } from "@ethersproject/providers";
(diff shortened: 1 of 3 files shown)


src/utils/ethers.ts
```diff
@@ -0,0 +1,26 @@
...
+export type GetEventType<
+  ContractType extends Contract,
+  EventName extends string
+> = ReturnType<
+  ContractType["filters"][EventName] extends Callable
+    ? ContractType["filters"][EventName]
+    : never
+> extends TypedEventFilter<infer T, infer S>
+  ? TypedEvent<T & S extends Result ? T & S : any>
+  : never;
```
daywiss, line 26: this was all copied from uma protocol sdk. was having issues building due to different ethers versions when importing `GetEventType` directly from sdk. wanted to avoid dealing with that so copied this code in

src/utils/tokenRoutes.ts
```diff
@@ -0,0 +1,236 @@
...
+export interface Route {
+  hubPoolAddress: string;
+  hubPoolChain: number;
+  fromChain: number;
+  toChain: number;
+  tokenAddress: string;
+  spokeAddress: string;
+  symbol: string;
+}
+export type Routes = Route[];
```
daywiss, line 40: this is the output we want from all these queries. basically a list of all routes between chains and tokens which we can use later to inform the app

src/utils/tokenRoutes.ts
```diff
@@ -0,0 +1,236 @@
...
+
+export class SpokePoolUtils {
```
daywiss, line 42: there are 2 classes here which work similiarly, hubpool and spokepool utilities fetch all relevent events, then provide some basic queries which iterate through those events to get data

src/utils/tokenRoutes.ts
```diff
@@ -0,0 +1,236 @@
...
+        (this.contract.filters as any).EnabledDepositRoute()
+      ),
+    ];
+    this.events = (await Promise.all(queries)).flat();
+    return this.events;
+  }
+  routesEnabled(): Record<string, string[]> {
+    const init: Record<string, Set<string>> = {};
+    const result = this.events.reduce((result, event) => {
+      if (event.event !== "EnabledDepositRoute") return result;
```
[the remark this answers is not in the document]
daywiss: hmm ok i can try that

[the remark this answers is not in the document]
daywiss: works well

PR #58, opened 2022-05-04
improve: refactor configuration and add config client

daywiss:

This fixes issues around the app configuration to make it more robust:

1. Configuration of apps available routes are specified through a route file which can be generated from on chain data, rather than hard coded in constants.
2. Bridge tab now reflects only the active chains available for the current hub pool
3. Constants are no longer accessed directly, but rather use helper functions to return data
4. Much of the constants data was redundant, now it only contains the metadata needed to run the app
5. Common constant queries are collected into a configuration class which can be imported anywhere in the app.
6. Tokens are no longer referenced by addressed, but by symbol and chain, this is much more robust when dealing with native tokens vs erc20s
7. Native tokens are now explicitly marked in the routes, and can be checked through the configuration class
8. dropdowns in send tab can now potentially be undefined, allowing us to specify non default data, which we couldnt before
9. Send form is more robust, and filters routes to find acceptable paths between chains, this ensures we never accidentally send a bad combination
10. send form reducer is more robust, and could potentially allow us to select a route based on changing any dropdown in any order, though current behavior has been respected.

The main changes to the app are the utility/constants file and the addition of the utility/configuration

# QA
The most important things to test here are the send page primarily, and the pool page secondarily. We are not looking for design issues necessarily, but more about functionality when bridging between chains and adding/removing liquidity.  Important to note that your transfers will not yet arrive on the next chain. Leave comments / screenshots in GH.


src/utils/bridge.ts
```diff
@@ -9,84 +9,19 @@ import {
...
-  TokenList,
+  ChainId,
+  hubPoolChainId,
+  hubPoolAddress,
+  getToken,
+  getProvider,
+  getRateModelAddress,
 } from "./constants";
 
-import { isValidString, parseEther, tagAddress } from "./format";
```
daywiss, line 36: you can see this file has been substantially simplified due to config changes

src/utils/token.ts
```diff
@@ -55,18 +68,21 @@ export function getBalances(
...
-  blockNumber?: number
+  tokenSymbol: string,
+  blockNumber: number | "latest" = "latest"
 ): Promise<ethers.BigNumber> {
-  const provider = PROVIDERS[chainId]();
-  // For ETH, allowance does not make sense
-  if (token === CHAINS[chainId].nativeCurrencyAddress) {
+  const provider = getProvider(chainId);
+  const config = getConfig();
+  const { isNative, address } = config.getTokenInfoBySymbol(
```
daywiss, line 78: we now have this `isNative` property on tokens which fixes a lot of issues with the app

src/views/Confirmation/Confirmation.tsx
```diff
@@ -46,30 +43,30 @@ type Props = {
...
-    (t) => t.address === deposit.token
+  const fromTokenInfo = config.getTokenInfoByAddress(
+    deposit.fromChain,
+    deposit.tokenAddress
   );
-  const isWETH = tokenInfo?.symbol === "WETH";
+  const ethInfo = getToken("ETH");
+  const fromChainInfo = getChainInfo(deposit.fromChain);
+  const toChainInfo = getChainInfo(deposit.toChain);
+  const isWETH = fromTokenInfo?.symbol === "WETH";
```
daywiss, line 55: this needs to be re-thought

src/utils/constants.ts
```diff
@@ -77,136 +87,61 @@ export const COLORS = {
...
 const defaultConstructExplorerLink =
   (explorerUrl: string) => (txHash: string) =>
     `${explorerUrl}/tx/${txHash}`;
 
-export const isProduction = () =>
-  process.env.REACT_APP_VERCEL_GIT_COMMIT_REF === "master";
-
-export const CHAINS: Record<ChainId, ChainInfo> = {
-  [ChainId.MAINNET]: {
+export const chainInfoList: ChainInfoList = [
```
daywiss, line 124: the chains are a list  now, which prioritizing ordering

src/utils/constants.ts
```diff
@@ -343,282 +238,266 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
...
+  logoURI: string;
+  // tokens require a mainnet address to do price lookups on coingecko, not used for anything else.
+  mainnetAddress?: string;
 };
+// enforce weth to be first so we can use it as a guarantee in other parts of the app
+export type TokenInfoList = TokenInfo[];
 
-export const DEFAULT_FROM_CHAIN_ID = ChainId.OPTIMISM;
-export const DEFAULT_TO_CHAIN_ID = ChainId.MAINNET;
+export const tokenList: TokenInfoList = [
```
daywiss, line 262: this now only contains metadata for each token, the symbols should be unique

src/utils/constants.ts
```diff
@@ -343,282 +238,266 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
...
+    name: "Ether",
+    symbol: "ARETH",
+    decimals: 18,
+    logoURI: "/logos/ethereum-logo.svg",
+    mainnetAddress: getAddress("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"),
+  },
+];
 
-/* Onboard config */
+export const lowerBounds = {
```
daywiss, line 362: these are block number lowerbounds used by transfer client

src/utils/constants.ts
```diff
@@ -343,282 +238,266 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
...
-    {
-      chainId: ChainId.KOVAN,
-      provider: PROVIDERS[ChainId.KOVAN](),
-      spokePoolContractAddr: SPOKE_ADDRESSES[ChainId.KOVAN],
-      lowerBoundBlockNumber: 30475937,
-    },
-  ];
 
-  return txHistoryConfigs;
+const Route = superstruct.object({
```
daywiss, line 476: superstruct lets us validate the data in the json files. app will fail to boot if anything is malformed.

src/utils/constants.ts
```diff
@@ -343,282 +238,266 @@ export const CHAINS: Record<ChainId, ChainInfo> = {
...
+  if (chainId === ChainId.KOVAN) {
+    superstruct.assert(KovanRoutes, RouteConfig);
+    return KovanRoutes;
+  }
+  throw new Error("No routes defined for chainId: " + chainId);
 }
 
-export const enableReactQueryDevTools =
-  process.env.REACT_APP_ENABLE_REACT_QUERY_DEV_TOOLS;
+export const routeConfig = getRoutes(hubPoolChainId);
```
daywiss, line 502: this is used to configure the config class

src/hooks/useSendForm.tsx
```diff
@@ -32,20 +32,16 @@ export enum FormStatus {
 
 type FormState = {
   status: FormStatus;
-  token: string;
   amount: ethers.BigNumber;
-  toChain: ChainId;
-  fromChain: ChainId;
+  tokenSymbol?: string;
+  toChain?: ChainId;
+  fromChain?: ChainId;
```
daywiss, line 38: you can see these are optional now, which means they could be undefined. this controls the send form, so that meants dropdowns can be set to empty rather than defaulting to a chain

src/hooks/useAllowance.ts
```diff
@@ -14,43 +15,52 @@ import { useBlock } from "./useBlock";
  * @returns The allowance of tokens granted by the `owner` to the `spender` and the and the UseQueryResult object
  */
 export function useAllowance(
-  token: string,
+  tokenSymbol?: string,
   chainId?: ChainId,
   owner?: string,
   spender?: string,
   blockNumber?: number
 ) {
```
Tulun, line 23: May seem like a dumb question but what happens if all the args here are empty? Seems odd this has no required values.

src/hooks/useBalance.ts
```diff
@@ -18,36 +21,56 @@ import { usePrevious } from "hooks";
...
     queryKey,
     async () => {
-      const balance = await getBalance(
-        chainId!,
-        token,
-        accountToQuery!,
-        blockNumberToQuery
+      if (!chainIdToQuery) return BigNumber.from(0);
+      if (!tokenSymbolToQuery) return BigNumber.from(0);
+      const tokenInfo = config.getTokenInfoBySymbol(
```
Tulun, line 60: Combine these 2 IF statements

src/hooks/useBalance.ts
```diff
@@ -68,53 +91,51 @@ export function useBalance(
...
   const accountToQuery = account ?? connectedAccount;
   const { block: latestBlock } = useBlock(chainId);
   const blockNumberToQuery = blockNumber ?? latestBlock?.number;
   // To fetch balances, we need a list of tokens, an account to get balances of, and a specified chainId.
   const enabledQuery =
     !!chainIdToQuery &&
     !!accountToQuery &&
-    tokens.length > 0 &&
+    tokenSymbols.length > 0 &&
     !!blockNumberToQuery;
```
Tulun, line 110: Shouldn't need to specify > 0

src/hooks/useBridge.ts
```diff
@@ -72,20 +74,28 @@ export function useBridge() {
...
-    if (!signer || status !== "ready" || !toAddress || !block || !fees) {
+    if (
+      !signer ||
+      status !== "ready" ||
+      !toAddress ||
+      !block ||
+      !fees ||
+      !selectedRoute
+    ) {
       return;
```
Tulun, line 89: Might be worth putting this in a variable.

Tulun (commented):
First pass.

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -53,74 +44,83 @@ const AddressSelection: React.FC = () => {
     clearInput,
     open,
     address,
+    availableToChains,
+    selectedToChainInfo,
   } = useAddressSelection();
```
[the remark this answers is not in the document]
daywiss: yea this is a bit wierd. its the same shape as the ChainInfo type defined in the constants file, has chain metadata. Its provided to prevent this component from needing to look up the selected chain info.

src/hooks/useAllowance.ts
daywiss: this was needed so that we could add undefined selections to the from chain and token selection. This and getting balance both just return 0 now if not supplied, which i think is ok.

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -106,14 +97,14 @@ export default function useCoinSelection() {
         );
       }
     }
```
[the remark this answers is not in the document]
daywiss: i could look into that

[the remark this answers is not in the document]
daywiss: was able to simplify it somewhat

src/hooks/useBalance.ts
daywiss: hmmm ok

src/hooks/useSendForm.tsx
```diff
@@ -82,7 +78,44 @@ type Action =
...
+    fromTokenSymbol: tokenSymbol,
+  });
+
+  // we have from, to, symbols that produce no routes, so reset the form to the first valid route
+  if (!availableRoutes.length) {
+    availableRoutes = config.filterRoutes({ fromTokenSymbol: tokenSymbol });
+    const [firstRoute] = availableRoutes;
+    fromChain = firstRoute.fromChain;
+    toChain = firstRoute.toChain;
+  }
```
Tulun, line 115: Just to clarify here, a route is just the idea of a chain to chain interaction we support?

Like we might support Boba -> Optimism but not say Optimism -> Arbitrum?

src/hooks/useSendForm.tsx
```diff
@@ -82,7 +78,44 @@ type Action =
...
+  const selectedRoute =
+    availableRoutes.length === 1 ? availableRoutes[0] : undefined;
+
```
Tulun, line 118: maybe better name as defaultRoute?

src/state/transferHistory.ts
```diff
@@ -1,16 +1,48 @@
...
-const client = new TransfersHistoryClient({
-  chains: chainConfigs,
-});
-// optional
-client.setLogLevel("debug");
+let client: transfersHistory.TransfersHistoryClient | undefined;
 
-export default client;
+export default function getClient() {
+  if (client) return client;
```
Tulun, line 36: Isn't this function actually setting the client?

src/utils/chains.ts
```diff
@@ -1,45 +1,47 @@
...
   try {
     await provider.send("wallet_switchEthereumChain", [
       {
         chainId: ethers.utils.hexValue(chainId),
       },
     ]);
   } catch (switchError: any) {
-    if (switchError.code === 4902) {
+    if (switchError.code === 4902 || switchError.code === -32603) {
       try {
```
Tulun, line 19: I'd probably add some comments to this. What the heck is 4902 or -32603.

Tulun (approved):
Mostly just nits.

src/hooks/useSendForm.tsx
daywiss: yea exactly. the app is configured with a list of routes, ie paths from chain to chain by token symbol.  There was a PR while you were out which could generate this from on chain queries.

src/hooks/useSendForm.tsx
daywiss: its not actually the default route, as the user is selecting values from the form ( from chain, to chain and token), the routes get filtered until there is just 1. thats considered the users selected route.  Each of these reducers handle a change from each element in the form ( change from chain, change to chain, change token), so the logic is pretty similar for each reducer.

src/state/transferHistory.ts
daywiss: yea its true, its not just a read, its initializing it once on read, caching it then reading for all calls afterwards.  This could theoretically be changed back since all calls are syncronous, but with this type of configuration we set the stage for a potentially async instantiation of the app ( if we wanted to dyanmically query routes for the app on load).  dont really need to go into all the details but if u are curious we can talk on slack

src/utils/chains.ts
daywiss: ok sure, maybe theres a better way to do this also

src/utils/config.ts
```diff
@@ -0,0 +1,167 @@
...
+  constructor(private config: constants.RouteConfig) {
+    this.init();
+  }
+  private init(): void {
+    this.config.routes.forEach((route) => {
+      this.spokeAddresses[route.fromChain] = route.fromSpokeAddress;
+      this.spokeChains.add(route.fromChain);
+      this.spokeChains.add(route.toChain);
+    });
+  }
```
mrice32, line 31: Why not just inline this in the constructor?

src/utils/config.ts
```diff
@@ -0,0 +1,167 @@
...
+  getHubPoolChainId(): constants.ChainId {
+    return this.config.hubPoolChain;
+  }
+  getHubPoolAddress(): string {
+    return this.config.hubPoolAddress;
+  }
+  getL1TokenAddressBySymbol(symbol: string) {
+    // all routes have an l1Token address, so just find the first symbol that matches
+    const route = this.config.routes.find((x) => x.fromTokenSymbol);
```
mrice32, line 53: Does this work? Don't you need to test `x.fromTokenSymbol === symbol`?

src/utils/config.ts
```diff
@@ -0,0 +1,167 @@
...
+    return result;
+  }
+  getSpokeChainIds(): constants.ChainId[] {
+    return this.getSpokeChains()
+      .map((chain) => chain.chainId)
+      .filter(constants.isSupportedChainId);
+  }
+  isSupportedChainId(chainId: number): boolean {
+    if (!constants.isSupportedChainId(chainId)) return false;
+    return this.spokeChains.has(chainId);
```
mrice32, line 82: nit: a little simpler
```suggestion
    return constants.isSupportedChainId(chainId) && this.spokeChains.has(chainId);
```

src/utils/config.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+  canBridge(fromChain: number, toChain: number): boolean {
+    const routes = this.filterRoutes({ fromChain, toChain });
+    return routes.length > 0;
+  }
+  getReachableChainIds(fromChain: number): number[] {
+    const routes = this.filterRoutes({ fromChain });
+    const result = new Set<number>();
+    routes.forEach((route) => result.add(route.toChain));
+    return [...result.values()].map(Number);
```
mrice32, line 145: nit: why do you need to call Number on the values if the values are already numbers (isn't that the type of the set)?

Why not something like this?

```ts
return new Set<number>(routes.map(({ toChain }) => toChain)).values();
```

src/utils/config.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+  getReachableChains(fromChain: number): constants.ChainInfoList {
+    const routes = this.filterRoutes({ fromChain });
+    const result = new Set<number>();
+    routes.forEach((route) => result.add(route.toChain));
+    return [...result.values()].map(constants.getChainInfo);
```
mrice32, line 151: Same question here.

mrice32 (approved):
LGTM! Only minor comments

src/utils/config.ts
daywiss: definately could, its simple enough, previously i thought it might have been async

src/utils/config.ts
daywiss: nice catch!

src/utils/config.ts
daywiss: good call, very nice

src/utils/config.ts
daywiss: turns out both these functions arent used, will remove them

jespergisslen:
Some UI/UX comments + image to explain thoughts

- Line break
- ETH as default
- Fees section
- Send to overlay

feedback-jesper

PR #59, opened 2022-05-05
fix: update About verbiage

amateima:


daywiss (approved):
probalby want OK from jesper or korpi

PR #60, opened 2022-05-06
feat: add mainnet routes

daywiss:

This adds configuration for mainnet, fixes some bugs that came up with this. To test this view the mainnet deployment (frontend-v2 preview).

Hubpool addresses is based on v1 hubpool, which probably isnt right but currently theres no v2 deployment.
weth address is an env for the pool tab, but i think we could probably update that to deduce from on chain data.

edit: hub pool has been updated with the config store address

### DIFF
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -13,6 +13,53 @@ import {
 import { usePrevious } from "hooks";
 import { BigNumber } from "ethers";
 
+export function useNativeBalance(
+  tokenSymbol?: string,
+  chainId?: ChainId,
+  account?: string,
+  blockNumber?: number
+) {
+  const { account: connectedAccount } = useConnection();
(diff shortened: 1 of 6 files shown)


src/hooks/useBalance.ts
```diff
@@ -13,6 +13,53 @@ import {
...
+        accountToQuery,
+        blockNumberToQuery,
+        chainIdToQuery,
+        tokenSymbolToQuery
+      )
+    : [
+        "DISABLED_BALANCE_QUERY",
+        {
+          chainIdToQuery,
+          tokenSymbolToQuery,
```
mrice32, line 41: Why do we put token symbol here if it's not always defined and it's not used in the query function? Could we completely drop it from this hook?

mrice32 (approved):
LGTM! One minor comment

Tulun (approved):

PR #62, opened 2022-05-06
feat: Create slide animation for sidebar

Tulun:
See Sidebar for animation.


daywiss (commented):
looks good on destkop, maybe a bit slow, but when on smaller screen its really broken
image

daywiss (approved):
lgtm

PR #66, opened 2022-05-09
fix: calculate if liquidity is sufficient for bridge

daywiss:

This checks the hubpool to make sure theres sufficient liquidity for the bridge amount. Theres some linting changes in here as well.

image


amateima (approved):

src/utils/bridge.ts
```diff
@@ -237,9 +237,18 @@ export default class LpFeeCalculator {
...
+  async isLiquidityInsufficient(
+    tokenAddress: string,
+    amount: utils.BigNumberish
+  ): Promise<boolean> {
+    const [, pooledTokens] = await Promise.all([
+      this.hubPoolInstance.callStatic.sync(tokenAddress),
+      this.hubPoolInstance.callStatic.pooledTokens(tokenAddress),
+    ]);
+    return pooledTokens.liquidReserves.lt(amount);
+  }
```
mrice32, line 249: This is the only non-lint change right?

mrice32 (approved):
LGTM!

src/utils/bridge.ts
daywiss: yes thats right!

PR #67, opened 2022-05-09
fix send to modal height

Tulun:
Story details: https://app.shortcut.com/uma-project/story/5448


daywiss (commented):
looks good, but when i use mobile view it looks longer than it should be? maybe its acceptable, idk 
image

Tulun:
> looks good, but when i use mobile view it looks longer than it should be? maybe its acceptable, idk image

I tried it using a few different mobile views, which one did you use?

amateima (approved):

PR #73, opened 2022-05-11
improve: add and update routes for testnets

daywiss:

This adds goerli, and updates the kovan deployment.  Also fixes linting staged files.

### DIFF
--- a/src/data/routes_5_0xA44A832B994f796452e4FaF191a041F791AD8A0A.json
+++ b/src/data/routes_5_0xA44A832B994f796452e4FaF191a041F791AD8A0A.json
@@ -1,0 +1,25 @@
+{
+  "hubPoolChain": 5,
+  "hubPoolAddress": "0xA44A832B994f796452e4FaF191a041F791AD8A0A",
+  "hubPoolWethAddress": "0x60D4dB9b534EF9260a88b0BED6c486fe13E604Fc",
+  "routes": [
+    {
+      "fromChain": 80001,
+      "toChain": 5,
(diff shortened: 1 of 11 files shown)


src/data/routes_5_0xA44A832B994f796452e4FaF191a041F791AD8A0A.json
```diff
@@ -0,0 +1,34 @@
...
+      "l1TokenAddress": "0x499d11E0b6eAC7c0593d8Fb292DCBbF815Fb29Ae"
+    },
+    {
+      "fromChain": 80001,
+      "toChain": 5,
+      "fromTokenAddress": "0xA6FA4fB5f76172d178d61B04b0ecd319C5d1C0aa",
+      "fromSpokeAddress": "0x45fF03629D024b7763275e732a2d80202c18b31C",
+      "fromTokenSymbol": "WETH",
+      "isNative": false,
+      "l1TokenAddress": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
```
daywiss, line 31: i manually added this, i dont think its right, its actually mainnet weth address, but it should be the token address of the hubpools chains equivalent token.  Normally i can get this from events, but it wasnt available @chrismaree

daywiss: ended up removing this route, since it doesnt seem like it should be added.

Tulun (approved):

PR #81, opened 2022-05-12
fix: fix matic gas estimate

mrice32:


src/utils/bridge.ts
```diff
@@ -286,7 +286,7 @@ class Queries implements relayFeeCalculator.QueryInterface {
     private provider: Provider,
     private chainId: ChainId,
     private priceSymbol = "eth",
-    private defaultGas = "305572"
+    private defaultGas = "120000"
```
mrice32, line 289: Just turned this down to aid in testing and because this is an overestimate on some networks.

daywiss (approved):

PR #88, opened 2022-05-13
feat: migration page

Tulun:
Story details: https://app.shortcut.com/uma-project/story/5498/across-v2-migration-page

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -13,13 +13,13 @@ import {
   WalletWrapper,
 } from "./Header.styles";
 import MenuToggle from "./MenuToggle";
-import { showMigrationPage } from "utils";
+import { migrationPoolV2Warning } from "utils";
 
-const LINKS = !showMigrationPage
+const LINKS = !migrationPoolV2Warning
   ? [
       { href: "/", name: "Bridge" },
       { href: "/pool", name: "Pool" },
-      { href: "/about", name: "About" },
+      { href: "/transactions", name: "Transactions" },
     ]
   : [];
 
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -13,12 +13,15 @@ import {
   WalletWrapper,
 } from "./Header.styles";
 import MenuToggle from "./MenuToggle";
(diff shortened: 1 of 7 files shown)


src/components/PoolSelection/PoolSelection.tsx
```diff
@@ -56,6 +63,22 @@ const PoolSelection: FC<Props> = ({ token, setToken, tokenList, chainId }) => {
   return (
     <AnimatePresence>
       <Wrapper>
+        {showMigrationPage ? (
```
daywiss, line 66: this already exists, there is an env which enables this, we should keep this so we can enable/disable with the env.

image

Imo we should use this copy in the existing migration warning

src/components/Header/Header.tsx
```diff
@@ -13,12 +13,15 @@ import {
...
-const LINKS = [
-  { href: "/", name: "Bridge" },
-  { href: "/pool", name: "Pool" },
-  { href: "/transactions", name: "Transactions" },
-];
+const LINKS = !showMigrationPage
+  ? [
+      { href: "/", name: "Bridge" },
+      { href: "/pool", name: "Pool" },
+      { href: "/about", name: "About" },
```
daywiss, line 22: this should be transactions, not about

Tulun: fixed

mrice32:
Screen Shot 2022-05-16 at 5 30 59 PM
@Tulun something odd I noticed in the v2 UI (maybe unrelated to this PR): when I scrolled down, I was able to scroll past the bottom of the box. This doesn't happen in the v1 UI.

mrice32 (approved):
Looks great! I ran this locally and noticed a minor difference between this and the v1 UI, but not sure if it's related to these changes: https://github.com/across-protocol/frontend-v2/pull/88#issuecomment-1128158741.

Tulun:
> Screen Shot 2022-05-16 at 5 30 59 PM
> 
> @Tulun something odd I noticed in the v2 UI (maybe unrelated to this PR): when I scrolled down, I was able to scroll past the bottom of the box. This doesn't happen in the v1 UI.

Should be unrelated, but we can make a bug ticket for it.

PR #89, opened 2022-05-16
feat(tx-history): add adjustable tx history page size 

amateima:
Screenshot 2022-05-16 at 20 52 54

### DIFF
--- a/src/views/Transactions/useTransactionsView.ts
+++ b/src/views/Transactions/useTransactionsView.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useMemo } from "react";
 import { useConnection } from "state/hooks";
 import { onboard } from "utils";
 import useWindowSize from "hooks/useWindowsSize";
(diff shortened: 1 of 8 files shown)


src/views/Transactions/useTransactionsView.ts
```diff
@@ -76,6 +82,11 @@ export default function useTransactionsView() {
     };
   }, [timer]);
 
+  function onPageSizeChange(value: number) {
+    setPageSize(value);
+    localStorage.setItem(TX_HISTORY_PAGE_SIZE_KEY, value.toString());
+  }
+
```
amateima, line 89: Cache the user preference in terms of page size

Tulun: I might split this logic into it's own utility function. WYDT?

Tulun (commented):
One small comment. But this is basically just a port from the OO UI?

amateima:
> One small comment. But this is basically just a port from the OO UI?

Actually I didn't take a look there to see how it was implemented

src/views/Transactions/useTransactionsView.ts
amateima: Sure. [Changes here](https://github.com/across-protocol/frontend-v2/pull/89/commits/73de817757abeded16343066dff43ad4d53052bf)

Tulun (approved):

Tulun (approved):

daywiss (commented):
code looks good, works well, but you also have to apply this to the mobile table

daywiss (approved):

PR #95, opened 2022-05-18
feat: add new relayer fee logic

mrice32:
This PR adds two features that come out of updates in the SDK:

1. It updates the way we get Query objects for computing gas costs on each chain (we now use chain-specific objects provided by the SDK).
2. It updates how we compute Network Gas Fees vs Bridge Fees. We group relayer capital costs and LP fees into bridge fees and put _only_ the relayer gas fees into gas fees.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@across-protocol/contracts-v2": "^0.0.34",
-    "@across-protocol/sdk-v2": "^0.1.2",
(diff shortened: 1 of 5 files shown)


mrice32:
This depends on https://github.com/across-protocol/sdk-v2/pull/35 being merged first.

daywiss (approved):
looks great. tested this across all chains. i think we may have to optimize coingecko usage as it ets rate limited switching between chains/currencies a lot

PR #96, opened 2022-05-18
fix: prevent amount and asset from resetting when switching chains

daywiss:

Previously form inputs would reset when changing any of the selections.  Now this will not reset the amount, and attempt to keep the token selection and chain selection the same. This does not work in all cases, where its an invalid selection it will try to prioritize resetting the toChain and then the token.

Also the ordering of tokens was fixed, its now set by the constants file token array.

### DIFF
--- a/src/components/CoinSelection/useCoinSelection.tsx
+++ b/src/components/CoinSelection/useCoinSelection.tsx
@@ -1,5 +1,5 @@
 import { useSelect } from "downshift";
-import { ethers } from "ethers";
+import { ethers, BigNumber } from "ethers";
 import { formatUnits, parseUnits } from "ethers/lib/utils";
 import React, { useEffect, useCallback } from "react";
 import { useConnection } from "state/hooks";
@@ -23,7 +23,6 @@ export default function useCoinSelection() {
     fromChain,
     error: formError,
     setError: setFormError,
-    status: formStatus,
     availableTokens,
   } = useSendForm();
 
@@ -38,7 +37,6 @@ export default function useCoinSelection() {
     selectedItem: availableTokens.find((t) => t.symbol === tokenSymbol),
     onSelectedItemChange: ({ selectedItem }) => {
       if (selectedItem) {
-        setInputAmount("");
         setTokenSymbol(selectedItem.symbol);
         // Matomo track token selection
         trackEvent({
@@ -54,6 +52,7 @@ export default function useCoinSelection() {
   );
   const balance = balances[selectedIndex];
   const { fees } = useBridgeFees(amount, toChain, selectedItem?.symbol);
+
   const [inputAmount, setInputAmount] = React.useState<string>(
     selectedItem && amount.gt("0")
       ? formatUnits(amount, selectedItem.decimals)
@@ -61,32 +60,31 @@ export default function useCoinSelection() {
   );
 
   useEffect(() => {
-    if (formStatus === "idle") {
-      setInputAmount("");
+    if (inputAmount === "" || inputAmount === "0") {
+      setAmount(ethers.constants.Zero);
+      return;
     }
-  }, [formStatus]);
+    try {
+      const amount = parseUnits(inputAmount, selectedItem?.decimals);
+      setAmount(amount);
+    } catch (e) {
+      // if we have a parsing error, we have to set this to 0 otherwise we might get fee calculation issues
+      setAmount(BigNumber.from(0));
+      // this can throw an error if parseUnits fails for an input amount that has more decimals than token decimals
+      setFormError(new ParsingError());
+    }
+  }, [inputAmount, selectedItem, setAmount, setFormError, tokenSymbol]);
 
   const handleInputChange = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
       const value = event.target.value;
       setInputAmount(value);
-      if (value === "") {
-        setAmount(ethers.constants.Zero);
-        return;
-      }
-      try {
-        const amount = parseUnits(value, selectedItem?.decimals);
-        setAmount(amount);
-      } catch (e) {
-        setFormError(new ParsingError());
-      }
     },
-    [selectedItem, setAmount, setFormError]
+    [setInputAmount]
   );
 
   const handleMaxClick = useCallback(() => {
     if (!balance || !selectedItem) {
-      setAmount(ethers.constants.Zero);
       setInputAmount(formatUnits(ethers.constants.Zero));
       return;
     }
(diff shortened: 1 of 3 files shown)


src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -54,39 +52,36 @@ export default function useCoinSelection() {
...
+      return;
+    }
+    try {
+      const amount = parseUnits(inputAmount, selectedItem?.decimals);
+      setAmount(amount);
+    } catch (e) {
+      setFormError(new ParsingError());
     }
-  }, [formStatus]);
+  }, [inputAmount, selectedItem, setAmount, setFormError]);
```
daywiss, line 73: the way the form amount is set was changed in order to maintain the value and validation across form changes.  The local input box is set independently, and changes to that value get passed to the global form state. The validation happens on the global amount in this effect.

src/hooks/useSendForm.tsx
```diff
@@ -246,15 +246,16 @@ function toChainReducer(state: FormState, chainId: ChainId): FormState {
...
-    const [firstRoute] = availableRoutes;
-    fromChain = firstRoute.fromChain;
-    tokenSymbol = firstRoute.fromTokenSymbol;
   }
+
+  const [firstRoute] = availableRoutes;
+  fromChain = firstRoute.fromChain;
+  tokenSymbol = firstRoute.fromTokenSymbol;
+  const selectedRoute = firstRoute;
+
```
daywiss, line 255: simplified some logic

src/utils/config.ts
```diff
@@ -29,12 +33,34 @@ export class ConfigClient {
...
+      assert(
+        constants.isSupportedChainId(chainId),
+        "Unsupported chainId: " + chainId
+      );
+      this.chainOrder[chainId] = Number(index);
+    });
+    // prioritize routes based on token symbol and tochain. This just gives us better route prioritization when filtering a fromChain
+    this.routes = sortBy(this.config.routes, (route) => {
+      return (
+        this.tokenOrder[route.fromTokenSymbol] + this.chainOrder[route.toChain]
```
daywiss, line 55: sorting the routes helps us select an ideal route to default to if we cant find one based on users selection. in this case we want to put routes with priority tokens and priority chains first

chrismaree (commented):
I cant comment on the code but testing the link is SOOO much better when I use it!

daywiss:
> I think technically fine, but I will say this code is getting harder and harder to reason about. Would really flag this probably needs a refactor as it's probably quite fragile and there's likely some nasty edge cases we aren't catching.

yea it is really complicated, i've tried to make this robust as possible by only allowing valid routes to be shown in the form, ie changing user selections when they are invalid, but its quite possible theres some wierd edge case we are missing. In past iterations of this code though it woudl have been much more fragile and probably not possible to add this feature.

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -54,39 +52,39 @@ export default function useCoinSelection() {
...
 
   useEffect(() => {
-    if (formStatus === "idle") {
-      setInputAmount("");
+    if (inputAmount === "" || inputAmount === "0") {
+      setAmount(ethers.constants.Zero);
+      return;
+    }
+    try {
+      const amount = parseUnits(inputAmount, selectedItem?.decimals);
```
mrice32, line 68: Does this mean we default to 18 decimals if there is no selected item? When would this happen?

src/hooks/useSendForm.tsx
```diff
@@ -180,24 +178,27 @@ function fromChainReducer(state: FormState, chainId: ChainId): FormState {
...
   });
-
-  // we have from, to, symbols that produce no routes, so reset the form to the first valid route
-  if (!availableRoutes.length) {
-    availableRoutes = config.filterRoutes({ fromChain });
-    const [firstRoute] = availableRoutes;
-    toChain = firstRoute.toChain;
-    tokenSymbol = firstRoute.fromTokenSymbol;
+  // if no valid routes found with user selection, de-prioritize the toChain to maintain token symbol.
+  if (!selectedRoutes.length) {
```
mrice32, line 193: nit: prefer a more explicit check
```suggestion
  if (selectedRoutes.length === 0) {
```

src/hooks/useSendForm.tsx
```diff
@@ -180,24 +178,27 @@ function fromChainReducer(state: FormState, chainId: ChainId): FormState {
...
-  let toChain = undefined;
+  let toChain = state.toChain;
   let tokenSymbol = state.tokenSymbol;
 
-  let availableRoutes = config.filterRoutes({
-    toChain,
+  // availble routes are all possible routes with this from chain selected
+  const availableRoutes = config.filterRoutes({ fromChain });
+  // selected routes are all possible chains given the users selection from chain and token symbol.
+  let selectedRoutes = config.filterRoutes({
```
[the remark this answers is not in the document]
mrice32: This is awesome!

src/utils/config.ts
mrice32: +1

mrice32 (approved):
LGTM! This is very nice

src/components/CoinSelection/useCoinSelection.tsx
amateima: it looks like we need 2 state variables: one to keep the input value and another one which keeps the calculated amount based on the input value and the selected token. This way defaulting to 18 decimals won't happen

daywiss: yea i can check for existence first and not do anything if not found

src/components/CoinSelection/useCoinSelection.tsx
```diff
@@ -54,39 +52,39 @@ export default function useCoinSelection() {
...
       ? formatUnits(amount, selectedItem.decimals)
       : ""
   );
 
   useEffect(() => {
-    if (formStatus === "idle") {
-      setInputAmount("");
+    if (inputAmount === "" || inputAmount === "0") {
+      setAmount(ethers.constants.Zero);
+      return;
```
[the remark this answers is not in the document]
daywiss: i think its ok to return early with undefined, but i can try to nest branches if this is not acceptable

src/hooks/useSendForm.tsx
daywiss: i updated this

src/components/CoinSelection/useCoinSelection.tsx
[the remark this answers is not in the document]
daywiss: i put this in a branch so to not return early

PR #97, opened 2022-05-18
fix: stop showing bad cache balance values

daywiss:

The issue was that a bad balance would flash for a second when switching from chain. 

I dont know exactly what this code did but with it in, it would flash an incorrect balance for a second. With this code the balance goes blank once selecting a new from chain, resolving the issue.

### DIFF
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -10,7 +10,6 @@ import {
   ChainId,
   getConfig,
 } from "utils";
-import { usePrevious } from "hooks";
 import { BigNumber } from "ethers";
 
 export function useNativeBalance(
@@ -162,14 +161,6 @@ export function useBalancesBySymbols(
         "DISABLED_BALANCES_QUERY",
         { chainIdToQuery, tokenSymbols, accountToQuery, blockNumberToQuery },
       ];
-  const prevAccount = usePrevious(accountToQuery);
-  const prevChain = usePrevious(chainIdToQuery);
-  const prevTokens = usePrevious(tokenSymbols);
-  // Keep the previous data only when blockNumberToQuery changes.
-  const keepPreviousData =
-    prevAccount === accountToQuery &&
-    prevChain === chainIdToQuery &&
-    JSON.stringify(prevTokens) === JSON.stringify(tokenSymbols);
   const { data: balances, ...delegated } = useQuery(
     queryKey,
     async () => {
@@ -184,9 +175,6 @@ export function useBalancesBySymbols(
       enabled: enabledQuery,
       // We already re-fetch when the block change, so we don't need to re-fetch.
       staleTime: Infinity,
-      // Old balances can be garbage collected immediately
-      cacheTime: 0,
-      keepPreviousData,
     }
   );
   return {
--- a/src/hooks/useBalance.ts
+++ b/src/hooks/useBalance.ts
@@ -1,14 +1,14 @@
-import { useQuery } from "react-query";
+import { useQuery, useQueries } from "react-query";
 import { useConnection } from "state/hooks";
 import { useBlock } from "./useBlock";
+import { usePrevious } from "hooks";
 import {
   getBalance,
   getNativeBalance,
-  getBalancesBySymbols,
   balanceQueryKey,
-  balancesQueryKey,
   ChainId,
   getConfig,
+  ConfigClient,
 } from "utils";
 import { BigNumber } from "ethers";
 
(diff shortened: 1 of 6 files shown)


Tulun (approved):

src/hooks/useBalance.ts
```diff
@@ -162,14 +161,6 @@ export function useBalancesBySymbols(
...
         { chainIdToQuery, tokenSymbols, accountToQuery, blockNumberToQuery },
       ];
-  const prevAccount = usePrevious(accountToQuery);
-  const prevChain = usePrevious(chainIdToQuery);
-  const prevTokens = usePrevious(tokenSymbols);
-  // Keep the previous data only when blockNumberToQuery changes.
-  const keepPreviousData =
-    prevAccount === accountToQuery &&
-    prevChain === chainIdToQuery &&
-    JSON.stringify(prevTokens) === JSON.stringify(tokenSymbols);
```
daywiss, line 172: this didnt seem to do anything other than show a wrong value when the query changed

src/utils/query-keys.ts
```diff
@@ -41,7 +41,7 @@ export function balancesQueryKey(
   blockNumber: number,
   chainId?: ChainId
 ) {
-  return ["balances", chainId, tokenSymbols, account, blockNumber];
+  return ["balances", chainId, tokenSymbols.join(","), account, blockNumber];
```
mrice32, line 44: What is this change?

src/hooks/useBalance.ts
```diff
@@ -184,9 +175,6 @@ export function useBalancesBySymbols(
       enabled: enabledQuery,
       // We already re-fetch when the block change, so we don't need to re-fetch.
       staleTime: Infinity,
-      // Old balances can be garbage collected immediately
-      cacheTime: 0,
-      keepPreviousData,
```
mrice32, line 189: I think all this logic that you removed was meant to solve the issue where the value clears when the block number changes, despite having a _somewhat_ valid past value to show.

The logic wanted to do that while also forcing an immediate invalidation if anything _other than_ the block number changed. My guess is that with this change, the block number updating (which is probably happening on a poll) will cause the balance to momentarily clear.

Does that match your intuition and do you think that's a reasonable UX tradeoff?

Note: see the origin PR: https://github.com/across-protocol/frontend-v2/pull/35

mrice32 (approved):
LGTM! Just a few minor questions

src/utils/query-keys.ts
daywiss: it just turns the symbols array into a string rather than an array, it probably doesnt change anything

src/hooks/useBalance.ts
daywiss: yea thats probably waht it was. the problem is that when the new balance was loading it showed a compeltely wrong value, and i wasnt able to figure out where it was coming from. might be worth spending more time to figure out why the previous data was wrong

daywiss: spent some more time on this. i think ultimately whats happening is the pattern we are using to "disable_balance_query" as a key in the cache is caching a random balance value when you switch between chains ( because block number is not available immediately, which is required for the cache key).  So whats happening is we swtich between chains, which uses the "disabled_balance_query", then the "keepPreviousData" flag displays the value in that cache, which seems to always be wrong ( it tends to hold the amount of weth), and it will return that value for all tokens.

daywiss: theres a few convoluting things.   one is that we are caching with a block number, so when that doesnt exist we have no ability to look back on the previously cached value, defaulting to a single value cached across all chains/tokens ( disabled key).  The other is this pattern of disabling the cache to a canonical location which does not play nice with the options on react query.

I think the correct solution here is to use cache invalidation on a timer for balance, rather than tyring to sync it to a specific block number, remove the block number from the cache key, and always query balances from "latest".

mrice32 (approved):
LGTM!

src/utils/query-keys.ts
```diff
@@ -20,29 +20,13 @@ export function latestBlockQueryKey(chainId: ChainId) {
  */
 export function balanceQueryKey(
   account: string,
-  blockNumber: number,
+  blockNumber?: number,
```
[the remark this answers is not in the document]
daywiss: yea exactly, it will just go to same cache slot, which now will just be 0

src/hooks/useBalance.ts
```diff
@@ -91,33 +132,21 @@ export function useBalanceBySymbol(
       )
     : [
         "DISABLED_BALANCE_QUERY",
-        {
-          chainIdToQuery,
-          tokenSymbolToQuery,
-          accountToQuery,
-          blockNumberToQuery,
-        },
+        chainIdToQuery,
```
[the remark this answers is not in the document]
daywiss: i dont know for sure, but that is my guess, react query probably just calls to string on the object to serialize it

PR #117, opened 2022-05-24
fix: fix call to isSuppotedChain on config

daywiss:


Tulun (approved):

Tulun:
@mrice32, @nicholaspai   -- good reason we really need to up our regression testing game.

Simple an unbound function call on a class.

PR #120, opened 2022-05-25
feat: add all txs history section

amateima:
Signed-off-by: amateima <amatei@umaproject.org>


daywiss (commented):
this is a really cool feature.  the first time i ran it there were a bunch of errors in console that lead to showing pending transactions in the history page, but i havent been able to reproduce that after refreshing.

the other issue i noticed is that the history page state is shared between my transactions and all transactions, which i think the users current page should be separate between tables.  you can select the last page in the all transactions table, then go to your history and the last page will also be selected.  other than that it seems great

PR #121, opened 2022-05-25
feat: add amount and asset to urlParams for send query.

Tulun:
Story details: https://app.shortcut.com/uma-project/story/5611

### DIFF
--- a/src/components/CoinSelection/useCoinSelection.tsx
+++ b/src/components/CoinSelection/useCoinSelection.tsx
@@ -10,10 +10,14 @@ import {
   InsufficientBalanceError,
   trackEvent,
   FEE_ESTIMATION,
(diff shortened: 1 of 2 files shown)


daywiss:
it seems like i should be to partially have query parameters and have the form fill accordingly
ie just set `to=1` or `to=1&amount=1` but this seems to only set the form if you have some minimum parameters

Tulun:
> it seems like i should be to partially have query parameters and have the form fill accordingly
> ie just set `to=1` or `to=1&amount=1` but this seems to only set the form if you have some minimum parameters

this is to spec -- should we adjust this?

daywiss (approved):
looks good works for me

amateima (approved):

PR #122, opened 2022-05-26
fix: refetch all pool balances on token changes

Tulun:
Update all pools on change instead of the singular one.

### DIFF
--- a/src/views/Pool.tsx
+++ b/src/views/Pool.tsx
@@ -96,14 +96,9 @@ const Pool: FC = () => {
       })
       .finally(() => {
         setLoadingPoolState(false);
-        // lazily load other pools
-        const otherPools = tokenList.filter((x) => x.address !== address);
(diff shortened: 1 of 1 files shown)


src/views/Pool.tsx
```diff
@@ -96,7 +96,13 @@ const Pool: FC = () => {
       })
       .finally(() => {
         setLoadingPoolState(false);
+        // lazily load other pools
+        const otherPools = tokenList.filter((x) => x.address !== address);
+        otherPools.forEach((x) => {
+          poolClient.updatePool(x.address);
```
daywiss, line 102: the issue is that the users token balances dont update (not lp balances), which is currently not handled in the pool client.  this is called through the chainApi file i believe.  the card has the steps to reproduce issue

daywiss:
its working!

daywiss (approved):

PR #123, opened 2022-05-30
feat: add max input button to remove form

Tulun:
Remove liquidity button added:
Screen Shot 2022-05-30 at 4 28 07 PM


Tulun:
@daywiss: I think the decimal rounding Korpi has suggested is affecting some of the UI here -- we may wanna adjust it back. WDYT?

daywiss:
> @daywiss: I think the decimal rounding Korpi has suggested is affecting some of the UI here -- we may wanna adjust it back. WDYT?


I'm seeing that hitting max does not allow you to withdraw, is that what you are referring to?

src/components/PoolForm/RemoveLiquidityForm.tsx
```diff
@@ -144,34 +150,106 @@ const RemoveLiqudityForm: FC<Props> = ({
...
+          {
+            totalDeposited: position,
+            feesEarned: max(feesEarned, 0),
+            positionValue: totalPosition,
+          },
+          removeAmountSlider / 100
+        )
+      : null;
+
+  const calculateRemoveAmount = (
```
daywiss, line 165: this is defined more than once https://github.com/across-protocol/frontend-v2/pull/123/files#diff-1e1d77462c947d5a2ed268da655215c899db2cfb25a6cbaac56dc3e2ea776a37R62

amateima (commented):
> why do we need 2 errors now? why not just disable the button and show the error under the input box like we do with other forms?

➕

src/components/PoolForm/RemoveLiquidityForm.tsx
Tulun: Removed

src/utils/token.ts
```diff
@@ -58,3 +58,17 @@ export async function getAllowance(
...
+
+export const calculateRemoveAmount = (
+  percent: number,
+  position: ethers.BigNumber,
+  decimals: number
+) => {
+  if (position.toString() === "0") return "0";
+  const scaler = ethers.BigNumber.from("10").pow(decimals);
+
+  const removeAmountToWei = toWeiSafe((percent / 100).toString(), decimals);
```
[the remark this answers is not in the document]
Tulun: So this is the preferred operation?

`
export const calculateRemoveAmount = (
  percent: number,
  position: ethers.BigNumber,
  decimals: number
) => {
  if (position.toString() === "0") return "0";
  const scaler = ethers.BigNumber.from("10").pow(decimals);

  const removeAmountToWei = toWeiSafe(percent.toString(), decimals);

  const weiAmount = position.mul(removeAmountToWei).div(scaler.mul(100));
  return formatUnits(weiAmount, decimals);
};
`

daywiss:
looking a lot better, but im able to send a transaction when my removal amount is set higher than my balance. it seems like its automatically limiting me but still shows an error
image

it seems like it shoudl either show me an error or tell me that my withdraw has been limited to 100% or something

src/components/PoolForm/RemoveLiquidityForm.tsx
```diff
@@ -248,4 +321,36 @@ const RemoveLiqudityForm: FC<Props> = ({
...
+      const valueToWei = toWeiSafe(value, decimals);
+      if (valueToWei.add(addLiquidityGas).gt(balance)) {
+        return setFormError("Transaction may fail due to insufficient gas.");
+      }
+    }
+  } catch (e) {
+    return setFormError("Invalid number.");
+  }
+  // clear form if no errors were presented. All errors should return early.
+  // setFormError("");
```
daywiss, line 353: we can prob remove?

Tulun:
> looking a lot better, but im able to send a transaction when my removal amount is set higher than my balance. it seems like its automatically limiting me but still shows an error image
> 
> it seems like it shoudl either show me an error or tell me that my withdraw has been limited to 100% or something

Fixed conditional so it looks disabled under an error.

amateima (commented):
Screenshot 2022-06-02 at 12 46 38

Nits:
- I think the error spacing needs to be adjusted a little bit
- Should we add a slight hovering effect to the slider thumb? (opacity change + `cursor: pointer`)

src/components/PoolForm/useSetLiquidityFormErrors.ts
```diff
@@ -0,0 +1,38 @@
...
+  decimals: number,
+  symbol: string,
+  setFormError: React.Dispatch<React.SetStateAction<string>>,
+  addLiquidityGas: ethers.BigNumber = ethers.BigNumber.from("0")
+) {
+  // Validate input on change
+  useEffect(() => {
+    try {
+      // liquidity button should be disabled if value is 0, so we dont actually need an error.
+      if (Number(value) === 0) return setFormError("");
```
daywiss, line 17: the logic in[ validate form](https://github.com/across-protocol/frontend-v2/pull/123/files#diff-204ee6271e34fe1a7645d34841041af01b490cdbba8e277e3ab8e7d9ad679d5dR324) looks identical to this, is there a way to reduce the duplicated code

daywiss:
should probably do a check to set remove amount to 0 if theres a form errror:
image

src/components/PoolForm/PoolForm.tsx
```diff
@@ -115,48 +116,39 @@ const PoolForm: FC<Props> = ({
...
+  useEffect(() => {
+    if (position.toString() && Number(removeAmount)) {
+      const wei = Number(toWeiSafe(removeAmount, decimals).toString());
+      const pos = Number(position.toString());
+      const percent = (wei / pos) * 100;
+      if (percent >= 100) {
+        setRemoveAmountSlider(100);
+        // Don't round up to 100% unless they max out.
+      } else if (percent >= 99.5 && percent < 100) {
+        setRemoveAmountSlider(99);
```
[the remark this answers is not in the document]
Tulun: Changed.

src/utils/token.ts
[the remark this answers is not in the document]
Tulun: Believe fixed :)

src/components/PoolForm/useSetLiquidityFormErrors.ts
Tulun: I'll give it a stab, but yeah, I tried to integrate this with both types of inputs but the recursiveness of the two values made this not work on submit validation as well, so I split it out.

amateima (approved):

PR #124, opened 2022-06-01
fix: use routes constant and burner address to simulate OP and ARB txns

mrice32:


daywiss (approved):

src/utils/constants.ts
```diff
@@ -543,42 +543,78 @@ export const migrationPoolV2Warning =
...
+const getQueriesTable = () => {
+  const optimismUsdcRoute = getRoute(1, 10, "USDC");
+  const arbitrumUsdcRoute = getRoute(1, 42161, "USDC");
+
+  // Note: this address is used as the from address for simulated relay transactions on Optimism and Arbitrum since
+  // gas estimates require a live estimate and not a pre-configured gas amount. This address has been pre-loaded with
+  // a USDC approval for the _current_ spoke pools on Optimism (0xa420b2d1c0841415A695b81E5B867BCD07Dff8C9) and Arbitrum
+  // (0xB88690461dDbaB6f04Dfad7df66B7725942FEb9C). It also has a small amount of USDC ($0.10) used for estimations.
+  const dummyFromAddress = "0x893d0d70ad97717052e3aa8903d9615804167759";
+
```
Tulun, line 569: This seems like a very fragile way of implementing this -- we could add it like an env variable  and force existence but it wouldn't necessarily inform us if this address is *still* valid.

Maybe as a sanity check we should attempt to test the balance of this address and throw some kind of Error?

amateima (approved):
I have no comments other than what Jason said

Tulun (approved):

PR #126, opened 2022-06-02
fix: Do a check on overrided address if it's a contract code and show a warning.

Tulun:
Story details: https://app.shortcut.com/uma-project/story/5675/warn-user-that-if-they-re-setting-deposit-recipient-to-a-smart-contract-that-they-will-receive-weth-therefore-tell-them-not

Do getCode(address) call to check if the address inputted is a Contract. If so, show a warning.

To test: Set bridge currency to ETH or WETH, adjust address manually to a known contract address on the toChain, then try to save the address.
Screen Shot 2022-06-03 at 1 29 39 PM

### DIFF
--- a/src/components/AddressSelection/AddressSelection.tsx
+++ b/src/components/AddressSelection/AddressSelection.tsx
@@ -24,6 +24,7 @@ import {
   RoundBox,
   ToggleChainName,
   Address,
+  WarningBox,
 } from "./AddressSelection.styles";
 import { AnimatePresence } from "framer-motion";
 import useAddressSelection from "./useAddressSelection";
@@ -47,6 +48,7 @@ const AddressSelection: React.FC = () => {
     availableToChains,
     selectedToChainInfo,
     fromChain,
+    showContractAddressWarning,
   } = useAddressSelection();
 
   return (
@@ -130,6 +132,13 @@ const AddressSelection: React.FC = () => {
                 Save Changes
               </SecondaryButton>
             </ButtonGroup>
+            {showContractAddressWarning && (
+              <WarningBox>
+                {`
+              Warning --- Address you inputted is a contract address. When sending ETH from L2 -> L1 WETH will arrive. Ensure the contract address will properly handle WETH or this may result in loss of funds.
+            `}
+              </WarningBox>
+            )}
           </Dialog>
         )}
       </LastSection>
--- a/src/components/AddressSelection/AddressSelection.tsx
+++ b/src/components/AddressSelection/AddressSelection.tsx
@@ -25,6 +25,7 @@ import {
   ToggleChainName,
   Address,
   WarningBox,
+  WarningButton,
 } from "./AddressSelection.styles";
 import { AnimatePresence } from "framer-motion";
 import useAddressSelection from "./useAddressSelection";
@@ -49,6 +50,7 @@ const AddressSelection: React.FC = () => {
     selectedToChainInfo,
     fromChain,
     showContractAddressWarning,
+    overrideAddress,
   } = useAddressSelection();
 
   return (
@@ -137,6 +139,9 @@ const AddressSelection: React.FC = () => {
                 {`
               Warning --- Address you inputted is a contract address. When sending ETH from L2 -> L1 WETH will arrive. Ensure the contract address will properly handle WETH or this may result in loss of funds.
             `}
+                <WarningButton onClick={overrideAddress}>
+                  Save Contract Address
+                </WarningButton>
               </WarningBox>
             )}
           </Dialog>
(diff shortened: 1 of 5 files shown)


Tulun:
@daywiss @nicholaspai  would love feedback. Added the second button because I want the user to option to forcefully override and it was the most straightforward way of inputting a *second*  kind of action.

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -56,16 +68,42 @@ export default function useAddressSelection() {
...
   const handleSubmit = () => {
+    setShowContractAddressWarning(false);
+
     if (isValid) {
-      if (address) {
-        setToAddress(address);
+      if (address && toChain) {
+        // Check to see if the toAddress they are inputting is a Contract on Mainnet
+        // If so, warn user because we send WETH and this could cause loss of funds.
+        if (tokenSymbol === "ETH" || tokenSymbol === "WETH") {
```
daywiss, line 77: im actually afraid they could change the recipient address and then switch token to eth or weth. may have to rethink how this is shown @chrismaree @Tulun

daywiss (approved):
simple enough change but may want to give feedback to design in case the user switches currency to eth/weth after they change to address

amateima (approved):

src/components/AddressSelection/useAddressSelection.ts
amateima: good catch

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -130,6 +134,16 @@ const AddressSelection: React.FC = () => {
                 Save Changes
               </SecondaryButton>
             </ButtonGroup>
+            {showContractAddressWarning && (
+              <WarningBox>
+                {`
+              Warning --- Address you inputted is a contract address. When sending ETH from L2 -> L1 WETH will arrive. Ensure the contract address will properly handle WETH or this may result in loss of funds.
```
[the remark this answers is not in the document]
Tulun: Message updated.

src/components/AddressSelection/AddressSelection.tsx
```diff
@@ -24,10 +24,12 @@ import {
   RoundBox,
   ToggleChainName,
   Address,
+  WarningBox,
 } from "./AddressSelection.styles";
 import { AnimatePresence } from "framer-motion";
 import useAddressSelection from "./useAddressSelection";
 
+const warningMessage = `Warning --- Address you inputted is a contract address. When sending ETH across chains WETH will arrive. Ensure the contract address will properly handle WETH or send to a different EOA and convert your WETH to ETH and send to the contract address, or this may result in loss of funds.`;
```
nicholaspai, line 32:
```suggestion
const warningMessage = `Warning --- You've inputted a contract address. When bridging ETH to a contract address, ETH will be wrapped into WETH. Ensure the contract address can receive WETH. If you are not sure whether it can, then you should send to a non-contract address to be safe and receive ETH to prevent any chance of a loss of funds.`;
```

src/utils/constants.ts
```diff
@@ -429,6 +429,8 @@ export const AddressZero = ethers.constants.AddressZero;
 export const ArbitrumProviderUrl =
   process.env.REACT_APP_CHAIN_42161_PROVIDER_URL ||
   `https://arbitrum-mainnet.infura.io/v3/${infuraId}`;
+
+console.log(ArbitrumProviderUrl);
```
nicholaspai, line 433: leftover log

nicholaspai (approved):
LGTM, left a couple comments

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -49,23 +62,52 @@ export default function useAddressSelection() {
...
+
     if (isValid) {
-      if (address) {
-        setToAddress(address);
+      if (address && toChain) {
+        // Check to see if the toAddress they are inputting is a Contract on Mainnet
+        // If so, warn user because we send WETH and this could cause loss of funds.
+        // Note: Removed check for WETH and ETH because they can change tokens outside of this modal.
+        getCode(address, toChain)
+          .then((addr) => {
```
daywiss, line 86: this technically is not returning an address

src/components/AddressSelection/useAddressSelection.ts
```diff
@@ -49,23 +62,52 @@ export default function useAddressSelection() {
...
+            if (addr !== noContractCode) {
+              setShowContractAddressWarning(true);
```
daywiss, line 88: where is the logic to set the toAddress if its a contract address?

PR #138, opened 2022-06-16
feat: bump to sdk-v2 and take advantage of new relayer fee curve

nicholaspai:
## Needs QA

**Overview**

Input amounts to send (on any network, network doesn't matter) for different tokens and see if fees charged make sense. Previously the dapp is charging 0.03% on all relays. The new curve defined for tokens in `src/utils/constants` will charge an increasing fee for the tokens with entries. If there is no entry for a token, then the flat 0.03% should be charged.

**Acceptance Criteria**

TODO

### How To Report QA

**Reporting QA Issues**

```
**QA Issue**

<!--
-->

**Repro Steps**

<!--

Example:
2. go to pool tab before going to transaction tab.
3. see that my pending transaction is missing.
-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->


```

**Reporting QA Regression**

**Reporting QA Passed**

```
**QA Pass**

<!--

Example:

-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@across-protocol/contracts-v2": "^0.0.34",
-    "@across-protocol/sdk-v2": "^0.1.7",
+    "@across-protocol/sdk-v2": "^0.1.13",
     "@datapunt/matomo-tracker-js": "^0.5.1",
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -17,6 +17,7 @@ import {
   getConfigStoreAddress,
   queriesTable,
   FLAT_RELAY_CAPITAL_FEE,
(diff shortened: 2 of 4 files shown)


nicholaspai:
@daywiss @Tulun any idea why this is failing to build?

api/constants.js
```diff
@@ -0,0 +1,52 @@
+import { ethers } from "ethers";
+
+export const relayerFeeCapitalCostConfig = {
```
daywiss, line 3: is this duplicate of whats in the utils file? if so could we try to dedupe

daywiss (commented):
taking a look at why the build fails

api/constants.js
nicholaspai: It is but I thought we can't import from outside `src`into `api`,let me try again

Tulun (approved):
See s fine. Build is also no longer failing?

PR #145, opened 2022-06-27
feat: add variable bridge time estimation to UI

mrice32:
**Overview**

This change allows us to provide users with a more accurate time estimate for their transfers. Instead of always seeing 1-3 minutes, users should see smaller numbers for smaller transfers and as their transfers get very large, the time estimate should tick up to something intermediate. The amount is different depending on the destination chain.

Instant transfers are estimated to take 1-4 minutes (not 1-3, since a typical bot run takes 2 minutes and a user could deposit on the back end of a run).

Transfers that could be funded by mainnet, but would require an L1->L2 transfer have the following estimates: 
- Optimism: 12-25 minutes.
- Polygon: 20-35 minutes.
- Boba: 12-25 (presuming that it's the same as Optimism).
- Arbitrum: 5-15 minutes.
- Default (not one of the above): 10-30 minutes. We should never hit this.

Anything larger than that will return an estimate of 2-4 hours.

### DIFF
--- a/src/hooks/useBridgeLimits.ts
+++ b/src/hooks/useBridgeLimits.ts
@@ -35,7 +35,6 @@ export function useBridgeLimits(
           `/api/limits?token=${token}&originChainId=${fromChainId}&destinationChainId=${toChainId}`
         )
       ).data;
-      console.log(response);
       return {
         minDeposit: BigNumber.from(response.minDeposit),
         maxDeposit: BigNumber.from(response.maxDeposit),
--- a/src/hooks/useBridgeLimits.ts
+++ b/src/hooks/useBridgeLimits.ts
@@ -46,6 +46,7 @@ export function useBridgeLimits(
       enabled: enabledQuery,
       // 5 mins.
       staleTime: 300000,
+      retry: 2,
     }
   );
   return {
--- a/src/hooks/useBridgeLimits.ts
+++ b/src/hooks/useBridgeLimits.ts
@@ -1,0 +1,56 @@
+import { useQuery } from "react-query";
+import { bridgeLimitsQueryKey, ChainId } from "utils";
+import axios from "axios";
+import { BigNumber } from "ethers";
+
+export interface BridgeLimits {
+  minDeposit: BigNumber;
+  maxDeposit: BigNumber;
+  maxDepositInstant: BigNumber;
+  maxDepositShortDelay: BigNumber;
+}
+
+/**
+ * This hook calculates the limit .
+ * @param token The token to get limits for.
+ * @param fromChainId The chain id from which the deposit transaction is sent.
+ * @param toChainId The chain id where the relay will be sent.
+ * @returns The limits datastructure returned from the serverless api.
+ */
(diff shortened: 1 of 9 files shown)


src/hooks/useBridgeLimits.ts
```diff
@@ -0,0 +1,55 @@
...
+  const queryKey = enabledQuery
+    ? bridgeLimitsQueryKey(token, fromChainId, toChainId)
+    : "DISABLED_BRIDGE_LIMITS_QUERY";
+  const { data: limits, ...delegated } = useQuery(
+    queryKey,
+    async (): Promise<BridgeLimits> => {
+      const response = (
+        await axios.get(
+          `/api/limits?token=${token}&originChainId=${fromChainId}&destinationChainId=${toChainId}`
+        )
```
daywiss, line 36: this doesnt currently seem to work on testnets, i'm not actually sure what is returned with RQ when this fails.  perhaps we just need a fallback if the limits don't exist, because it just ends up not showing anything estimation on anything but mainnet

daywiss (approved):
really nice, looks good

amateima (approved):
🔥

src/utils/bridge.ts
```diff
@@ -148,8 +149,30 @@ export async function getBridgeFees({
...
+  } else if (amount.lte(limits.maxDepositShortDelay)) {
+    // This is just a rough estimate of how long 2 bot runs (1-4 minutes allocated for each) + an arbitrum transfer of 3-10 minutes would take.
+    if (toChain === 42161) return "~5-15 minutes";
+
+    // Optimism transfers take about 10-20 minutes anecdotally. Boba is presumed to be similar.
+    if (toChain === 10 || toChain === 288) return "~12-25 minutes";
+
+    // Polygon transfers take 20-30 minutes anecdotally.
+    if (toChain === 137) return "~20-35 minutes";
+
```
Tulun, line 169: I would use ChainID type here to avoid magic numbers.

src/views/Confirmation/Confirmation.tsx
```diff
@@ -55,14 +61,17 @@ const Confirmation: React.FC<Props> = ({ deposit, onClose }) => {
...
 
   return (
     <Layout>
       <Wrapper>
         <Header>
           <Heading>Deposit succeeded</Heading>
           <SubHeading>
-            Your funds will arrive in {getConfirmationDepositTime()}
+            Your funds will arrive in {confirmationTime || "loading"}
           </SubHeading>
```
Tulun, line 75: I would probably adjust this string to be like: `Your fund will arrive in confirmationTime()` || "Loading time estimate"

Tulun (approved):
Couple nits, overall good.

src/utils/bridge.ts
mrice32: Great point -- done!

src/views/Confirmation/Confirmation.tsx
mrice32: Good point -- fixed!

PR #147, opened 2022-06-28
feat: Desktop Referral widget

Tulun:
Screen Shot 2022-06-28 at 2 28 13 PM

*Overview**

Styles for Referral Widget.
Note: Only for Desktop, responsive CSS coming.

**Acceptance Criteria**
Should match design


src/views/Rewards/comp/RewardReferral/RewardReferral.tsx
```diff
@@ -1,45 +1,105 @@
...
+              <LearnMoreText>
+                Learn more <ArrowUpRight />
+              </LearnMoreText>
+            </ConnectRow>
+          )}
+        </ReferralLinkBlock>
+        <ReferralTierBlock>
+          <TierSmHeader>Current referral tier</TierSmHeader>
+          <TierHeader>Bronze</TierHeader>
+          <Stepper currentStep={2} numSteps={5} />
```
daywiss, line 77: is it right these values are going to eventually be calculated by some api call or client?

src/components/Stepper/useStepper.tsx
```diff
@@ -0,0 +1,54 @@
+import { useState, useEffect, useCallback } from "react";
+import { StyledStepperItem } from "./Stepper.styles";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faCheck } from "@fortawesome/free-solid-svg-icons";
+
+export default function useStepper(currentStep: number, numSteps: number) {
```
daywiss, line 6: would be great to get some comments on what this hook does and how it works, the logic is hard to follow

daywiss (commented):
just want to get some clarity on what the stepper function is doing since theres no way to really test this.  it seems though like a good candidate for a unit test

src/views/Rewards/comp/RewardReferral/RewardReferral.tsx
Tulun: Yup. This is just a stub.

NumSteps should be static (it's just the amount in the stepper itself), but current step would be calculated based on what tier they are on.

src/components/Stepper/useStepper.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [stepItems, setStepItems] = useState<JSX.Element[]>([]);
+  const createStepsItems = useCallback(() => {
+    const items: JSX.Element[] = [];
+    const midPoint = Math.round(numSteps / 2);
+    for (let i = 0; i < numSteps; i++) {
+      const completed = i + 1 < currentStep;
```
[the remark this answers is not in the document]
Tulun: Sure, I can readjust this to start at i = 1.

nicholaspai:
Some QA suggestions

- Boxes are too close to side edge of screen in non-full screen mode 
Screen Shot 2022-06-29 at 15 10 10
- Clicking copy doesn't just copy the URL as expected, it copies this: "Across blurb www.across.to/referrer=0x...."

Otherwise LGTM!

src/components/Stepper/useStepper.tsx
daywiss: i see so this kind of does css dynamically based on the number of steps

daywiss (approved):

src/components/Stepper/useStepper.tsx
amateima: I see this hook and the associated CSS too complicated for what we want to achieve. I think we can have the same result with some relatively simple CSS as [showcased here](https://jsfiddle.net/dyjt05gf/). Sorry if I didn't fully understand it and my example doesn't cover your needs.

Screenshot 2022-06-29 at 22 36 00

src/views/Rewards/comp/ConnectTableOverlay/ConnectTableOverlay.styles.tsx
```diff
@@ -0,0 +1,23 @@
+import styled from "@emotion/styled";
+import { SecondaryButtonWithoutShadow } from "components/Buttons";
+import { COLORS } from "utils";
+
+export const Overlay = styled.div`
+  position: absolute;
+  width: 100%;
+  max-width: 1406px;
+  margin-top: -200px;
+  left: 11%;
```
amateima, line 10: I don't see why this overlay should be relative to the whole < Wrapper />. This overlay and the Referrals table should be contained inside a separate container that will also have `display: relative`. This way you won't need to define `left`, `max-width` and `height` attributes. Also, please place CSS attributes in the order of their importance. `display` and `height` should be above `text-align` and `background-color`.

An example of weird behaviour caused by the `left` usage

Screenshot 2022-06-29 at 22 40 10

amateima (changes requested):

src/components/Stepper/useStepper.tsx
Tulun: Nice. I think it is too simple but I might be able to adapt that and avoid the use of pseudoelements.

amateima (approved):

PR #154, opened 2022-07-06
feat: Add delimiter to referrer tag

nicholaspai:
To make it easier for clients to identify referred transactions, add a delimitter in front of referrer address. By adding in front, this allows referrer-identification logic to be backwards compatible

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -677,3 +677,5 @@ const getQueriesTable = () => {
 };
 
 export const queriesTable = getQueriesTable();
+
+export const referrerDelimitterHex = "0xaaaabbbbccccdddd";
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -678,4 +678,4 @@ const getQueriesTable = () => {
 
 export const queriesTable = getQueriesTable();
 
-export const referrerDelimiterHex = "0xaaaabbbbccccdddd";
+export const referrerDelimiterHex = "0xdooddeaffeeddeadbeef";
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -678,4 +678,4 @@ const getQueriesTable = () => {
 
 export const queriesTable = getQueriesTable();
 
-export const referrerDelimitterHex = "0xaaaabbbbccccdddd";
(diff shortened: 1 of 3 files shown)


src/utils/constants.ts
```diff
@@ -677,3 +677,5 @@ const getQueriesTable = () => {
 };
 
 export const queriesTable = getQueriesTable();
+
+export const referrerDelimitterHex = "0xaaaabbbbccccdddd";
```
Tulun, line 681: Would this cause major issues if this string changed without notice?

src/utils/format.ts
```diff
@@ -88,7 +89,7 @@ export function stringToHex(value: string) {
 // appends hex tag to data
 export function tagHex(dataHex: string, tagHex: string) {
   assert(ethers.utils.isHexString(dataHex), "Data must be valid hex string");
-  return ethers.utils.hexConcat([dataHex, tagHex]);
+  return ethers.utils.hexConcat([dataHex, referrerDelimitterHex, tagHex]);
```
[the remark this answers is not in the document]
nicholaspai: good call

src/utils/constants.ts
nicholaspai: minor issues i think: the API that @amateima built would just have to be aware of all historical delimitters we've used in past.

Tulun: Should we have a way of tracking this then?

daywiss: delimitter  should be delimiter

daywiss (commented):
im not totaly convinced this is the way to go.  you still need to support the previous way, especially if you expect third parties to be using their own referral making code

src/utils/constants.ts
nicholaspai: what do you mean by "delimitter should be delimiter" @daywiss ?

amateima (approved):

src/utils/constants.ts
nicholaspai: oh lol

mrice32 (approved):
LGTM!

amateima (changes requested):
Screenshot 2022-07-08 at 12 25 08

It breaks the send tab

amateima (approved):

PR #156, opened 2022-07-07
feat: tooltip modals. Swapped API url.

Tulun:
Screen Shot 2022-07-07 at 3 55 12 PM

Modals for Stepper.

Note: need content for some of the info stuff below still.

### DIFF
--- a/src/hooks/useReferralSummary.ts
+++ b/src/hooks/useReferralSummary.ts
@@ -52,6 +52,6 @@ export function useReferralSummary(account: string) {
  */
 async function getReferrals(account: string) {
   return axios.get<ReferralsSummary>(
-    `${rewardsApiUrl}/referrals/summary?address=${account}`
+    `${rewardsApiUrl}referrals/summary?address=${account}`
   );
 }
--- a/src/hooks/useReferralSummary.ts
+++ b/src/hooks/useReferralSummary.ts
@@ -52,6 +52,6 @@ export function useReferralSummary(account: string) {
  */
 async function getReferrals(account: string) {
   return axios.get<ReferralsSummary>(
-    `${rewardsApiUrl}referrals/summary?address=${account}`
+    `${rewardsApiUrl}/referrals/summary?address=${account}`
   );
 }
--- a/src/views/Rewards/comp/RewardReferral/RewardReferral.tsx
+++ b/src/views/Rewards/comp/RewardReferral/RewardReferral.tsx
@@ -1,8 +1,6 @@
 import { useMemo, useState } from "react";
 import ReactDOMServer from "react-dom/server";
 import { ethers } from "ethers";
-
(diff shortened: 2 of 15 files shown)


src/views/Rewards/comp/StepperWithTooltips/useStepperWithTooltips.tsx
```diff
@@ -0,0 +1,145 @@
...
+export default function useStepperWithTooltips(
+  currentStep: number,
+  numSteps: number,
+  tooltips: TooltipProps[],
+  tooltipId: string
+) {
+  const [stepItems, setStepItems] = useState<JSX.Element[]>([]);
+  const createStepsItems = useCallback(() => {
+    const items: JSX.Element[] = [];
+    for (let i = 1; i <= numSteps; i++) {
```
nicholaspai, line 25: why start at 1?

src/views/Rewards/comp/RewardReferral/RewardReferral.tsx
```diff
@@ -116,10 +118,49 @@ const RewardReferral: React.FC<Props> = ({
...
+                title: "Bronzer tier",
+                body: "Requires a bridge volume of more than $50,000 or 3 unique referrals. ",
+              },
+              {
+                title: "Silver tier",
+                body: "Stub for silver tier",
+              },
+              {
+                title: "Gold tier",
+                body: "Stub for gold tier",
```
nicholaspai, line 140: Totally to leave stubs for now thanks for this!

nicholaspai (approved):
This is awesome great work!

amateima (approved):

src/hooks/useReferralSummary.ts
```diff
@@ -52,6 +52,6 @@ export function useReferralSummary(account: string) {
  */
 async function getReferrals(account: string) {
   return axios.get<ReferralsSummary>(
-    `${rewardsApiUrl}/referrals/summary?address=${account}`
+    `${rewardsApiUrl}referrals/summary?address=${account}`
```
daywiss, line 55: this change implies that the url provided in the env has a trailing slash? its probably safer to not assume this unless you are explicitly checking this, as it will subtly break when the url is defined wrong

daywiss (approved):

src/hooks/useReferralSummary.ts
Tulun: Fair, changed.

src/views/Rewards/comp/StepperWithTooltips/useStepperWithTooltips.tsx
Tulun: As you're comparing the length of the steps, it's easier to start index at 1 instead of 0 to simplify the logic.

PR #159, opened 2022-07-12
fix: fix issue with remove liquidity when decimals are input

daywiss:

Errors were not being cleared when a valid number was detected in the remove liquidity form input. Now they are.


amateima (approved):
I'm wondering if this validation function should exist in the `AddLiquityForm` component as well. We don't have one as far as I could see

nicholaspai (approved):

PR #162, opened 2022-07-14
improve: update the header and footer UIs

amateima:
Screenshot 2022-07-14 at 03 28 01
Screenshot 2022-07-14 at 03 27 08
Screenshot 2022-07-14 at 03 26 52
Screenshot 2022-07-14 at 03 26 42


src/components/Wallet/Wallet.tsx
```diff
@@ -41,18 +40,15 @@ const Wallet: FC<Props> = ({ setOpenSidebar }) => {
...
-          <BalanceWallet>0 ACX</BalanceWallet>
-        </div>
-        {account && (
-          <Account>{ensName ?? shortenAddress(account, "...", 4)}</Account>
-        )}
-      </BalanceButton>
-    </Wrapper>
+    <BalanceButton onClick={() => setOpenSidebar(true)}>
+      <Logo />
+      <BalanceWallet>0 ACX</BalanceWallet>
```
nicholaspai, line 47: Just curious, do you use a linter on the front end that automatically styles this part?

nicholaspai (approved):
LGTM this looks super clean

src/components/Wallet/Wallet.tsx
amateima: We use a linter, but I don't think it caused many changes in this PR.

PR #163, opened 2022-07-14
feat: rework tooltip and popovers

amateima:
Improve the UI implementation of tooltips/popovers

One particular thing I paid attention is preventing tooltips to overflow the screen by shifting or flipping them

Examples:

- Has room below it:

Screenshot 2022-07-14 at 18 16 11

- Flip vertically if there's no enough space below

Screenshot 2022-07-14 at 18 16 23

- There's enough space in the right side

Screenshot 2022-07-14 at 18 15 47

- Translate to left if there's not enough space 

Screenshot 2022-07-14 at 18 15 58


src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useState } from "react";
+import { usePopper } from "react-popper";
```
nicholaspai, line 2: This is a nice tool, pretty small too!

PR #164, opened 2022-07-15
improve(api): Set Cache-Control header to reduce Serverless Invocations

nicholaspai:
Currentl Vercel serverless invocations 
Screen Shot 2022-07-15 at 13 45 11

Theoretically this # can be reduced by caching responses. Caching the /suggested-fees responses doesn't make sense since the `amount` request parameter will usually be different, but we should be able to cache /limits which does seem like the more requested route.

Idea to cache came from this document which I found by looking up "vercel serverless functions 429"
https://vercel.com/support/articles/what-should-i-do-if-i-receive-a-429-error-on-vercel

### DIFF
--- a/api/limits.js
+++ b/api/limits.js
@@ -149,6 +149,11 @@ const handler = async (request, response) => {
       ).toString(),
     };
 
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
LGTM!

PR #169, opened 2022-07-18
feat(api): Add simple jest test that we can include in CI

nicholaspai:
## Motivation

I accidentally merged a PR that included a bug in the `/api` path that didn't affect the React application, however it did break the Vercel serverless function. This bug would have been prevented in Vercel CI if we had included a simple unit test for the /api that would have attempted to import files from. I'll leave it to a follow up PR to actually test /limits and /suggested-fees, but the unit test in this PR will at least prevent any compile-time errors, which are exacerbated because the files are in .js not .ts.


package.json
```diff
@@ -52,6 +52,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
+    "test-api": "jest",
```
nicholaspai, line 55: We can add `test-api` to the Vercel "build" command to include the API tests in CI

test/api/main.test.js
```diff
@@ -0,0 +1,11 @@
+const handler = require("../../api/health");
+
+test("serverless function is alive", async () => {
+  const response = {};
+  response.status = jest.fn().mockReturnValue(response);
```
nicholaspai, line 5: jest mocked functions are awesome!

mrice32 (commented):
This is awesome! Great to see tests added for these! 

I'm just curious what problem this endpoint and this test is meant to solve.

If the idea is that it's an example for how to build tests for the serverless API, I think that makes sense 100%. I just don't think this test will catch many (if any) errors, and probably serves more as an example.

Thoughts?

mrice32:
> Importantly, this test will fail if any of the imports in the /api are incorrect. For example, I pushed code last week that tried to import from a folder "utils.js" which was actually renamed to "_utils.js", and this test would have caught that simple fix. This is a very crude way to catch those issues

@nicholaspai this may be a really dumb question, but I don't see how this test would fail if imports in suggested-fees.js or limits.js were wrong? The reason I say that is that I don't think those files are in the import tree for this test, so they would never be touched in the course of this test (you import the health file _directly_).

mrice32 (commented):
https://github.com/across-protocol/frontend-v2/pull/169#issuecomment-1189551644

nicholaspai:
> > Importantly, this test will fail if any of the imports in the /api are incorrect. For example, I pushed code last week that tried to import from a folder "utils.js" which was actually renamed to "_utils.js", and this test would have caught that simple fix. This is a very crude way to catch those issues
> 
> @nicholaspai this may be a really dumb question, but I don't see how this test would fail if imports in suggested-fees.js or limits.js were wrong? The reason I say that is that I don't think those files are in the import tree for this test, so they would never be touched in the course of this test (you import the health file _directly_).

Great point. Do you think we should try to import functions from limits.js and suggestedFees.js into health.js? Or just in the test? I'm curious if you have ideas for testing this particular case where the api/ .js files don't build properly

mrice32 (approved):
LGTM!

PR #173, opened 2022-07-20
fix: temporary disable refetch interval

amateima:
Temporary disable the automatic refetch interval for rewards API


pxrl (approved):
OK based on description from Alex - this is a temporary measure to reduce load.

PR #178, opened 2022-07-20
fix: fix recieve amount to match ui

daywiss:


daywiss:
this turns out to be an issue with rounding.will re-open once we have a solution

PR #180, opened 2022-07-21
fix: change formatting style for numbers

daywiss:
this pr is updated with a new strategy proposed by korpi:
if number is less than 1, minimum use significant digits
if greater than one use maximum fraction digits
for numbers 1000 or greater, we drop any fractional values
This seems to work well for big and small values

I think with this technique we dont need any tool tips since we both minimize what we show and show necessary decimals when numbers are really small 

new behavior
image
image

old behavior
image
image

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -51,15 +51,23 @@ export function shortenTransactionHash(hash: string): string {
   return `${hash.substring(0, 5)}...`;
 }
 
+// for number less than 1, this will ensure at least 1 digit is shown ( not rounded to 0)
 export const smallNumberFormatter = (num: number) =>
   new Intl.NumberFormat("en-US", {
     minimumSignificantDigits: 1,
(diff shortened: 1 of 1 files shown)


mrice32:
Screen Shot 2022-07-21 at 3 25 36 PM
Is this a problem?

mrice32 (commented):
https://github.com/across-protocol/frontend-v2/pull/180#issuecomment-1191852255

daywiss:
> Is this a problem?

I think this is a bad example because the fees actually probably are 0 since btc only has 6 decimal resolution

mrice32:
I think it’s 8 decimal resolution.

amateima (approved):

daywiss:
> I think it’s 8 decimal resolution.

Yea you are right, so it might not be 0, but try with the new configuration, it should work better @mrice32

PR #182, opened 2022-07-21
fix: decimal parsing of diffent tokens in transfer

daywiss:
parsing token amounts was not using decimal value of token on transfer history page


mrice32 (approved):
LGTM!

PR #183, opened 2022-07-21
fix: fix pagination issues

mrice32:
This fixes a few issues:
1. There was a hardcoded pagination limit of 30, so users couldn't see more than 30 referrals.
2. The elements per page selector disappeared once a user selected a page size that contained more than their total number of elements. So they couldn't reverse the action. This change allows the user to _always_ select a page size even if they don't need it.

To accomplish this, I had to change the way we query a bit. We lazily load each page of the referrals. In the future, we may _not_ want to do this to reduce the total number of queries, but it seemed like the simplest approach for now.

In the code, this also fixes some issues with the way the hooks were built. We were previously using a constant reactQuery key. This was dangerous because we were using account in the query, so theoretically the user could switch accounts and see stale data from the previous account from the cache.

### DIFF
--- a/src/utils/query-keys.ts
+++ b/src/utils/query-keys.ts
@@ -68,3 +68,27 @@ export function allowanceQueryKey(
 ) {
   return ["allowance", chainId, tokenSymbol, owner, spender, blockNumber];
 }
+
+/**
+ * Generates query keys for react-query `useQuery` hook, used in the `useReferrals` hook.
+ * @param account  The address that referrals are being queried for.
+ * @param limit The limit on the number of results that are returned (page size).
+ * @param offset The number of elements to omit before returning results.
+ * @returns An array of query keys for react-query `useQuery` hook.
+ */
+export function referralsQueryKey(
+  account: string,
+  limit: number,
+  offset: number
+) {
+  return ["referrals", account, limit, offset];
+}
+
+/**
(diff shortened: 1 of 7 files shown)


src/utils/query-keys.ts
```diff
@@ -68,3 +68,27 @@ export function allowanceQueryKey(
 ) {
   return ["allowance", chainId, tokenSymbol, owner, spender, blockNumber];
 }
+
+/**
```
mrice32, line 72: We should _always_ use keys with all dependencies to avoid issues with stale cached values being used.

nicholaspai (approved):
THis is a good refactor I have no issues, note I have not QA'd only looked at code

mrice32:
> THis is a good refactor I have no issues, note I have not QA'd only looked at code

I did a good bit of self-QA for this with a wallet that used multiple pages. So I'm confident that it works and preserves all transactions. Just due to the fact that this is running in prod, I'm going to go ahead and merge. @daywiss will take a closer look at the code tonight.

src/utils/query-keys.ts
daywiss: i agree, great catch

daywiss (commented):
did not see any issues during testing, lg, nice improvements

PR #184, opened 2022-07-22
fix: fix referral summary query key

mrice32:
This PR fixes the referral summary query key so that it includes the account parameter to avoid cache collisions between accounts.


daywiss (approved):
good find

amateima (approved):

amateima (approved):

PR #188, opened 2022-07-25
improve: pools route should return full pool client object

nicholaspai:
DefiLlama needs additional information like TVL in USD we should export full object from client


daywiss (commented):
If you are just returning pool state, dont bother nesting the key and just return pool state directly

daywiss (approved):

PR #189, opened 2022-07-26
feat: align Tables; Refactor Table to be exportable and Baseline

Tulun:
Overview:

More closely align the Referral and Transaction Tables. Also refactored the baseline component to be more generalizable and use it as an Example. If you need to go beyond the Baseline, take the exported styled components and build as needed.
Screen Shot 2022-07-26 at 2 56 44 PM


src/App.tsx
```diff
@@ -12,6 +12,16 @@ assert(
...
+assert(
+  process.env.REACT_APP_CHAIN_137_PROVIDER_URL,
+  "REACT_APP_CHAIN_137_PROVIDER_URL must be defined."
+);
+
+assert(
+  process.env.REACT_APP_CHAIN_42161_PROVIDER_URL,
+  "REACT_APP_CHAIN_42161_PROVIDER_URL must be defined."
+);
+
```
Tulun, line 24: I had issues with my TX table until I added these envs, we should have a check on these imo.

src/components/Pagination/Pagination.styles.tsx
```diff
@@ -2,43 +2,41 @@ import { keyframes } from "@emotion/react";
 import styled from "@emotion/styled";
```
amateima, line 2: From what I see, the styles from `RewardTablePagination.styles.tsx` are identical to the components from this file. Can we delete most of the components from `RewardTablePagination.styles.tsx` and use the ones from here?

src/views/Transactions/TransactionsTable/createTransactionTableJSX.tsx
```diff
@@ -103,7 +98,6 @@ function formatTransactionRows(
     };
 
     const symbol: ICell = {
-      size: "xs",
       value: (
         <>
           <TableLogo src={token?.logoURI} alt={`${token?.name}_logo`} />{" "}
```
amateima, line 103: Screenshot 2022-07-27 at 14 33 45

Let's display the token symbol here

amateima (changes requested):
Screenshot 2022-07-27 at 14 32 37

I'll ask Tim if we need to keep the accordion row style or use the same behaviour as the rewards table

src/components/Pagination/Pagination.styles.tsx
Tulun: Yup, good call, as the styles are the default now it can be removed

src/views/Transactions/TransactionsTable/createTransactionTableJSX.tsx
Tulun: > Screenshot 2022-07-27 at 14 32 37
> 
> I'll ask Tim if we need to keep the accordion row style or use the same behaviour as the rewards table

That's fine -- I just left it as is as the data we're displaying is a bit different, but if we wanna scroll instead that's fine.

Tulun: @amateima chatted with Tim, he's happy with the current look, may do some redesign at a later date.

daywiss (approved):
did not notice any issues in my testing, code lgtm, nice work

amateima (approved):
Looks good, but let's increase the bottom padding of this wrapper because the page size dropdown doesn't have enough room. I think it's 4rem on the Rewards Page

Screenshot 2022-07-28 at 11 42 14

PR #190, opened 2022-07-27
feature: allow ref to be used; add catch for when both exist.

Tulun:
ref is an allowable queryParam for referrer.

Note: If for some reason both exist I have defaulted to referrer. If anyone has thoughts on this, let me know.

### DIFF
--- a/src/hooks/useBridge.ts
+++ b/src/hooks/useBridge.ts
@@ -1,5 +1,4 @@
 import { ethers } from "ethers";
-import { useQueryParams } from "./useQueryParams";
 import { FormStatus, useSendForm } from "./useSendForm";
 import { useBalanceBySymbol } from "./useBalance";
 import { useBridgeFees } from "./useBridgeFees";
@@ -21,6 +20,7 @@ import {
   trackEvent,
   formatUnits,
 } from "utils";
+import useReferrer from "./useReferrer";
 
 enum SendStatus {
   IDLE = "idle",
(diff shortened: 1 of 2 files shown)


src/hooks/useBridge.ts
```diff
@@ -36,7 +36,13 @@ type SendError =
 
 export function useBridge() {
   const config = getConfig();
-  const { referrer } = useQueryParams();
+  const { referrer, ref: refParam } = useQueryParams();
+  let r: string | undefined = undefined;
```
daywiss, line 40: Why not r = referrer || ref?  thats a pretty standard way of setting fallback

src/hooks/useBridge.ts
```diff
@@ -36,7 +36,13 @@ type SendError =
...
+  // If ref and referrer params exist, prefer referrer param.
+  // Not likely to happen but should have a catch if we get a bad link.
+  // TODO? Test which of these is a good value?
```
nicholaspai, line 43: i agree with this TODO: we should default to referrer

nicholaspai (approved):
I think we should implement the TODO! After that LGTM

src/hooks/useBridge.ts
Tulun: Done.

src/hooks/useReferrer.ts
```diff
@@ -0,0 +1,10 @@
+import { useQueryParams } from "./useQueryParams";
+
+export default function useReferrer() {
+  const { referrer, ref: refParam } = useQueryParams();
```
daywiss, line 4: Is ref a reserved word or something, why does it need to be renamed?

src/hooks/useReferrer.ts
```diff
@@ -0,0 +1,10 @@
...
+  // If ref and referrer params exist, prefer referrer param.
+  // Not likely to happen but should have a catch if we get a bad link.
+  // TODO? Test which of these is a good value?
+  let r = referrer || refParam;
```
daywiss, line 8: Just return the value here

daywiss (approved):
Lg just minor points

PR #192, opened 2022-07-28
feat: Implement Rewards Banner warning behind env var.

Tulun:
Screen Shot 2022-07-28 at 12 22 45 PM

Banner hidden behind REACT_APP_REWARDS_BANNER_WARNING = true. Only shows on /rewards page.

### DIFF
--- a/src/components/Banner/Banner.styles.tsx
+++ b/src/components/Banner/Banner.styles.tsx
@@ -1,0 +1,27 @@
+import styled from "@emotion/styled";
+
+export const Wrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 0 30px;
+  height: 60px;
+  color: var(--color-gray);
+  background-color: var(--color-error);
+  border-bottom: 1px solid var(--color-gray);
+
+  & button {
+    background-color: inherit;
+    font-size: inherit;
+    color: var(--color-gray);
+    text-decoration: underline;
+    cursor: pointer;
+    border: none;
+    padding: 0;
+    margin: 0;
+    display: inline-flex;
+    &:hover {
+      color: var(--color-black);
+    }
+  }
+`;
--- a/src/components/Banner/Banner.styles.tsx
+++ b/src/components/Banner/Banner.styles.tsx
@@ -10,6 +10,12 @@ export const Wrapper = styled.div`
   background-color: #202024;
   border-bottom: 1px solid #3e4047;
   font-size: ${16 / 16}rem;
+  position: sticky;
+  width: 100%;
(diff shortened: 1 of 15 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
...
+    if (!container.current) {
+      // we know this to always be defined.
+      const root = document.getElementById("root") as HTMLDivElement;
+      const div = document.createElement("div");
+      div.id = "banner";
+      root.insertBefore(div, root.firstChild);
+      container.current = div;
+    }
+  }, []);
+  if (!container.current) {
```
nicholaspai, line 20: Out of curiosity, does it increase performance to exit earlier if !container above where it's first defined?

nicholaspai (approved):
LGTM whenever we want to start displaying the warning

src/components/Banner/Banner.tsx
Tulun: Hm it might, not sure if it would be a huge diff.

daywiss:
@nicholaspai @Tulun i've added the env to enable the banner just for preview and development

daywiss:
looks good, but just fyi the banner at the top extends the length of the page causing a scrollbar to show up. this was the same issue with the navbar that was fixed recently.  you could fix in this pr or make an issue for it, up to you

daywiss (approved):
only issue is that banner causes page length to be beyond window height, you can fix in this pr or make an issue for it since its pretty minor

src/components/Banner/Banner.styles.tsx
```diff
@@ -0,0 +1,29 @@
...
+    padding: 0 10px;
+  }
+  svg {
+    margin-right: 16px;
+  }
+
+  span {
+    padding: 10px 0;
+    @media ${QUERIES.tabletAndDown} {
+      font-size: ${13 / 16}rem;
```
amateima, line 23: this should be 14, not 13

src/components/Banner/Banner.styles.tsx
```diff
@@ -0,0 +1,29 @@
...
+    }
+    @media ${QUERIES.mobileAndDown} {
```
amateima, line 25: let's use the new breakpoints from the design. From what I could see in the design, the styles are changed only on mobile ( <= 428px )

amateima (changes requested):

src/components/Banner/Banner.styles.tsx
Tulun: I've changed this but it won't look good at all breakpoints.

amateima (approved):

PR #193, opened 2022-07-29
fix: hide pagination when not connected.

Tulun:
Add condition where pagination component only shows when connected, as it's not clickable anyway.

Signed-off-by: Tulun <jaykiraly@gmail.com>


daywiss (approved):
nice

PR #195, opened 2022-08-02
feat: remove BADGER from pool and send.

Tulun:
Badger route removed.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -612,6 +612,12 @@ export const relayerFeeCapitalCostConfig: {
     cutoff: ethers.utils.parseUnits("5000").toString(),
     decimals: 18,
   },
+  BADGER: {
+    lowerBound: ethers.utils.parseUnits("0.0003").toString(),
+    upperBound: ethers.utils.parseUnits("0.001").toString(),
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

src/utils/constants.ts
```diff
@@ -619,12 +612,6 @@ export const relayerFeeCapitalCostConfig: {
     cutoff: ethers.utils.parseUnits("5000").toString(),
     decimals: 18,
   },
-  BADGER: {
-    lowerBound: ethers.utils.parseUnits("0.0003").toString(),
-    upperBound: ethers.utils.parseUnits("0.001").toString(),
-    cutoff: ethers.utils.parseUnits("5000").toString(),
-    decimals: 18,
-  },
```
daywiss, line 627: you could leave the constants file alone, it wont hurt anything to keep the definitions in since it doesnt control whats visible in the UI.  will just save us work if we ever have to re-enable badger

daywiss (approved):

src/utils/constants.ts
Tulun: Sounds good!

PR #198, opened 2022-08-03
fix: copy url is more proper link to fix youtube encoding

Tulun:
Youtube has strict encoding to create a link on its platform. Adhere to that standard, which should work in other places just fine.


daywiss (approved):

nicholaspai (approved):
LGTM, this matches the docsite example as well!  https://docs.across.to/v2/how-to-use-across/referral-program#how-do-i-refer-users-with-a-custom-referral-url

PR #199, opened 2022-08-03
docs: improve "getting-started" experience

dohaki:
While trying to run the app locally, I noticed some things that this tries to improve:
- more inviting README file
- example .env file to easily get started
- refined docs on which env vars are required/optional


.env.example
```diff
@@ -0,0 +1,61 @@
+#########################
+### REQUIRED ENV VARS ###
+#########################
+
+# Chain ID where HubPool contract is deployed
+REACT_APP_HUBPOOL_CHAINID=1
```
nicholaspai, line 6: Should this be defaulted to 1 in the app? I can't imagine that this is ever changed in practice (i'm not sure app would work on testnets currently)

src/utils/constants.ts
```diff
@@ -398,11 +398,18 @@ assert(
...
 assert(
   process.env.REACT_APP_REWARDS_API_URL,
   "Missing process.env.REACT_APP_REWARDS_API_URL"
 );
+assert(
+  process.env.REACT_APP_CHAIN_137_PROVIDER_URL,
+  "REACT_APP_CHAIN_137_PROVIDER_URL must be defined."
+);
+assert(
+  process.env.REACT_APP_CHAIN_42161_PROVIDER_URL,
```
nicholaspai, line 410: Its strange we only check for 137/42161. Should we also check for 1/288/10?

nicholaspai (commented):
Looks great, will wait for @Tulun to review

src/utils/constants.ts
amateima: @nicholaspai these two providers were overwritten with custom provider urls because they had poor performance on the tx history page while fetching events. We dont have to check for 1, 288, 10 because we're using the default Infura providers for them

CONTRIBUTING.md
```diff
@@ -0,0 +1,133 @@
...
+
+```bash
+yarn start
+```
+
+### Build production bundle
+
+```bash
+yarn build
+```
```
amateima, line 28: Personally I'd put these lines back to the readme file, but I don't have a strong opinion about it

amateima (approved):

.env.example
```diff
@@ -0,0 +1,61 @@
+#########################
+### REQUIRED ENV VARS ###
```
Tulun, line 2: I would probably comment on each of these potential good values (or set of expected values ie: 
true | false)  given this is an example.

src/utils/constants.ts
Tulun: Maybe we should document this in the README @amateima ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,133 @@
...
+This section outlines best practices for internal contributors, these are people who have direct access to the repository.
+We currently do not have a process for external contributors but will add a section for that when applicable.
+
+### Best Practices
+
+- It's recommended to branch directly from the repo for easier collaboration. This doc will assume you branch from `origin/master`.
+- All changes to master should first be a pull request.
+- PR titles should conform to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
+- Branch names should ideally be generated by [Shortcut](www.shortcut.com) issues, but otherwise there's no particular naming convention.
+- Try to keep PRs focused on a single feature, bug, or issue.
```
Tulun, line 95: We no longer use shortcut.

Tulun (changes requested):
Couple comments.

CONTRIBUTING.md
dohaki: There is still a [`Getting started` section in the README](https://github.com/across-protocol/frontend-v2/blob/a83d6d541fbc832aafe7c063a66775d9bc7d41dd/README.md#getting-started) that contains most of these lines (except the build step). Should I still move these back?

daywiss: i can understand the reason for putting this into the contributors section, because this is an app not a library, the readme should probably be high level for end users, while contributors are the developers.

.env.example
```diff
@@ -0,0 +1,61 @@
+#########################
```
daywiss, line 1: im wary about having 2 sources of truth, one in the example and one in the contributor file. it just makes it error prone to maintain as envs change.  You could just use this as the source of truth for documentation and link to it in the docs

.env.example
dohaki: I changed it so that we default to 1 if this env var is not set. Therefore this env var is not required but optional now.

src/utils/constants.ts
dohaki: I added some more information in the comments for these env vars. See https://github.com/across-protocol/frontend-v2/pull/199/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR15-R21 and let me know if this suffices.

CONTRIBUTING.md
dohaki: Ok, I moved it back into the README and removed it from the CONTRIBUTING doc for now.

.env.example
dohaki: I added some more information on potential good values in the example env var file https://github.com/across-protocol/frontend-v2/pull/199/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR5-R21. Let me know if this is what thought you of.

CONTRIBUTING.md
dohaki: Changed to linear

.env.example
dohaki: Yea good point. I removed the env var docs from the README/CONTRIBUTING and will keep the example file as the single source of truth.

daywiss (approved):

james-a-morris (approved):
This PR is succinct. 

Creating an example env file will help new developers/contributors better onboard into this repository. 

I agree with @dohaki 's decision to place the instructions to build the frontend-v2 in a separate contributing markdown file. Additionally, the decision to remove information related to contributions from the main readme feels like a great way to reduce redundant info from the codebase.

Finally, @dohaki has further asserted the optional/required nature of several env variables (including the Hubpool Chain Id) within the contstants file. I think that's a vital step toward keeping the docs/codebase in sync.

src/utils/constants.ts
nicholaspai: Great stuff @dohaki

nicholaspai (approved):
LGTM

.env.example
Tulun: Yep, I might instead of saying "alchemy" I would put like: https://mainnet.infura.io/v3/<infura-key>

Tulun (approved):

.env.example
dohaki: done

PR #200, opened 2022-08-09
test(integration): rudimentary cypress tests

dohaki:
This adds rudimentary integration tests and the required cypress setup. The test scenarios are still very minimal and use a mocked web3 provider for now. The scope was to make sure that

- each route renders
- basic UI interactions work

We can add more/refine tests as we see fit in the future.

Integration with GitHub Actions will follow in a separate PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ yarn-error.log*
 # cypress
 /cypress/videos
 /cypress/downloads
+/cypress/screenshots
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ yarn-debug.log*
 yarn-error.log*
 
 .vercel
+
+# cypress
+/cypress/videos
+/cypress/downloads
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,0 +1,7 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: "http://localhost:3000",
+  },
+});
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -2,6 +2,7 @@ import { defineConfig } from "cypress";
 
 export default defineConfig({
   e2e: {
+    // Override by setting `CYPRESS_BASE_URL`
     baseUrl: "http://localhost:3000",
   },
 });
--- a/cypress/e2e/transactions.cy.ts
+++ b/cypress/e2e/transactions.cy.ts
@@ -1,0 +1,9 @@
+describe("transactions", () => {
+  beforeEach(() => {
+    cy.visit("/transactions", { withInjectedMockProvider: true });
+  });
+
+  it("render in initial state", () => {
+    cy.dataCy("connect-wallet").should("be.visible");
+  });
+});
--- a/cypress/e2e/transactions.cy.ts
+++ b/cypress/e2e/transactions.cy.ts
@@ -1,6 +1,6 @@
 describe("transactions", () => {
   beforeEach(() => {
-    cy.visit("/transactions", { withInjectedMockProvider: true });
+    cy.visit("/transactions");
   });
 
   it("render in initial state", () => {
--- a/cypress/e2e/wallet.cy.ts
+++ b/cypress/e2e/wallet.cy.ts
@@ -1,0 +1,18 @@
+describe("wallet", () => {
+  beforeEach(() => {
+    cy.visit("/", { withInjectedMockProvider: true });
+  });
+
+  it("display ACX balance for connected wallet", () => {
+    cy.connectInjectedWallet("connect-wallet");
+
+    cy.dataCy("acx-balance").should("be.visible");
+  });
+
+  it("display warning for unsupported network", () => {
+    cy.visit("/", { withInjectedMockProvider: true, chainId: 1234 });
+    cy.connectInjectedWallet("connect-wallet");
+
+    cy.dataCy("unsupported-network").should("be.visible");
+  });
+});
--- a/cypress/e2e/wallet.cy.ts
+++ b/cypress/e2e/wallet.cy.ts
@@ -1,16 +1,13 @@
 describe("wallet", () => {
-  beforeEach(() => {
-    cy.visit("/", { withInjectedMockProvider: true });
-  });
-
   it("display ACX balance for connected wallet", () => {
+    cy.visit("/");
     cy.connectInjectedWallet("connect-wallet");
 
     cy.dataCy("acx-balance").should("be.visible");
   });
 
   it("display warning for unsupported network", () => {
-    cy.visit("/", { withInjectedMockProvider: true, chainId: 1234 });
+    cy.visit("/", { chainId: 1234 });
     cy.connectInjectedWallet("connect-wallet");
 
     cy.dataCy("unsupported-network").should("be.visible");
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1,0 +1,62 @@
+/// <reference types="cypress" />
+
+import {
+  MockWeb3Provider,
+  MockProviderOptions,
+  DEFAULT_ACCOUNT,
+} from "../utils/ethereum";
+
+/**
+ * Overwrite default `visit` command to allow injecting web3 provider to
+ * `window.ethereum`.
+ */
+Cypress.Commands.overwrite(
+  "visit",
+  (
+    original,
+    url: string | Partial<Cypress.VisitOptions>,
+    options?: Partial<Cypress.VisitOptions>
+  ) => {
+    return original({
+      ...options,
+      url: url as string,
+      onBeforeLoad: (win) => {
+        if (options?.withInjectedMockProvider) {
+          win.localStorage.clear();
+          (win as any).ethereum = MockWeb3Provider.createDefault({
+            chainId: options?.chainId || 1,
+            privateKey: options?.privateKey || DEFAULT_ACCOUNT.privateKey,
+            address: options?.address || DEFAULT_ACCOUNT.address,
+          });
+        }
+      },
+    });
+  }
+);
+
+/**
+ * Custom command to select DOM element by data-cy attribute.
+ * @example cy.dataCy('greeting')
+ */
+Cypress.Commands.add("dataCy", (value) => {
+  cy.get(`[data-cy=${value}]`);
+});
+
+/**
+ * Inject mocked web3 provider to `window.ethereum`.
+ */
+Cypress.Commands.add("injectMockProvider", (options: MockProviderOptions) => {
+  cy.on("window:before:load", (win) => {
+    win.localStorage.clear();
+    (win as any).ethereum = MockWeb3Provider.createDefault(options);
+  });
+});
+
+/**
+ * Connect detected wallet via bnc-onboard.
+ */
+Cypress.Commands.add("connectInjectedWallet", (connectWalletElementDataId) => {
+  cy.dataCy(connectWalletElementDataId).click();
+  cy.contains("Show More").click();
+  cy.contains("Detected Wallet").click();
+});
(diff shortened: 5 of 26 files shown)


cypress/utils/ethereum.ts
```diff
@@ -0,0 +1,45 @@
...
+export type MockProviderOptions = {
+  chainId?: number;
+  address: string;
+  privateKey: string;
+};
+
+export const DEFAULT_ACCOUNT = {
+  address: "0xB98bD7C7f656290071E52D1aA617D9cB4467Fd6D",
+  privateKey:
+    "de926db3012af759b4f24b5a51ef6afa397f04670f634aa4f48d4480417007f3",
```
daywiss, line 12: hope this isnt real, but perhaps its possible we use an .env file to provide this and inject it as a config

cypress/utils/ethereum.ts
```diff
@@ -0,0 +1,45 @@
+import { MockProvider } from "@rsksmart/mock-web3-provider";
```
daywiss, line 1: never heard of this, you have experience with it?

cypress.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: "http://localhost:3000",
```
daywiss, line 5: this should be a default and overridden by env

cypress/utils/ethereum.ts
```diff
@@ -0,0 +1,45 @@
...
+      ...overrides,
+    });
+  }
+
+  async request({ method, params }: any): Promise<any> {
+    if (method === "eth_getBalance") {
+      return "0x0";
+    }
+
+    return super.request({ method, params });
```
daywiss, line 43: can this pass through to a real provider? i think in this app we use our own providers for reading state and only metamask for getting user state and submitting txs, so its probably not a big deal, but just wondering

daywiss (commented):
this is really great, thanks for working on this.  I attempted to run in my dev environment and was able to get this running though some tests failed with timeout ( not sure if thats a blocker for this pr).  also is there further work needed for this to run in CI?

cypress/utils/ethereum.ts
dohaki: no worries, this is only a test account :D but yea injecting this as a config would make sense.

cypress/utils/ethereum.ts
dohaki: I used it in another project, but the implementation is pretty straightforward (nothing special). See https://github.com/rsksmart/mock-web3-provider/blob/main/src/index.ts. But if it is preferred, I can also use a custom implementation.

cypress/utils/ethereum.ts
dohaki: Yea, so in theory if we want to use a real provider we could adapt this one or implement a new one that we inject. When writing a test, we could then specify if we want to inject a mocked provider or "real" one. Similar to how Uniswap is doing it https://github.com/Uniswap/interface/blob/main/cypress/support/ethereum.ts#L23

dohaki:
> this is really great, thanks for working on this. I attempted to run in my dev environment and was able to get this running though some tests failed with timeout ( not sure if thats a blocker for this pr). also is there further work needed for this to run in CI?

Hmm, ideally they should succeed in your dev env 🤔 CI via GitHub Action will be introduced in an upcoming PR.

.gitignore
```diff
@@ -25,3 +25,7 @@ yarn-debug.log*
 yarn-error.log*
 
 .vercel
+
+# cypress
+/cypress/videos
```
nicholaspai, line 30: should we just ignore anything in `/cypress`?

nicholaspai: nevermind i see you place all tests in `/cypress`. Do tests have to be placed in this file for `cypress` to work? Should we place tests instead in a `test/` folder?

cypress/support/commands.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import {
+  MockWeb3Provider,
+  MockProviderOptions,
+  DEFAULT_ACCOUNT,
+} from "../utils/ethereum";
+
+/**
+ * Overwrite default `visit` command to allow injecting web3 provider to
+ * `window.ethereum`.
```
nicholaspai, line 11: @Tulun is this injection into `window.ethereum` how every web3 provider works? Metamask, walletconnect, coinbase wallet, etc.

cypress/utils/ethereum.ts
nicholaspai: Seems useful to me

nicholaspai (commented):
This is awesome! Really excited to get this initial test PR in

.gitignore
dohaki: They don't have to be in the `/cypress` folder but are by default. If you don't have a strong opinion on this, I would like keep them for now in there until CI is set up. Because I already have it working with this structure on a fork 😅

package.json
```diff
@@ -60,7 +60,9 @@
     "lint-check": "prettier ./src",
     "predeploy": "yarn sitemap",
     "sitemap": "babel-node src/sitemap-generator.js",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "cypress:open": "cypress open --browser chrome --e2e",
+    "cypress:run": "cypress run --browser chrome --e2e"
```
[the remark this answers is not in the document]
dohaki: yes good point will change

cypress/utils/ethereum.ts
daywiss: im good with it, just curious about it

cypress/e2e/wallet.cy.ts
```diff
@@ -0,0 +1,18 @@
...
+    cy.dataCy("acx-balance").should("be.visible");
+  });
+
+  it("display warning for unsupported network", () => {
+    cy.visit("/", { withInjectedMockProvider: true, chainId: 1234 });
+    cy.connectInjectedWallet("connect-wallet");
+
+    cy.dataCy("unsupported-network").should("be.visible");
+  });
+});
```
Tulun, line 18: This test seems to not work for some reason when I run it locally.

Seems to loop infinitely.
Screen Shot 2022-08-09 at 12 35 58 PM

cypress/e2e/transactions.cy.ts
```diff
@@ -0,0 +1,9 @@
+describe("transactions", () => {
+  beforeEach(() => {
+    cy.visit("/transactions", { withInjectedMockProvider: true });
+  });
+
+  it("render in initial state", () => {
+    cy.dataCy("connect-wallet").should("be.visible");
+  });
+});
```
Tulun, line 9: Similar issue with this test, seems to not end and run infinitely.
Screen Shot 2022-08-09 at 12 35 26 PM

cypress.config.ts
Tulun: +1.

cypress/support/commands.ts
Tulun: This is how MetaMask works, which would be our default test case anyway. I'm not 100% sure on the rest.

cypress/support/index.d.ts
```diff
@@ -0,0 +1,22 @@
+declare namespace Cypress {
+  interface ApplicationWindow {
+    ethereum: any;
```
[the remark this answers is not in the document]
Tulun: @james-a-morris it might be annoying to, and as this is largely for testing purposes, I don't think we should stress too much.

cypress/utils/ethereum.ts
Tulun: @daywiss @dohaki  I will say we should make sure we shouldn't use what ethersprojects isn't doing instead over RSK -- this was an extendable implementation I did for umaverse about a year ago, which I had to add some methods to make it EIP1559 compatible.

We might want to default to an *extendable wallet* simply so we can keep our tests up to date with cutting edge changes to EIP / wallets.

https://github.com/UMAprotocol/umaverse/blob/master/cypress/utils/CustomizedBridge.js

cypress/utils/ethereum.ts
Tulun: @daywiss like the one I did ;) https://github.com/UMAprotocol/umaverse/blob/master/cypress/utils/CustomizedBridge.js

package.json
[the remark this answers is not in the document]
Tulun: Also in theory this should be headless in production (helpful for dev though) -- @evaldofelipe probably doesn't care what browser this runs on in CI, and it's probably more of a headache to run it as such anyway.

Tulun (commented):
Great start.

I might add it might be cool to try to get as far as you can on a harder test case. Largely this is a smoke screen test -- it visits and renders and does a quick connect, but when I was working with cypress the most meat and bones interesting test cases were when I had to actual transaction submission, which should involve forking with hardhat.

cypress/utils/ethereum.ts
dohaki: I changed it so that this can be overridden via `CYPRESS_DEFAULT_TEST_PK`. Note, that the `CYPRESS_` prefix is necessary in order to be detectable by cypress. See https://docs.cypress.io/guides/guides/environment-variables.

cypress/utils/ethereum.ts
dohaki: Yea I agree. I changed it so that we use the provider implementations from ethers. See https://github.com/across-protocol/frontend-v2/pull/200/commits/a081fe2b89c8a40e7d60bf2dfcb4f114a31076b8

cypress.config.ts
dohaki: Cypress actually natively allows this. You have to set `CYPRESS_BASE_URL`, see https://docs.cypress.io/guides/guides/environment-variables#Overriding-Configuration. I added some documentation on this to the README and the config file.

cypress/utils/ethereum.ts
dohaki: I changed the approach to have the option to either use a `MockProvider` or for more meaningful test cases a `JsonRpcProvider`. See again my last commit https://github.com/across-protocol/frontend-v2/pull/200/commits/a081fe2b89c8a40e7d60bf2dfcb4f114a31076b8.

cypress/support/index.d.ts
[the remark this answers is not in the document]
dohaki: Yea I also think it is a bit hard to type. Would prefer to leave it like this for now.

cypress/e2e/wallet.cy.ts
dohaki: Can you try again with the latest changes?

cypress/e2e/transactions.cy.ts
dohaki: Can you try again with the latest changes? Also note, that cypress logs all HTTP requests even though the test might have succeeded/ended.

dohaki:
> I might add it might be cool to try to get as far as you can on a harder test case. Largely this is a smoke screen test -- it visits and renders and does a quick connect, but when I was working with cypress the most meat and bones interesting test cases were when I had to actual transaction submission, which should involve forking with hardhat.

@Tulun Yes, I agree that this PR covers the rudimentary "easy" test cases. I would prefer to implement the "real" e2e test cases in a separate task/PR though when the setup and CI are done. Especially as we would need to discuss how and what exactly we want to cover there.

cypress/utils/ethereum.ts
daywiss: great points @Tulun

daywiss (approved):
just ran this locally and tests pass, but i needed to add a dev dependency, make sure you do that before merging
`yarn add -D @ethersproject/experimental`

package.json
```diff
@@ -85,13 +87,15 @@
   },
   "devDependencies": {
     "@babel/plugin-transform-modules-commonjs": "^7.18.6",
+    "@ethersproject/experimental": "^5.6.3",
```
daywiss, line 90: thanks for pointing this out, all good, must have been an issue on my side

Tulun (approved):

PR #203, opened 2022-08-09
feat: ENSName resolves to an address and is used in place of the referrer.

Tulun:
## Needs QA

**Overview**
Create a hook that attempts to resolve the ENS name to an address and use that.

**Acceptance Criteria**

Example:
Use the generated kovan link and first send a transaction between any chains, then go to transfer tab.
You should see your active transfer, and the new speedup button.

1. Valid ENS addresses should not show an error and should be sent as a referral tx (converts the ENS name to the underlying address)
2. Incorrect ENS name (e.g: asdfasadfas.eth ) shows an Invalid ENS name error banner.
3. Incorrect address (e.g: 0x13241241234) show an incorrect address error.

### How To Report QA

**Reporting QA Issues**

```
**QA Issue**

<!--
-->

**Repro Steps**

<!--

Example:
4. go to pool tab before going to transaction tab.
5. see that my pending transaction is missing.
-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->


```

**Reporting QA Regression**

**Reporting QA Passed**

```
**QA Pass**

<!--

Example:

-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->

```

Signed-off-by: Tulun <jaykiraly@gmail.com>


src/hooks/useENSNameToAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns Promise<string>, resolves to the address of the ENS name if it exists
+ */
+async function lookupAddress(
+  referrer: string,
+  provider: ethers.providers.Provider
+) {
+  try {
+    const resolvedAddress = await provider.resolveName(referrer);
+    return resolvedAddress || "";
+  } catch (e) {
```
daywiss, line 31: i know we tend to do a lot of try catches, but consider removing this entire function and doing the catch in the `useEnsNameToAddress` hook.  for example:

```js
    if (provider && referrer) {
      provider.resolveName(referrer).then(setAddress).catch(err=>{
         // log error or whatever, then set address undefined
         console.warn("error resolving name", e);
         setAddress(undefined)
          //optionally setAddress(referrer) so you always have a valid address returned
      })
    }
```

src/hooks/useBridge.ts
```diff
@@ -37,8 +38,8 @@ type SendError =
 export function useBridge() {
   const config = getConfig();
   const referrer = useReferrer();
-
-  const { chainId, account, signer } = useConnection();
+  const { chainId, account, signer, provider } = useConnection();
+  const resolvedAddress = useENSNameToAddress(referrer, provider);
```
daywiss, line 42: you could call this hook `useENSNameOrAddress(referrer,provider)` and have it automatically return the address if the ens name is not found, saving you a little logic below, and also making the state inside the hook a bit more natural

amateima:
> If it doesn't resolve, should we show some kind of error?

Yeah, I think we should notify the user that the ENS name was not resolved

src/hooks/useBridge.ts
Tulun: @daywiss yeah that's a good call. I think we could add some catches to the ref or referral query if say.. the address is invalid (say a character was lost somehow), or the ENS address is not valid.

This should have some kind of visual component, I may just pitch to Tim we show a warning banner perhaps on the bridge tab?

daywiss: sounds fine to me

src/hooks/useENSNameToAddress.ts
Tulun: Adjusted.

Tulun:
@daywiss @amateima @nicholaspai
Screen Shot 2022-08-10 at 1 03 46 PM
Screen Shot 2022-08-10 at 1 03 34 PM

Rough idea for showing an error.

src/hooks/useENSNameToAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+      setReferrerError("");
+      if (!ethers.utils.isAddress(referrer)) {
+        provider
+          .resolveName(referrer)
+          .then((ra) => {
+            setAddress(ra || "");
+            if (ra) {
+              setReferrerError("");
+            } else {
+              setReferrerError("Invalid address or ENS name");
```
daywiss, line 21: this should specify "referral address"

src/hooks/useBridge.ts
```diff
@@ -105,7 +109,7 @@ export function useBridge() {
         isNative: selectedRoute.isNative,
         relayerFeePct: fees.relayerFee.pct,
         timestamp: await hubPool.getCurrentTime(),
-        referrer,
+        referrer: resolvedAddress || referrer,
```
daywiss, line 112: if the ens name is invalid we will end up setting the referrer to the ens string because resolevdAddress will be `""`.  I dont think thats the behavior we want?

src/hooks/useBridge.ts
```diff
@@ -37,8 +38,11 @@ type SendError =
 export function useBridge() {
   const config = getConfig();
   const referrer = useReferrer();
```
daywiss, line 40: im wondering if `useReferrer` should be responsible for calling the ens name resolution and basically returning, an address, empty string or an error

Tulun: Yeah that's reasonable, I'll adapt useReferrer to be more complicated.

src/hooks/useReferrer.ts
```diff
@@ -1,9 +1,45 @@
...
+      setReferrerError("");
+      if (!ethers.utils.isAddress(r)) {
+        provider
+          .resolveName(r)
+          .then((ra) => {
+            setAddress(ra || "");
+            if (ra) {
+              setReferrerError("");
+            } else {
+              setReferrerError("Invalid ENS name");
```
daywiss, line 25: make sure this error references that its the referral ens name that is invalid

src/hooks/useReferrer.ts
```diff
@@ -1,9 +1,45 @@
...
   const { referrer, ref: refParam } = useQueryParams();
+  // Default to referrer if query ref isn't provided.
+  let r = refParam || referrer;
+
+  const [address, setAddress] = useState<string>("");
+  const [referrerError, setReferrerError] = useState<string>("");
+  useEffect(() => {
+    if (provider && r) {
+      setReferrerError("");
+      if (!ethers.utils.isAddress(r)) {
```
daywiss, line 17: i feel like this should be followed by an else which sets the referrer address

dohaki: +1 AFAICT if `r` is indeed an address then the state variable `address` will still have the value `""`.

src/hooks/useReferrer.ts
```diff
@@ -1,9 +1,45 @@
+import { useState, useEffect } from "react";
+import { ethers } from "ethers";
 import { useQueryParams } from "./useQueryParams";
+import { useConnection } from "state/hooks";
 
 export default function useReferrer() {
+  const { provider } = useConnection();
   const { referrer, ref: refParam } = useQueryParams();
+  // Default to referrer if query ref isn't provided.
+  let r = refParam || referrer;
```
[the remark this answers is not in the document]
Tulun: Done!

src/hooks/useReferrer.ts
Tulun: Done!

src/hooks/useReferrer.ts
Tulun: Fair, updated.

Tulun:
> I enjoyed reviewing this PR!
> 
> Not sure if this is in scope ( or where to suggest this in general ), but the way that this `referrerError` is displayed on the UI gives me the impression that I cannot proceed with the bridge.
> 
> AFAICT this is not the case and the `<SuperHeader>{referrerError}</SuperHeader>` is only informing the user of the invalid Ref.

I think it's ok for now -- the intention is to move this to a Toast component notification, but I'm building this.

I think if the referral link is off the user should feel some pressure to fix, although it doesn't actually blocking sending.

daywiss (approved):

PR #204, opened 2022-08-10
Bump sdk-v2 version to 0.1.16

kevinuma:


daywiss (approved):

nicholaspai (approved):

nicholaspai:
@kevinuma I just pushed a commit that should make the serverless function use the new coingecko pro API as long as `REACT_APP_COINGECKO_PRO_API_KEY` is set in the vercel app, which I'll do now.

PR #205, opened 2022-08-10
feat(api): add coingecko route

nicholaspai:
Adds route that can cache coingecko pro API responses via Vercel edge caching. We'll be able to call this route internally from /limits and /suggested-fees to reduce coingecko pro API hits

### DIFF
--- a/api/coingecko.js
+++ b/api/coingecko.js
@@ -3,7 +3,7 @@ const ethers = require("ethers");
 const { InputError, isString, getTokenPrice } = require("./_utils");
 
 const handler = async (request, response) => {
-  console.log(`INFO(pools): Handling request to /coingecko ${request}`);
+  console.log(`INFO(pools): Handling request to /coingecko`, request);
 
   try {
     let { l1Token } = request.query;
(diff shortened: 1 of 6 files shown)


dohaki (commented):
In general LGTM 👍 Only minor remarks

api/coingecko.js
```diff
@@ -0,0 +1,40 @@
...
+    const price = await getTokenPrice(l1Token, 1);
+
+    // This tells our CDN the value is fresh for one second. If a request is repeated within the next second,
+    // the previously cached value is still fresh. The header x-vercel-cache present in the response will show the
+    // value HIT. If the request is repeated between 1 and 60 seconds later, the cached value will be stale but
+    // still render. In the background, a revalidation request will be made to populate the cache with a fresh value.
+    // x-vercel-cache will have the value STALE until the cache is refreshed.
+    response.setHeader(
+      "Cache-Control",
+      "s-maxage=1, stale-while-revalidate=59"
```
kevinuma, line 24: Arguably we don't need to keep the prices fresh every 1 second. I think 10 seconds or longer might be better as on average, we're only getting 1-2 RPS. This means that if it's set to 1 second TTL, it's almos 100% cache miss always. Furthermore the CDN is location based, that means that if requests hit different locations, cache miss rate would be even higher. I even suspect we can do up to 30 secs of cache TTL. Let's experiment with that?

api/limits.js
```diff
@@ -99,6 +99,8 @@ const handler = async (request, response) => {
       hubPool.interface.encodeFunctionData("pooledTokens", [l1Token]),
     ];
 
+    // TODO: Get token price first from /coingecko route, which we'll use as input to getRelayerFeeDetails
```
kevinuma, line 102: It's an interesting consideration but this introduces potential serial request. We really need that latency observability soon...

dohaki (approved):
🚀

kevinuma (approved):

PR #208, opened 2022-08-10
improve(api): Add GCP logging

kevinuma:


api/_utils.js
```diff
@@ -5,22 +5,64 @@ const {
...
+  VERCEL_ENV,
+} = process.env;
 const {
   relayerFeeCapitalCostConfig,
   disabledL1Tokens,
 } = require("./_constants");
 
+const log = (gcpLogger, severity, at, message, extraData) => {
+  gcpLogger.write(
+    this.log.entry(
```
nicholaspai, line 23: This is kind of confusing, what is `this.log`?

api/_utils.js
```diff
@@ -5,22 +5,64 @@ const {
...
+        severity: severity,
+      },
+      { message, ...extraData }
+    )
+  );
+};
+
+const getLogger = () => {
+  const gcpLogger = new Logging({
+    projectId: JSON.parse(REACT_APP_GOOGLE_SERVICE_ACCOUNT).project_id,
```
nicholaspai, line 38: Since we're in the API and we care about speed, i'm going to be nitty about this: JSON.parse the env once and then destructure

api/_utils.js
```diff
@@ -5,22 +5,64 @@ const {
...
+
 const getTokenDetails = async (provider, l1Token, l2Token, chainId) => {
   const hubPool = HubPool__factory.connect(
     "0xc186fA914353c44b2E33eBE05f21846F1048bEda",
     provider
   );
-  console.log(
-    `INFO: Fetching token details for ${l1Token} ${l2Token} on chain ${chainId}`
-  );
+  logger.info("getTokenDetails", "Fetching token details", {
```
nicholaspai, line 61: Once this deploys to staging you should test out the latency when logging to GCP

api/_utils.js
```diff
@@ -5,22 +5,64 @@ const {
...
+    info: (at, message, extraData) =>
+      log(gcpLogger, "INFO", at, message, extraData),
+    warn: (at, message, extraData) =>
+      log(gcpLogger, "WARN", at, message, extraData),
+    error: (at, message, extraData) =>
+      log(gcpLogger, "ERROR", at, message, extraData),
+  };
+};
+// Singleton logger so we don't create multiples.
+const logger = getLogger();
```
nicholaspai, line 54: If the `REACT_APP_GOOGLE_SERVICE_ACCOUNT` isn't set can you just log to console by default? Would be helpful to debug

api/coingecko.js
```diff
@@ -35,12 +33,12 @@ const handler = async (request, response) => {
     );
     response.status(200).json({ price });
   } catch (error) {
-    console.log(`ERROR(coingecko): Error found: ${error}`);
-
     let status;
     if (error instanceof InputError) {
+      logger.warn("logger", "400 input error", { error });
```
nicholaspai, line 38: This should be `coingecko` not `logger` right? Same below

nicholaspai (commented):
Can we make all `error` level logs `warn` or make sure they don't alert PD yet until we test in prod for a bit?

api/_utils.js
kevinuma: Fixed.

api/_utils.js
kevinuma: Updated

api/_utils.js
kevinuma: Updated

kevinuma:
> Can we make all `error` level logs `warn` or make sure they don't alert PD yet until we test in prod for a bit?

Keeping warn for now as we can configure PD alerts on the GCP side. This way we wouldn't need to change the code every time

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
+/shelf/
```
nicholaspai, line 2: What is .idea?

api/_utils.js
```diff
@@ -3,25 +3,76 @@ const {
...
+    return sdk.relayFeeCalculator.DEFAULT_LOGGER;
+  }
+
+  const gcpLogger = new Logging({
+    projectId: GOOGLE_SERVICE_ACCOUNT.project_id,
+    credentials: {
+      client_email: GOOGLE_SERVICE_ACCOUNT.client_email,
+      private_key: GOOGLE_SERVICE_ACCOUNT.private_key,
+    },
+  }).log(VERCEL_ENV, { removeCircular: true });
```
nicholaspai, line 53: Should there be a default VERCEL_ENV value set in the code?

nicholaspai (commented):
Left one suggestion but I am also curious what .idea/ is? Should that have been gitignored?

api/coingecko.js
kevinuma: Updated

nicholaspai:
> Why is the `yarn.lock` diff so large? Is `@google-cloud/loggin` that large?

Looks like something in the `yarn.lock` is preventing Vercel from building correctly: https://vercel.com/uma/goerli-frontend-v2/Eg9CaG8aYUsh4tWkSgDDmSoyUrHP

api/_utils.js
kevinuma: Latency didn't seem to get worse. Also these logs are done as fire-and-forgets so the endpoint handler should not wait for them to finish.

api/_utils.js
```diff
@@ -75,7 +145,11 @@ class InputError extends Error {}
 
 const infuraProvider = (name) => {
   const url = `https://${name}.infura.io/v3/${REACT_APP_PUBLIC_INFURA_ID}`;
-  console.log(`INFO: Using infura provider at ${url}`);
+  getLogger().info({
+    at: "infuraProvider",
+    message: "Using an Infura provider",
```
nicholaspai, line 150: You should pretty much make most of these debug level logs, info might log to slack channels which we dont want. That's how it works in other bots

api/limits.js
```diff
@@ -166,9 +175,11 @@ const handler = async (request, response) => {
       liquidReserves = liquidReserves.sub(
         ethers.utils.parseEther(REACT_APP_WETH_LP_CUSHION || "0")
       );
-      console.log(
-        `INFO(limits): Adding WETH cushioning to LP liquidity: ${liquidReserves}`
-      );
+      logger.info({
```
nicholaspai, line 178: same comment: https://github.com/across-protocol/frontend-v2/pull/208/files#r944850480

nicholaspai (commented):
This almost looks ready to go. Main comment is remove .info's that you don't want to log to a slack channel

api/limits.js
kevinuma: Updated

api/_utils.js
kevinuma: Updated

PR #209, opened 2022-08-11
ci: initial lint + cypress test run github actions

dohaki:
This PR adds initial GitHub Action workflows for linting and running cypress tests. Someone who is an admin of this repository needs to add the following secrets before the run succeeds:
- `CYPRESS_PUBLIC_INFURA_ID`
- `CYPRESS_PUBLIC_ONBOARD_API_KEY`
- `CYPRESS_REWARDS_API_URL`
- `CYPRESS_CHAIN_137_PROVIDER_URL`
- `CYPRESS_CHAIN_42161_PROVIDER_URL`

You can have a look at my forked repository where I tested this https://github.com/dohaki/across-frontend-v2/pull/1 to see a successful run.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -1,0 +1,58 @@
+name: Lint and Test
+
+on:
+  # manual trigger
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - run: yarn lint-check
+
+  cypress-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - uses: actions/cache@v3
+        id: cypress-cache
+        with:
+          path: /home/runner/.cache/Cypress
+          key: ${{ runner.os }}-cypress-${{ hashFiles('node_modules/cypress') }}
+      - if: steps.cypress-cache.outputs.cache-hit != 'true'
+        run: yarn cypress install
+      - name: Cypress run Chrome
+        uses: cypress-io/github-action@v4
+        with:
+          install: false
+          browser: chrome
+          start: yarn start
+          wait-on: "http://localhost:3000"
+          wait-on-timeout: 300
+        env:
+          REACT_APP_PUBLIC_INFURA_ID: ${{ secrets.CYPRESS_PUBLIC_INFURA_ID }}
+          REACT_APP_PUBLIC_ONBOARD_API_KEY: ${{ secrets.CYPRESS_PUBLIC_ONBOARD_API_KEY }}
+          REACT_APP_REWARDS_API_URL: ${{ secrets.CYPRESS_REWARDS_API_URL }}
+          REACT_APP_CHAIN_137_PROVIDER_URL: ${{ secrets.CYPRESS_CHAIN_137_PROVIDER_URL }}
+          REACT_APP_CHAIN_42161_PROVIDER_URL: ${{ secrets.CYPRESS_CHAIN_42161_PROVIDER_URL }}
+      - name: Upload screenshots
+        uses: actions/upload-artifact@v2
+        if: failure() # only upload screenshots if tests fail
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+      - name: Upload videos
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: cypress-videos
+          path: cypress/videos
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -30,20 +30,22 @@ jobs:
           key: ${{ runner.os }}-cypress-${{ hashFiles('node_modules/cypress') }}
       - if: steps.cypress-cache.outputs.cache-hit != 'true'
         run: yarn cypress install
-      - name: Cypress run Chrome
-        uses: cypress-io/github-action@v4
-        with:
-          install: false
-          browser: chrome
-          start: yarn start
-          wait-on: "http://localhost:3000"
-          wait-on-timeout: 300
+      - name: Build app
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,58 @@
...
+          browser: chrome
+          start: yarn start
+          wait-on: "http://localhost:3000"
+          wait-on-timeout: 300
+        env:
+          REACT_APP_PUBLIC_INFURA_ID: ${{ secrets.CYPRESS_PUBLIC_INFURA_ID }}
+          REACT_APP_PUBLIC_ONBOARD_API_KEY: ${{ secrets.CYPRESS_PUBLIC_ONBOARD_API_KEY }}
+          REACT_APP_REWARDS_API_URL: ${{ secrets.CYPRESS_REWARDS_API_URL }}
+          REACT_APP_CHAIN_137_PROVIDER_URL: ${{ secrets.CYPRESS_CHAIN_137_PROVIDER_URL }}
+          REACT_APP_CHAIN_42161_PROVIDER_URL: ${{ secrets.CYPRESS_CHAIN_42161_PROVIDER_URL }}
```
dohaki, line 46: I think it would make sense to create a new project on Infura and use them instead of our production ones. So that rate limits and analytics can be tracked separately.

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,58 @@
...
+          path: /home/runner/.cache/Cypress
+          key: ${{ runner.os }}-cypress-${{ hashFiles('node_modules/cypress') }}
+      - if: steps.cypress-cache.outputs.cache-hit != 'true'
+        run: yarn cypress install
+      - name: Cypress run Chrome
+        uses: cypress-io/github-action@v4
+        with:
+          install: false
+          browser: chrome
+          start: yarn start
```
dohaki, line 38: Ideally, we should create a production build and serve that one to run the tests against. But a production build took around ~ 7 mins via GHA, whereas starting a dev server took something around 4 mins. So maybe we should first investigate why these builds take so long and eventually change this one.

nicholaspai: Thats strange why the build takes so long, do you think thats much longer than expected?

.github/workflows/lint-test.yaml
nicholaspai: @mrice32 wdyt?

nicholaspai (commented):
Do you plan to add `eslint` in this PR or a followup?

dohaki:
> Do you plan to add `eslint` in this PR or a followup?

Yes will do so in a separate PR because there might be some linting errors/fixes that I did not want to include here.

nicholaspai:
Sounds good!

.github/workflows/lint-test.yaml
dohaki: My guess would be that we have some dependencies that are pretty big and maybe can not be optimized / tree-shaken easily. 

Updating our build tool from create-react-app@v4/webpack@v4 to v5 or changing to another tool/bundler, e.g. vite, esbuild, rollup, etc. could also speed this up. Note, that updating or migrating doesn't always go smoothly as someone might think though.

nicholaspai: Yeah changing build tool will be messy, are there any tools that are worth upgrading to that are pretty easy to use?

nicholaspai (approved):
Looks like your new `lint-and-test` workflow is failing in CI actions, otherwise LGTM

.github/workflows/lint-test.yaml
dohaki: I found this one https://github.com/pradel/create-react-app-esbuild/tree/main/packages/craco-esbuild which should in theory just be plug and play. Maybe worth to try out at some point and revisiting this.

.github/workflows/lint-test.yaml
mrice32: Sounds good!

.github/workflows/lint-test.yaml
mrice32: A few questions:
- Do you think a production build is still worth doing _even if_ it takes longer, since it's possible for these to have different behavior?
- Is it possible that the build takes a long time _because it is doing a lot of tree-shaking_ and other optimizations? Regardless, a bundle size analysis is probably a good idea to see if there are any issues with the bundle size!

.github/workflows/lint-test.yaml
mrice32: We are out of infura projects, but I repurposed one that doesn't seem to be used anymore and renamed it for this purpose: `cypress-ci`

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,58 @@
...
+        if: failure() # only upload screenshots if tests fail
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+      - name: Upload videos
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: cypress-videos
+          path: cypress/videos
```
mrice32, line 58: Just OOC, why do we want to always upload videos, but only upload screenshots on failure?

mrice32 (approved):
LGTM!

.github/workflows/lint-test.yaml
dohaki: > Do you think a production build is still worth doing even if it takes longer, since it's possible for these to have different behavior?

Yea, now that I think about it, it is better to run the tests against a production build, even though it takes a bit longer. I will change that.

> Is it possible that the build takes a long time because it is doing a lot of tree-shaking and other optimizations? Regardless, a bundle size analysis is probably a good idea to see if there are any issues with the bundle size!

Yes, I think there are some dependencies that are either not tree-shakable or we import them in a sub-optimal way. If you look at the build output in the console, you can see some chunks with a size of > 2MB (after gzip). That is quite large and can be optimized I believe. I will open an issue to track this.

.github/workflows/lint-test.yaml
dohaki: It is the default behavior of [Cypress](https://docs.cypress.io/guides/guides/screenshots-and-videos#Screenshots). My guess would be that a screenshot of a failed test gives you a quick preview of what went wrong without the need to watch the whole video.

PR #213, opened 2022-08-12
feat: create toast widget + hook up referrer notification to new Toast.

Tulun:
## Needs QA

**Overview**
Add new toast element to across UI.

Screen Shot 2022-08-12 at 4 12 06 PM
Screen Shot 2022-08-12 at 4 11 51 PM

**Acceptance Criteria**
1. Toast should display when you put in a bad query string (e.g: ?ref=0x123432142 or referrer=0x123412314) of a bad address.
2. Toast should display an when you put in an ENS value that doesn't resolve (e.g: ?ref=nonsensestring.eth  or ?referrer=nonsensestring.eth)

### How To Report QA

**Reporting QA Issues**

```
**QA Issue**

<!--
-->

**Repro Steps**

<!--

Example:
4. go to pool tab before going to transaction tab.
5. see that my pending transaction is missing.
-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->


```

**Reporting QA Regression**

**Reporting QA Passed**

```
**QA Pass**

<!--

Example:

-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->

```


Penny-Panda:
**QA Report**

Found AssertionError for goerli-frontend-v2, dapp doesn't load up.
And, Toast doesn't show up when adding false/incorrect/invalid ENS names (as shown in the pictures), when wallet is not connected. 
It shows up when attempting to connect wallet, and works fine when wallet is connected.
It doesn't work for DNSSEC ENS domains though, such as "truts.xyz"


**Screenshots**

frontend-v2:
image
image
image
image
image
image




goerli-frontend-v2:
image

**Testing Environment**
Google Chrome
Version 104.0.5112.81 (Official Build) (64-bit)
Desktop

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [cid, setCid] = useState(0);
+  const addToast = useCallback(
+    (item: PartialToast) => {
+      setList([...list, { ...item, id: cid }]);
+      setCid((pv) => pv + 1);
+    },
+    [cid, list]
+  );
+
+  console.log("list", list);
```
daywiss, line 30: stray

src/hooks/useReferrer.ts
```diff
@@ -2,46 +2,60 @@ import { useState, useEffect } from "react";
...
           })
           .catch((e) => {
             // Error here would imply an issue with the provider call, not with the ENS name necessarily.
             console.warn("error resolving name", e);
             setAddress("");
             if (!ethers.utils.isAddress(r)) {
-              setReferrerError("Invalid referral address");
-            } else {
-              setReferrerError("");
+              addToast({
```
daywiss, line 40: i dont think the userReferrer hook should depend on the toast component, can you find another way to trigger it, perhaps in a higher level component

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function useToastManager() {
+  const [list, setList] = useState<ToastProperties[]>([]);
+  // Current id for toast
+  const [cid, setCid] = useState(0);
+  const addToast = useCallback(
+    (item: PartialToast) => {
+      setList([...list, { ...item, id: cid }]);
+      setCid((pv) => pv + 1);
```
daywiss, line 25: im not sure i understand the reason for this, do you just need a unique id for react or something?

src/assets/icons/cross.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M13.1971 3.69677C13.49 3.40388 13.49 2.92901 13.1971 2.63611C12.9042 2.34322 12.4293 2.34322 12.1364 2.63611L7.91653 6.85599L3.69665 2.63611C3.40376 2.34322 2.92888 2.34322 2.63599 2.63611C2.3431 2.92901 2.3431 3.40388 2.63599 3.69677L6.85587 7.91665L2.63599 12.1365C2.3431 12.4294 2.3431 12.9043 2.63599 13.1972C2.92888 13.4901 3.40376 13.4901 3.69665 13.1972L7.91653 8.97731L12.1364 13.1972C12.4293 13.4901 12.9042 13.4901 13.1971 13.1972C13.49 12.9043 13.49 12.4294 13.1971 12.1365L8.97719 7.91665L13.1971 3.69677Z" fill="#9DAAB2"/>
+</svg>
```
amateima, line 3: By looking at the Figma design, this icon should come in 2 sizes: 16px and 20px

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { ReactComponent as CloseIcon } from "assets/icons/cross.svg";
+import { ReactComponent as UnstyledInfoIcon } from "assets/icons/info-24.svg";
+
+interface IWrapper {
+  position: ToastPosition;
+}
+export const ToastContainer = styled.div<IWrapper>`
+  font-size: 14px;
+  position: fixed;
+  z-index: 999999;
```
amateima, line 13: Please add the used z-index to the contributing.md file where all z-indexes are mentioned

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+  width: 400px;
+  background-color: transparent;
+  box-shadow: 0px 12px 24px rgba(0, 0, 0, 0.24);
```
amateima, line 16: why the container needs to have a box shadow if the background is transparent? My understanding is that the `ToastContainer` can contain multiple toasts messages, the box-shadow being needed for each toast message, not the whole container

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+  border-radius: 16px;
```
amateima, line 17: is the border radius needed for the container itself?

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+      return "transform 0.6s ease-in-out";
+    } else {
+      return "transform 0.6s ease-in";
+    }
+  }};
+  animation: ${({ position }) => {
+    if (position === "top-right" || position === "bottom-right") {
+      return css`
+        ${toastInRight} 0.7s
+      `;
```
amateima, line 44: not sure I agree with this behaviour. In my view, the container should always be displayed on the screen having the clicks disabled and the visual effects should be applied to the toast messages directly

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+`;
+
+interface ITitle {
+  type: ToastType;
+}
+export const Title = styled.div<ITitle>`
+  font-weight: 700;
+  font-size: 16px;
+  text-align: left;
+  width: 300px;
```
amateima, line 107: font-weight and line-height are not as in the design. Also mobile breakpoint styles are missing

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+      `;
+    } else {
+      return css`
+        ${toastInLeft} 0.7s
+      `;
+    }
+  }};
+  @media screen and (max-width: 428px) {
+    width: 95%;
+    margin-left: 12px;
```
amateima, line 53: I don't see why the width should be set to 95%. Seems like a random value

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const Toast: React.FC<ToastProps> = ({ position }) => {
+  const { toastList, deleteToast } = useToast();
+
+  return (
+    <>
+      <ToastContainer position={position}>
+        {toastList.map(({ type, title, body, id }, i) => (
+          <ToastWrapper key={i} type={type}>
+            <ToastElement>
```
amateima, line 29: Can't we have a single ToastWrapper component? Why a second ToastElement component is necessary ?

src/hooks/useReferrer.ts
amateima: I agree. useToast should be pulled out and used in the same component where useReferrer is called

src/components/Toast/Toast.styles.tsx
```diff
@@ -0,0 +1,188 @@
...
+  @media screen and (max-width: 428px) {
+    width: 100%;
+  }
+`;
+
+export const Body = styled.div`
+  margin: 0;
+  text-align: left;
+  white-space: nowrap;
+  color: #c5d5e0;
```
amateima, line 129: line-height is not as in the design. Also mobile breakpoint styles are missing

src/components/Toast/Toast.styles.tsx
amateima: the padding should be changed here as well

src/components/Toast/useToast.tsx
amateima: I'm not sure why we need it too

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,65 @@
...
+  deleteToast: (id: number) => void;
+}
+
+interface PartialToast {
+  type: ToastType;
+  title: string;
+  body: string;
+}
+
+function useToastManager() {
```
amateima, line 18: Based on a previous discussion I had with Tim, he wants the toast message to behave very similar to the toasts from Linear app. You can see an example if you go to any issue and click the branch name button

Screenshot 2022-08-15 at 16 29 08

You'll see that we need to provide the option for the toasts to auto-hide after some time. Also, if you click on the same button multiple times, you'd still have a single toast being displayed alongside a discrete scaling effect. In order to achieve this behaviour, I think an optional id string should be provided.

amateima (changes requested):

src/hooks/useReferrer.ts
Tulun: Sure, it can be composed that way.

src/components/Toast/useToast.tsx
Tulun: The reason here is because right now you can do an array of Toast elements rendering.

When you delete them, you can delete them in any order, so we need to make sure the ID is unique (as you can have a stack of like 1, 3, 7, 8 , 10 notifications for example) so you don't delete two notifications when you click "X"

src/components/Toast/Toast.styles.tsx
Tulun: Yep, good call.

src/components/Toast/Toast.styles.tsx
Tulun: Right I'll move it down.

src/components/Toast/Toast.tsx
Tulun: because there's the container (which contains all Toasts), and the Toasts themselves, which can have CSS changes based on type (IE: warning vs error). The border around the toast changes.

src/components/Toast/useToast.tsx
Tulun: I see. This was definitely not documented anywhere, but this toast can be redesigned to just show 1 notification at a time.

My question would be is what would happen if you get 2 pertinent warnings at once? As I think the Linear way just overwrites the toast.

Tulun: Oh I see the effect. I dunno this is weird. You'd have to know the bug is from the same error, but what if you make a 2nd request and get a similar response? It might be hard to know if the error is generated from the same *type* of problem.

Tulun:
> **QA Report**
> 
> Found AssertionError for goerli-frontend-v2, dapp doesn't load up. And, Toast doesn't show up when adding false/incorrect/invalid ENS names (as shown in the pictures), when wallet is not connected. It shows up when attempting to connect wallet, and works fine when wallet is connected. It doesn't work for DNSSEC ENS domains though, such as "truts.xyz"
> 
> **Screenshots**
> 
> frontend-v2: image image image image image image
> 
> goerli-frontend-v2: image
> 
> **Testing Environment** Google Chrome Version 104.0.5112.81 (Official Build) (64-bit) Desktop

Thanks for the QA!

So we wait for the provider to be defined so the fact it only works when connected is fine -- in fact since you can't make a transaction until you are connected, it seems fine to wait until the user has clicked through before peppering them with a warning / error.

I'm not as familiar with the .xyz extension stuff, is this supported out of the box with ethers? Or how is this normally resolved?

src/components/Toast/useToast.tsx
Tulun: fixed

src/components/Toast/useToast.tsx
Tulun: Tagged y'all on a topic in linear.

src/hooks/useReferrer.ts
Tulun: Adjusted

src/components/Toast/Toast.styles.tsx
Tulun: Removed.

src/components/Toast/Toast.styles.tsx
Tulun: What did you have in mind instead?

src/components/Toast/Toast.styles.tsx
Tulun: Adjusted.

src/hooks/useReferrer.ts
```diff
@@ -2,46 +2,60 @@ import { useState, useEffect } from "react";
...
-  const [referrerError, setReferrerError] = useState<string>("");
   useEffect(() => {
     if (provider && r) {
-      setReferrerError("");
-      if (!ethers.utils.isAddress(r)) {
+      if (ethers.utils.isAddress(r)) {
+        return setAddress(r);
+      }
+
+      if (r.slice(-4) === ".eth") {
```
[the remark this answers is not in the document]
Tulun: I think we aren't too worried about general users testing referrer links on testnets :)

src/components/Toast/useToast.tsx
```diff
@@ -0,0 +1,124 @@
...
+interface PartialToast {
+  type: ToastType;
+  title: string;
+  body: string;
+  createdAt: number;
+  id: string;
+}
+
+function useToastManager() {
+  const [list, setList] = useState<ToastProperties[]>([
```
daywiss, line 21: this default state while good for testing will need to be changed before merge to master

src/hooks/useReferrer.ts
```diff
@@ -6,42 +6,44 @@ import { useConnection } from "state/hooks";
...
   useEffect(() => {
+    setError("");
     if (provider && r) {
-      setReferrerError("");
-      if (!ethers.utils.isAddress(r)) {
+      if (ethers.utils.isAddress(r)) {
+        return setAddress(r);
+      }
+
+      if (r.slice(-4) === ".eth") {
```
daywiss, line 22: i feel like this check isnt necessary since it will fail when `resolveName` fails

daywiss (approved):
looks good to me, just minor nits

src/assets/icons/info-sm.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M7.99998 14.6667C11.6819 14.6667 14.6666 11.6819 14.6666 8C14.6666 4.3181 11.6819 1.33334 7.99998 1.33334C4.31808 1.33334 1.33331 4.3181 1.33331 8C1.33331 11.6819 4.31808 14.6667 7.99998 14.6667Z" stroke="#9DAAB2" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M8 10.6667V8" stroke="#9DAAB2" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+<path d="M8 5.33363H8.00667" stroke="#9DAAB2" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+</svg>
```
amateima, line 5: small nit, don't we already have an [info-16](https://github.com/across-protocol/frontend-v2/blob/8a4dd9fe36e55827a77a6e98f1e86833d3cdad73/src/assets/icons/info-16.svg) icon?

src/hooks/useReferrer.ts
Tulun: Removed.

src/assets/icons/info-sm.svg
Tulun: Wasn't sure this was exactly the same, adjusted.

PR #216, opened 2022-08-16
ci: add eslint to github actions

dohaki:
Eslint is already included in CRA, but we didn't really enforce it.


amateima (approved):
It looks good. ESLint & Prettier warnings weren't displayed in VSCode for me. This may fix the issue on my end

PR #220, opened 2022-08-17
refactor: transactions view components

dohaki:
## Needs QA

**Overview**

While working on the “All Transactions” view, I had some problems understanding and reusing the existing code. This PR, therefore, tries to improve that by implementing some best practices
- replace `createTransactionTableJSX` and `createMobileTransactionTableJSX` with proper JSX components
  - it was very hard to understand how the `TransactionsTable` should be used or how the rows look like
  - we now use rows and cells components that in a more declarative way describe the ui
- localize ui state where possible
  - we had a lot of state/handlers in the top-level component/hook, even though they could be moved down the hierarchy
- extract hooks and components into separate functions
- improve separation of concerns and abstractions
- improve folder/file structure

FYI, draft PR for all transactions view that is based on this branch https://github.com/across-protocol/frontend-v2/pull/221.

**Acceptance Criteria**

Use the Vercel preview link and visit the `/transactions` page.

1. See and verify that everything works and looks as before on desktop.
2. See and verify that everything works and looks as before on mobile.

### How To Report QA

**Reporting QA Issues**

```
**QA Issue**

<!--
-->

**Repro Steps**

<!--

Example:
2. go to pool tab before going to transaction tab.
3. see that my pending transaction is missing.
-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->


```

**Reporting QA Regression**

**Reporting QA Passed**

```
**QA Pass**

<!--

Example:

-->

**Testing Environment**

<!--
test differently from yours.

Example:
-->

```


src/views/Transactions/hooks/useTxClient.ts
```diff
@@ -0,0 +1,108 @@
...
+          );
+          setFilledTransferTuples((prevFilledTransfers) =>
+            getUpdatedTransferTuples(prevFilledTransfers, nextFilledTransfers)
+          );
+          setPendingTransferTuples((prevPendingTransfers) =>
+            getUpdatedTransferTuples(prevPendingTransfers, nextPendingTransfers)
+          );
+        }
+      );
+    }
```
daywiss, line 47: this need a unsubscribe cleanup?

src/views/Transactions/hooks/useTxClient.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (account && txClient) {
+        txClient.stopFetchingTransfers(account);
+        setFilledTransferTuples([]);
+        setPendingTransferTuples([]);
+      }
+    };
+  }, [account, txClient]);
+
+  useEffect(() => {
+    return () => {
```
daywiss, line 77: not sure why this isnt just included in the previous use effect where the timer is set

daywiss (commented):
i think this is an improvement over prev implementation as code is smaller and more composable, but we just need to be careful of sweeping refactors and make sure we have a good justification.  this is also fairly high risk as it could break a major feature on the app, so we should be doing rigorous QA.  in my testing i didnt see any issues though, so looks good to me, just a couple nits

src/views/Transactions/hooks/useTxClient.ts
dohaki: The `TransfersHistoryClient` of the SDK doesn't seem to provide methods to remove event listeners / unsubscribe AFAICT 🤔 There is only the `stopFetchingTransfers` method which we call in the clean-up effect below.

src/views/Transactions/hooks/useTxClient.ts
dohaki: Most of the code was copy&pasted so not 100% sure. But yea I could move this clean-up to the previous hook

Tulun (approved):
Seems good to me.

Echoing David's comment this is a bit of a departure from our previous structure but I'm more inclined for this kind of folder structure.

We might want to be very cautious doing this in other parts of the app though. TX page is a read only thing, if this had been done on Pool or Bridge might be a lot of pushback. A refactor may not be crazy but I'd check in first, and we'd need to do some major QA.

daywiss (approved):

amateima (approved):

PR #222, opened 2022-08-17
improve(coingecko): Increase cache duration to 5 mins

nicholaspai:
Our goal is to reduce cache misses as they result in very slow response times for callers because of the lambda function instantiation


mrice32 (approved):
LGTM!

PR #223, opened 2022-08-17
feat: basic route + scaffolding of claim page

Tulun:
1. Create basic view for the Claim page
2. Connect to router.
3. Fix the header to properly highlight the route in the header.

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -53,10 +53,9 @@ const Header: React.FC<Props> = ({ openSidebar, setOpenSidebar }) => {
         <List>
           {LINKS.map(({ href, name }) => {
             const subRoute = parentRoutes.includes(href);
-            let ariaSelected = location.pathname === href;
-
-            if (subRoute && location.pathname.includes(href))
-              ariaSelected = true;
+            const ariaSelected =
+              location.pathname === href ||
+              (subRoute && location.pathname.includes(href));
             return (
               <Item key={href} aria-selected={ariaSelected}>
                 <Link
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -29,6 +29,9 @@ interface Props {
   openSidebar: boolean;
   setOpenSidebar: React.Dispatch<React.SetStateAction<boolean>>;
 }
+
+const parentRoutes = ["/rewards"];
+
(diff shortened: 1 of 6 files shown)


src/components/Header/Header.tsx
```diff
@@ -48,7 +48,16 @@ const Header: React.FC<Props> = ({ openSidebar, setOpenSidebar }) => {
       <Navigation>
         <List>
           {LINKS.map(({ href, name }) => (
-            <Item key={href} aria-selected={location.pathname === href}>
+            <Item
+              key={href}
+              aria-selected={
+                location.pathname.includes("/rewards") && href === "/rewards"
```
nicholaspai, line 54: Why is this change necessary?

nicholaspai (commented):
One question to understand better

src/components/Header/Header.tsx
Tulun: you want the rewards tab to have that "bubble" underneath it even for sub-routes.

Hence you gotta look (atm) if it includes "/rewards". We may need to make this more robust as we add more routes.

nicholaspai: Can't this be condensed into one || statement? I feel like double ? Operator is less readable

Tulun: I actually kind of doubt it. The actual correct way of doing this is to probably create an array of parent routes ["/rewards", "/example2"] and check to see if the href includes it.

Let me attempt that anyway.

src/components/Header/Header.tsx
```diff
@@ -47,18 +51,25 @@ const Header: React.FC<Props> = ({ openSidebar, setOpenSidebar }) => {
...
-                  search: location.search,
-                }}
-              >
-                {name}
-              </Link>
-            </Item>
-          ))}
+          {LINKS.map(({ href, name }) => {
+            const subRoute = parentRoutes.includes(href);
+            let ariaSelected = location.pathname === href;
```
nicholaspai, line 56: nit suggestion, does this work?
```suggestion
            const ariaSelected = location.pathname === href || (subRoute && location.pathname.includes(href));
```

Tulun: yep that does, cool.

nicholaspai (approved):
LGTM

PR #233, opened 2022-08-23
feat(claim): view-only components eligible claim

dohaki:
This adds (mostly) view-only components for an eligible claim. Ignore the stubbed hooks for now.


amateima (approved):

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Loader as LoaderIcon } from "react-feather";
+
+type Props = {
+  size?: number;
+};
+
+export const Loader = styled(LoaderIcon)`
+  width: ${(props: Props) => props.size || 24}px;
+  height: ${(props: Props) => props.size || 24}px;
+  animation: rotation 2s infinite linear;
```
Tulun, line 11: You can pass in the Props as an arg like styled(LoaderIcon)<Props>``

I'd also deconstruct, makes the code cleaner.

src/views/Claim/hooks/useClaimView.tsx
```diff
@@ -1,51 +1,40 @@
...
-  const eligibleState = useIsEligible();
+  const { isConnected, provider } = useConnection();
+
+  const isEligibleQuery = useIsEligible();
+  const claimableTokensQuery = useClaimableTokens();
+  const { handleClaim, claimState, hasClaimed } = useMerkleDistributor();
+
+  const handleAddTokenToWallet = async () => {
+    if (provider) {
+      await (provider as any).send("wallet_watchAsset", {
```
Tulun, line 18: Bit odd here. provider has a clear typing in ethers, why are you using any?

src/views/Claim/hooks/useClaimableTokens.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { useConnection } from "state/hooks";
+
+export function useClaimableTokens() {
+  const { isConnected, account } = useConnection();
+
+  return useQuery(
+    ["claimable", account],
+    () => getClaimableTokens(account as string),
+    {
```
Tulun, line 12: isn't the account a string type? should we just not run the query if there's no account?

src/views/Claim/components/EligibleWallet.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ReactComponent as PieChartWave } from "assets/claim-pie-chart-wave.svg";
+
+import { StepCard } from "./StepCard";
+import { ClaimAirdrop, Props as ClaimAirdropProps } from "./ClaimAirdrop";
+import { WaysToEarn } from "./WaysToEarn";
+
+type Props = ClaimAirdropProps;
+
+export function EligibleWallet(props: Props) {
+  const [expandedStepIndex, setExpandedStepIndex] = useState(0);
```
Tulun, line 13: I'd probably deconstruct the props to clean this up a bit.

src/views/Claim/components/ClaimAirdrop.tsx
```diff
@@ -0,0 +1,194 @@
...
+    communityClaim: BigNumberish;
+    totalClaim: BigNumberish;
+  };
+  onClickClaim: () => void;
+  onClickAddToken: () => void;
+};
+
+export function ClaimAirdrop(props: Props) {
+  return (
+    <Container>
```
Tulun, line 28: Let's deconstruct the props you're using.

Tulun (commented):
Mostly minor stuff, good job. Looks good.

If my comments on the hooks are a bit off that's ok I know that's not the focus here.

src/views/Claim/hooks/useClaimView.tsx
dohaki: Yea a bit weird here. The typing of `Web3Providers.send` is as follows`:
```ts
function send(method: string, params: Array<any>): Promise<any> {}
```
But the [`wallet_watchAsset` RPC method](https://docs.metamask.io/guide/rpc-api.html#wallet-watchasset) requires an object in the form of
```ts
interface WatchAssetParams {
  type: 'ERC20'; // In the future, other standards will be supported
  options: {
    address: string; // The address of the token contract
    'symbol': string; // A ticker symbol or shorthand, up to 11 characters
    decimals: number; // The number of token decimals
    image: string; // A string url of the token logo
  };
}
```
as a param. If I call `provider.send("wallet_watchAsset", [watchAssetParams])` it fails, instead I have to call `provider.send("wallet_watchAsset", watchAssetParams)`. But then TS errors, so I had to use `any` here...

src/views/Claim/hooks/useClaimableTokens.tsx
dohaki: Yea, if you look at the line below, we only enable the query if `!!acount`. But TS can not infer the type here, that the query will run only if the account is of type string. So I had to cast this one.

src/views/Claim/components/EligibleWallet.tsx
dohaki: In this case, I would leave it like this because I am spreading the props to a child component here https://github.com/across-protocol/frontend-v2/pull/233/files/d24ec3d16312d18632e193e0cc50654eb5080da4#diff-68562032e2b88e1241ad158234719813c580502fe19649ce19503174e7c85fb9R32

src/views/Claim/components/ClaimAirdrop.tsx
```diff
@@ -0,0 +1,195 @@
...
+          We want to thank you for being part of the community powering the
+          cheapest, fastest and most secure cross-bridge protocol.
+        </h6>
+      </TextContainer>
+      <BreakdownCardContainer>
+        <BreakdownTitle>Airdrop breakdown</BreakdownTitle>
+        <BreakdownStats>
+          <BreakdownRow>
+            <h6>Liquidity providing</h6>
+            <h6>{utils.formatUnits(claimable?.liquidityClaim || 0)} ACX</h6>
```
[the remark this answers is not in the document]
dohaki: Yes, makes sense! Changed

src/views/Claim/hooks/useClaimableTokens.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// TODO: use correct functions
+async function getClaimableTokens(account: string) {
+  await new Promise((resolve) => setTimeout(() => resolve(true), 5_000));
+  return {
+    liquidityClaim: utils.parseEther("10.322"),
+    bridgingClaim: utils.parseEther("22.012"),
+    communityClaim: utils.parseEther("41.124"),
+    totalClaim: utils.parseEther("73.458"),
```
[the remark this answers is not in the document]
dohaki: Eventually, it would make sense yes. But here I am parsing this just for stubbing purposes, so this will change when we implement the integration fetching logic. So I would just leave it for now.

PR #234, opened 2022-08-23
feat: enable coinbase wallet + extension

Tulun:
1. Adds in walletLink (extension) and Coinbase (mobile) to list of wallets.
2. Fixes minor issue with ensName.

Requires extensive QA for both mobile + desktop flow, but this should be all we need to do to enable the app on the current iteration.


src/utils/onboard.ts
```diff
@@ -18,10 +18,10 @@ export function onboardBaseConfig(): Initialization {
   return {
     dappId: onboardApiKey,
     networkId: hubPoolChainId,
-    hideBranding: true,
+    // hideBranding: true,
```
mrice32, line 21: What does this do?

src/utils/onboard.ts
```diff
@@ -33,7 +33,8 @@ export function onboardBaseConfig(): Initialization {
           },
           preferred: true,
         },
-        // { walletName: "coinbase", preferred: true },
+        { walletName: "coinbase", preferred: true, label: "Coinbase" },
+        { walletName: "walletLink", preferred: true, label: "Coinbase" },
```
mrice32, line 37: Why did we add labels to all three of these, but not tally?

mrice32 (commented):
Looks good! A few minor questions

src/utils/onboard.ts
Tulun: Want to make sure that it displays Coinbase. label is an optional field. Metamask shows as "MetaMask", which I think we wanted to adjust anyway.

We could display this as Coinbase (mobile) or Coinbase (desktop) or something as well.

mrice32: Okay, SG!

mrice32:
> from my testing theres not really a distinction in the UI between selecting the coinbase wallet vs metamask. having both installed, it seems like coinbase takes over. having just metamask installed, it seems like you can connect to it via coinbase or metamask. i dont think thats a problem for adding this, but something that still a problem is that coinbase does not tell you what network you are on, leaving users vulnerable to sending from wrong chain if UI desyncs.

Does this mean you _cannot_ use metamask if coinbase wallet is installed? Is that typical? i.e. is this the case for all dapps that use both? Or just an issue with onboard?

Tulun:
> > from my testing theres not really a distinction in the UI between selecting the coinbase wallet vs metamask. having both installed, it seems like coinbase takes over. having just metamask installed, it seems like you can connect to it via coinbase or metamask. i dont think thats a problem for adding this, but something that still a problem is that coinbase does not tell you what network you are on, leaving users vulnerable to sending from wrong chain if UI desyncs.
> 
> Does this mean you _cannot_ use metamask if coinbase wallet is installed? Is that typical? i.e. is this the case for all dapps that use both? Or just an issue with onboard?

This seems to be an issue with the extension itself. It seems like it uses the window.ethereum object and it appears to confuse the onboard integration. This seems to be true with the lts version of onboard too.

mrice32:
> > > from my testing theres not really a distinction in the UI between selecting the coinbase wallet vs metamask. having both installed, it seems like coinbase takes over. having just metamask installed, it seems like you can connect to it via coinbase or metamask. i dont think thats a problem for adding this, but something that still a problem is that coinbase does not tell you what network you are on, leaving users vulnerable to sending from wrong chain if UI desyncs.
> > 
> > 
> > Does this mean you _cannot_ use metamask if coinbase wallet is installed? Is that typical? i.e. is this the case for all dapps that use both? Or just an issue with onboard?
> 
> This seems to be an issue with the extension itself. It seems like it uses the window.ethereum object and it appears to confuse the onboard integration. This seems to be true with the lts version of onboard too.

Before merging, can we verify that this issue exists with HOP or some other dapp that supports both?

Tulun:
> > > > from my testing theres not really a distinction in the UI between selecting the coinbase wallet vs metamask. having both installed, it seems like coinbase takes over. having just metamask installed, it seems like you can connect to it via coinbase or metamask. i dont think thats a problem for adding this, but something that still a problem is that coinbase does not tell you what network you are on, leaving users vulnerable to sending from wrong chain if UI desyncs.
> > > 
> > > 
> > > Does this mean you _cannot_ use metamask if coinbase wallet is installed? Is that typical? i.e. is this the case for all dapps that use both? Or just an issue with onboard?
> > 
> > 
> > This seems to be an issue with the extension itself. It seems like it uses the window.ethereum object and it appears to confuse the onboard integration. This seems to be true with the lts version of onboard too.
> 
> Before merging, can we verify that this issue exists with HOP or some other dapp that supports both?

It does exist with Hop. 

It actually still exists with the v2 upgrade :(. See: https://github.com/across-protocol/frontend-v2/pull/237

mrice32 (approved):
LGTM!

Tulun:
v2 upgrade done. Can close.

PR #246, opened 2022-08-30
feat: Disable Arbitrum Routes and show banner when 

Tulun:
1. Show banner when REACT_APP_ARBITRUM_DISABLED is truthy.
2. Delete routes for Arbitrum.
Screen Shot 2022-08-30 at 1 55 53 PM
Screen Shot 2022-08-30 at 1 55 49 PM


daywiss (approved):
did some QA on this, cannot access arbitrum (good). Also multiple banners show up correctly by stacking.  Only thing id be concerned with is restoring abitrum correctly, need to make sure those removed routes can be undone without error.

mrice32:
> did some QA on this, cannot access arbitrum (good). Also multiple banners show up correctly by stacking. Only thing id be concerned with is restoring abitrum correctly, need to make sure those removed routes can be undone without error.

We'll just do a revert. We should ideally not touch that file in the meantime.

PR #252, opened 2022-08-31
Blockchain e2e tests

Tulun:
1. bridge.cy.ts tests submitted a transaction.
2. getProvider returns our Mock Provider when window.isCypress = true
3. pool.cy.ts tests add and remove liquidity.


hardhat-scripts/helpers/createERC20ContractInstance.js
```diff
@@ -0,0 +1,15 @@
...
+
+const ethers = require("ethers");
+
+// Use contracts-node because this script is being run in node.
+const { getAbi, getAddress } = require("@uma/contracts-node");
+
+const abi = getAbi("ERC20");
+
+// Default to UMA Mainnet Contract Address.
+module.exports = async function createERC20ContractInstance(signer, networkId) {
```
daywiss, line 11: you probably want to have address as a parameter here, or name this differently

src/hooks/useBalance.ts
```diff
@@ -66,6 +66,7 @@ const QueryBalanceBySymbol =
     tokenSymbolToQuery?: string;
     blockNumberToQuery?: number;
     config: ConfigClient;
+    testProvider?: ethers.providers.Web3Provider;
```
daywiss, line 69: i dont think this is the right approach to take here. The business logic should not be aware its in a testing environment if at all possible.  Would much rather you alter the constants file in the spots where providers are defined:

- add a new provider urls object called providerTestUrls [here](https://github.com/across-protocol/frontend-v2/blob/master/src/utils/constants.ts#L468)
- switch the variable providerUrlsTable based on if this is testing [here](https://github.com/across-protocol/frontend-v2/blob/master/src/utils/constants.ts#L484)

everything else should work the same after this point, as long as the app uses `getProvider` everywhere, [here](https://github.com/across-protocol/frontend-v2/blob/master/src/utils/constants.ts#L496)

daywiss (changes requested):

daywiss (approved):
looks good!

cypress/e2e/bridge.cy.ts
```diff
@@ -6,18 +6,30 @@ describe("bridge", () => {
...
+    cy.dataCy("bridge-amount-input").click().type("1");
 
     cy.dataCy("send").should("be.disabled");
     cy.dataCy("fees-box").should("be.visible");
   });
 
-  it("display balance of 'from' asset ", () => {
-    cy.connectInjectedWallet("connect-wallet");
+  it("Clicks send and submits a tx and shows the success page", () => {
+    cy.wait(6000);
```
mrice32, line 24: These waits are interesting. Is this the recommended approach? It seems pretty inefficient as opposed to some sort of function that you could call that would allow you to _wait until the condition is met_ with a high timeout.

Note: in their docs, I see a call out that you _shouldn't_ need waits in your tests https://docs.cypress.io/guides/overview/why-cypress#Features.

mrice32 (commented):
This is cool! One comment about the waiting.

cypress/e2e/bridge.cy.ts
```diff
@@ -1,23 +1,30 @@
...
+  it("display balance of 'from' asset ", () => {
+    cy.connectInjectedWallet("connect-wallet");
+    cy.dataCy("balance").should("be.visible");
+  });
+
   it("render fees box on input", () => {
-    cy.dataCy("select-from-chain").click();
-    cy.dataCy("from-chain-1").click();
-    cy.dataCy("amount-input").type("1");
+    cy.wait(7000);
```
mrice32, line 14: Just OOC, why do we still need this wait?

cypress/e2e/pool.cy.ts
```diff
@@ -7,18 +7,49 @@ describe("pool", () => {
...
+
+    cy.get("#amount").should("not.be.disabled");
+    cy.get("#amount").click().type("100");
+
+    cy.dataCy("add-liquidity-button").click();
+    cy.dataCy("bouncing-loader").should("be.visible");
+    cy.wait(30000);
+    // TX won't resolve to non-loading state because of dependency on NotifyJS.
+    cy.visit("/pool");
+    cy.wait(4000);
```
mrice32, line 37: Same here.

mrice32 (approved):

dohaki (approved):
Nice 💪

cypress/e2e/bridge.cy.ts
Tulun: This one seems to fail without a wait.

cypress/e2e/pool.cy.ts
Tulun: > Same here.

This one was removable :)

PR #258, opened 2022-09-02
feat: :sparkles: add new optional parameter to serverless CG route

james-a-morris:
This PR is aimed at adding an additional query parameter to the `api/coingecko` endpoint. When the optional `baseCurrency` is supplied, the endpoint will return the price of the token in terms of the specified currency. 

However, if the parameter is not supplied, the default case will be enabled which assumes a base currency of ETH (which is what is done today).

`baseCurrency` is not case sensitive. If it is not eth, btc, or usd, the SDK's symbol mapping will be used to look it up independently.

Breaking Changes: None. Everything is returned in terms of ETH today, so this works as expected.


api/_utils.js
```diff
@@ -433,6 +438,15 @@ const filterMapArray = (array, filterFn, mappingFn, mapFirst) => {
   return array.reduce(reducerFn, []);
 };
 
+/**
+ * Tests if input value is a numeric string
+ * @param {string} value The value that will be tested
+ * @returns A boolean if value matches a string of numeric digits from 0 - 9
+ */
+const isStringNumeric = (value) => {
```
nicholaspai, line 446: Why not just use `isNaN`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

api/coingecko.js
```diff
@@ -1,17 +1,27 @@
...
       throw new InputError("Must provide l1Token as query param");
 
+    if (!isStringNumeric((destinationId = destinationId ?? "1"))) {
+      throw new InputError("Destination must a non-negative integer");
+    }
+
     l1Token = ethers.utils.getAddress(l1Token);
 
-    const price = await getTokenPrice(l1Token, 1);
+    const price = await getTokenPrice(l1Token, Number(destinationId));
```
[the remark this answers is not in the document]
nicholaspai: Currently all prices are returned with ETH as the denominator, does this PR break this assumption if we set the chainId to `137`? Or is returning with ETH as the denominator still expected?

api/coingecko.js
```diff
@@ -5,13 +5,18 @@ const { getLogger, InputError, isString, getTokenPrice } = require("./_utils");
 const handler = async (request, response) => {
   const logger = getLogger();
   try {
-    let { l1Token } = request.query;
+    let { l1Token, destinationId } = request.query;
```
mrice32, line 8: This seems to be more limiting than allowing the user to specify base currency. It also seems less natural for an endpoint that just provides price values.

For instance, one might imagine that an across bot wants to query the dollar value of a currency (the dataworker does this). Any reason not to have the caller figure out the appropriate base currency symbol for their chain and pass that instead?

Thoughts?

mrice32 (commented):
Just one high level comment!

api/coingecko.js
mrice32: Great points!!

Yeah, I think this endpoint should directly call the CG client so it can be general purpose. Ideally, we will actually have some sort of generalized class that can grab arbitrary prices that would be backed by this endpoint.

But would be interested to hear the rest of the team's feedback on that.

mrice32: Thoughts @nicholaspai?

mrice32 (commented):
https://github.com/across-protocol/frontend-v2/pull/258#discussion_r965183330

api/coingecko.js
nicholaspai: I've talked with Matt and he's going to take over this PR. I think we should pay the upfront technical debt and have this `coingecko` endpoint instantiate the `BaseQuery` class to directly call the CG API and pass through the requested param's `baseCurrency`. This effectively implements Matt's suggestion to 

>  I think this endpoint should directly call the CG client so it can be general purpose

This will require some changes to #256 but we'll cross that road after this PR

nicholaspai (changes requested):
https://github.com/across-protocol/frontend-v2/pull/258#discussion_r966172962

api/_utils.ts
```diff
@@ -302,10 +302,15 @@ export const queries: Record<number, () => QueryBase> = {
  * @returns An instance of the `RelayFeeCalculator` for the specific chain specified by `destinationChainId`
  */
 export const getRelayerFeeCalculator = (destinationChainId: number) => {
+  const queryFn = queries[destinationChainId];
+  if (typeof queryFn === undefined) {
```
pxrl, line 306: Is this working as intended? typeof produces a string, so won't this always evaluate false?

Separately, I did something similar and was encouraged to use:

```suggestion
  if (queryFn === undefined) {
```

api/_utils.ts
```diff
@@ -302,10 +302,15 @@ export const queries: Record<number, () => QueryBase> = {
...
+    throw new InputError("Invalid destination chain Id");
```
pxrl, line 307: For any troubleshooting purposes that might arise it could be nice to include the actual chain ID that was supplied here:

```suggestion
    throw new InputError(`Invalid destination chain Id: ${chainId}`);
```

api/coingecko.ts
```diff
@@ -2,25 +2,82 @@ import { VercelResponse } from "@vercel/node";
...
+        l1Token,
+        cgBaseCurrency
+      );
+    } else {
+      // No match, so we try to look up the base currency directly.
+      const baseCurrencyToken = SymbolMapping[baseCurrency];
+
+      if (!baseCurrencyToken)
+        throw new InputError(
+          "baseCurrency not supported in Coingecko and not found in address mapping"
```
pxrl, line 52: Might also be nice to include the requested base currency in the log message:

```suggestion
          `base currency ${baseCurrency} not supported in Coingecko and not found in address mapping"
```

api/coingecko.ts
```diff
@@ -2,25 +2,82 @@ import { VercelResponse } from "@vercel/node";
...
   const logger = getLogger();
   try {
     if (!isString(l1Token))
       throw new InputError("Must provide l1Token as query param");
 
+    // Start with the symbol being strictly upper-case, as that's how symbols are typically represented.
+    if (!isString(baseCurrency)) baseCurrency = "ETH";
+    else baseCurrency = baseCurrency.toUpperCase();
+
     l1Token = ethers.utils.getAddress(l1Token);
```
pxrl, line 27: Have you tested this with a range of upper-case addresses? I _think_ I've seen it throw exceptions on some addresses, and I _think_ that was because the first alpha character in the address was coincidentally upper-case in the checksummed version, so getAddress() decides it has been supplied a checksummed address and then gets upset when it's not.

From the ethers docs:

```
// Throws if a checksummed address is provided, but a
// letter is the wrong case
// ------------v (should be lower-case)
getAddress("0x8Ba1f109551bD432803012645Ac136ddd64DBA72")
// [Error: bad address checksum] {
//   argument: 'address',
//   code: 'INVALID_ARGUMENT',
//   reason: 'bad address checksum',
//   value: '0x8Ba1f109551bD432803012645Ac136ddd64DBA72'
// }
```

pxrl, line 27: I had a comment here about validating the length of l1Token and normalising it (in the event that it's supplied upper-case for some reason?), but realised that it's not really the focus of this review, and can see that the exception thrown by ethers is cleanly presented to the user. Just wanted to note it in case anyone saw the comment and was confused that it's gone.

api/coingecko.ts
```diff
@@ -2,25 +2,82 @@ import { VercelResponse } from "@vercel/node";
...
+      if (l1Token.toLowerCase() === baseCurrencyToken.address.toLowerCase())
+        price = 1;
+      else {
+        // Always use usd as the base currency for the purpose of conversion.
+        const [price1, price2] = await coingeckoClient.getContractPrices(
+          [l1Token, baseCurrencyToken.address],
+          "usd"
+        );
+
+        console.log(price1, price2);
```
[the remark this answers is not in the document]
mrice32: Oh, no, this was just a debug log that I mistakenly left there. WIll remove!

api/_utils.ts
mrice32: Ah yes, great point. You're 100% right.

api/_utils.ts
mrice32: Agreed!

api/coingecko.ts
mrice32: Good idea!

api/coingecko.ts
```diff
@@ -2,25 +2,82 @@ import { VercelResponse } from "@vercel/node";
...
+    // Start with the symbol being strictly upper-case, as that's how symbols are typically represented.
+    if (!isString(baseCurrency)) baseCurrency = "ETH";
+    else baseCurrency = baseCurrency.toUpperCase();
+
     l1Token = ethers.utils.getAddress(l1Token);
 
-    const price = await getTokenPrice(l1Token, 1);
+    // Coingecko doesn't seem to be case sensitive, but there doesn't seem to be anything in their documentation
+    // guaranteeing this, so to be safe, we lower case before sending to CG.
+    const cgBaseCurrency = baseCurrency.toLowerCase();
```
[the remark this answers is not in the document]
mrice32: We don't, but we need an extra upper case in the else below.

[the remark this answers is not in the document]
mrice32: Updated!

api/coingecko.ts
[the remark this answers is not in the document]
mrice32: Done!

api/_constants.ts
```diff
@@ -58,3 +58,7 @@ export const relayerFeeCapitalCostConfig = {
 };
 
 export const BLOCK_TAG_LAG = -1;
+
+// Note: this is a small subset of all the supported base currencies, but since we don't expect to use the others,
+// we've decided to keep this list small for now.
+export const SUPPORTED_CG_BASE_CURRENCIES = new Set(["btc", "eth", "usd"]);
```
nicholaspai, line 64: Is there any reason to add "btc" now versus just waiting until we need it?

api/coingecko.ts
```diff
@@ -2,25 +2,78 @@ import { VercelResponse } from "@vercel/node";
...
+        l1Token,
+        baseCurrency
+      );
+    } else {
+      // No match, so we try to look up the base currency directly.
+      const baseCurrencyToken = SymbolMapping[baseCurrency.toUpperCase()];
+
+      if (!baseCurrencyToken)
+        throw new InputError(
+          `base currency ${baseCurrency} not supported in Coingecko and not found in address mapping`
```
nicholaspai, line 50: Does this error get thrown when the base current isn't supported in coingecko? It looks like it gets thrown only when we don't expect it, rather than looking it up against the CG API list explicity

nicholaspai (approved):
LGTM

api/_constants.ts
mrice32: Yeah, I didn't have a very good reason to do this. I think you're right.

api/coingecko.ts
mrice32: Good point, I will update the language.

mrice32: Updated.

api/_constants.ts
mrice32: Updated.

PR #260, opened 2022-09-06
feat: speed up modal

dohaki:
This adds a first-pass version of the speed up modal. There is one optional TODO/blocker but this can be adjusted in an upcoming PR.

Testing this feature is a bit tricky. I tested it by setting a very low relayer in the `useBridge.ts` hook locally, e.g.
```
relayerFeePct: ethers.utils.parseEther("0.0001"), // 0.01%
```

**TODOs:**
- [x] Use `/suggested-fees` API / hook from https://github.com/across-protocol/frontend-v2/pull/256 (optional)

Videos of user flow

https://user-images.githubusercontent.com/17645687/190171829-df8abbe4-24da-4977-aac6-9a8d671bcb77.mov

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -162,7 +162,5 @@ export function formatPoolAPY(
 }
 
 export function formatWeiPct(wei: ethers.BigNumberish, precision: number = 3) {
-  return new Intl.NumberFormat("en-US", {
-    maximumFractionDigits: precision,
-  }).format(Number(ethers.utils.formatEther(wei)) * 100);
+  return (Number(ethers.utils.formatEther(wei)) * 100).toFixed(precision);
 }
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -84,6 +84,10 @@ export function formatUnits(
   return smallNumberFormatter(value);
 }
 
+export function makeFormatUnits(decimals: number) {
+  return (wei: ethers.BigNumberish) => formatUnits(wei, decimals);
+}
+
 export function formatEther(wei: ethers.BigNumberish): string {
   return formatUnits(wei, 18);
 }
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -157,11 +157,8 @@ export function formatPoolAPY(
   );
 }
 
-export function formatWeiPct(
-  wei: ethers.BigNumberish,
-  maxFractions: number = 3
-) {
+export function formatWeiPct(wei: ethers.BigNumberish, precision: number = 3) {
   return new Intl.NumberFormat("en-US", {
-    maximumFractionDigits: maxFractions,
+    maximumFractionDigits: precision,
   }).format(Number(ethers.utils.formatEther(wei)) * 100);
 }
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -162,5 +162,7 @@ export function formatPoolAPY(
 }
 
 export function formatWeiPct(wei: ethers.BigNumberish, precision: number = 3) {
-  return (Number(ethers.utils.formatEther(wei)) * 100).toFixed(precision);
+  return new Intl.NumberFormat("en-US", {
+    maximumFractionDigits: precision,
+  }).format(Number(ethers.utils.formatEther(wei)) * 100);
 }
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -156,3 +156,12 @@ export function formatPoolAPY(
     Number(ethers.utils.formatUnits(wei, decimals))
   );
 }
+
+export function formatWeiPct(
+  wei: ethers.BigNumberish,
+  maxFractions: number = 3
+) {
+  return new Intl.NumberFormat("en-US", {
+    maximumFractionDigits: maxFractions,
+  }).format(Number(ethers.utils.formatEther(wei)) * 100);
+}
--- a/src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
+++ b/src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
@@ -41,6 +41,7 @@ export function SpeedUpModal({ isOpen, onClose, txTuple }: Props) {
     setChain,
     suggestedRelayerFeePct,
     isFetchingFees,
+    speedUpTxLink,
   } = useSpeedUp(transfer, token);
 
   useEffect(() => {
@@ -144,6 +145,13 @@ export function SpeedUpModal({ isOpen, onClose, txTuple }: Props) {
               </ConfirmButton>
             </ButtonsRow>
           </>
+        )}
+        {speedUpTxLink && (
+          <SpeedUpTxLinkContainer>
+            <a href={speedUpTxLink} target="_blank" rel="noreferrer">
+              View transaction
+            </a>
+          </SpeedUpTxLinkContainer>
         )}
       </Content>
     </Overlay>
@@ -367,5 +375,16 @@ const CancelButton = styled(ButtonV2)`
     ::after {
       border: none;
     }
+  }
+`;
+
+const SpeedUpTxLinkContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+
+  a {
+    text-decoration: underline;
+    color: #9daab2;
   }
 `;
--- a/src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
+++ b/src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
@@ -52,23 +52,26 @@ export function SpeedUpModal({ isOpen, onClose, txTuple }: Props) {
     }
   }, [suggestedRelayerFeePct, didSetInitialFee]);
 
-  const handleInputChange = (e: React.FormEvent<HTMLInputElement>) => {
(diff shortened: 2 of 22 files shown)


src/utils/format.ts
```diff
@@ -156,3 +156,12 @@ export function formatPoolAPY(
     Number(ethers.utils.formatUnits(wei, decimals))
   );
 }
+
+export function formatWeiPct(
+  wei: ethers.BigNumberish,
+  maxFractions: number = 3
```
nicholaspai, line 162: nit: Could use "precision" instead of "maxFractions" which is more of a standard term, but your call

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
```diff
@@ -0,0 +1,354 @@
...
+  }, [suggestedRelayerFeePct, didSetInitialFee]);
+
+  const handleInputChange = (e: React.FormEvent<HTMLInputElement>) => {
+    const input = e.currentTarget.value;
+    setRelayFeeInput(input);
+
+    try {
+      validateFeeInput(input, {
+        maxFeeInput: Number(removePercentageSign(maxRelayFee)),
+        minFeeInput: 0.001,
```
nicholaspai, line 61: I assume these are %'s and therefore min fee should be 0.0003 (0.03%). Max fee lets just lower to like 25% for now

nicholaspai: Oh also, `minFee` should be Math.min(currentFee, 0.0003)

nicholaspai (commented):
Left some small nits on the fee calc

src/utils/format.ts
dohaki: Ok will change. Also, like the term "precision" better.

src/views/Transactions/hooks/useSpeedUp.tsx
```diff
@@ -0,0 +1,108 @@
...
+        newRelayerFeePct,
+        transfer.depositId,
+        depositorSignature
+      );
+      setSpeedUpTxLink(
+        getChainInfo(transfer.sourceChainId).constructExplorerLink(
+          txResponse.hash
+        )
+      );
+      await txResponse.wait();
```
dohaki, line 66: We can not use notify here because it doesn't support optimism and arbitrum.

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
dohaki: done

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
```diff
@@ -0,0 +1,390 @@
...
+  font-weight: 400;
+`;
+
+const InfoBox = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding: 16px;
+  background-color: rgba(249, 210, 108, 0.05);
+  border: 1px solid rgba(249, 210, 108, 0.15);
```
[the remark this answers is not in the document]
dohaki: So I prefer hex with opacity but in this case, I needed different opacity values for background and border.

src/utils/format.ts
```diff
@@ -156,3 +160,9 @@ export function formatPoolAPY(
     Number(ethers.utils.formatUnits(wei, decimals))
   );
 }
+
+export function formatWeiPct(wei: ethers.BigNumberish, precision: number = 3) {
+  return new Intl.NumberFormat("en-US", {
+    maximumFractionDigits: precision,
+  }).format(Number(ethers.utils.formatEther(wei)) * 100);
+}
```
Tulun, line 168: A lot of these seem pretty similar, there a reason we aren't largely just using formatEther directly?

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
```diff
@@ -0,0 +1,386 @@
...
+import { SupportedTxTuple } from "../../types";
+
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
+  txTuple: SupportedTxTuple;
+};
+
+const maxRelayFee = 0.25; // 25%
+const minRelayFee = 0.0003; // 0.03%
```
Tulun, line 27: Might be better if this lived in constants. Also, is this configurable?

src/views/Transactions/components/SpeedUpModal/SpeedUpStats.tsx
```diff
@@ -0,0 +1,129 @@
...
+          ).toString()}
+        </div>
+      </StatRow>
+      <Divider />
+      <StatRow highlightValue>
+        <div>New fee %</div>
+        <div>
+          {inputError || isInitiallyFetchingFees || isInputInvalid
+            ? "-"
+            : appendPercentageSign(feeInput)}
```
Tulun, line 71: Couldn't you just add a % sign here with string interpolation?

src/views/Transactions/components/SpeedUpModal/SpeedUpStats.tsx
```diff
@@ -0,0 +1,129 @@
...
+        </div>
+      </StatRow>
+      <StatRow highlightValue>
+        <div>New fee in {token.symbol}</div>
+        <div>
+          {inputError || isInitiallyFetchingFees || isInputInvalid
+            ? "-"
+            : formatTokenUnits(
```
Tulun, line 79: I probably wouldn't do wei math division in line a render function.

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
```diff
@@ -0,0 +1,386 @@
...
+      <ButtonsRow>
+        <CancelButton size="md" onClick={onClose}>
+          Close
+        </CancelButton>
+      </ButtonsRow>
+    </>
+  );
+}
+
+function validateFeeInput(
```
Tulun, line 231: Shouldn't this in be in your util file?

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
```diff
@@ -0,0 +1,386 @@
...
+const SpeedUpTxLinkContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+
+  a {
+    text-decoration: underline;
+    color: #9daab2;
+  }
+`;
```
Tulun, line 386: There's a lot of code in this file, might be worth breaking out the styled components into their own file.

src/views/Transactions/hooks/useSpeedUp.tsx
Tulun: Might be nice to use Notify for supported chains, though!

Tulun (commented):
Some initial comments.

src/views/Transactions/components/SpeedUpModal/SpeedUpStats.tsx
dohaki: I opted to use this helper because I allow the user to input the symbol `%`.  And this helper makes sure that there are not double percentage symbols.

src/views/Transactions/hooks/useSpeedUp.tsx
dohaki: I added a new hook [`useNotify`](https://github.com/across-protocol/frontend-v2/pull/260/commits/77973052970c9cd491aef2177250341107dc1e09) that will use notify if supported and fallback to `txResponse.wait` if not. I also discovered this one https://docs.blocknative.com/notify#notification. So we could actually create the same behavior by triggering our own notifications if the chain is not supported. I would do that in a separate PR though.

src/utils/format.ts
dohaki: Not sure about the other ones, but this formatter, in particular, is used to display percentages in `BigNumber` format as values from `0% - 100%` with a max. precision. So only with `formatEther`, I would still need to do these transformations.

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
dohaki: Ok moved them. AFAIK these are not configurable but more like hard-capped input values.

src/views/Transactions/components/SpeedUpModal/SpeedUpStats.tsx
dohaki: I moved the calculation into a separate function. But why wouldn't you do the wei math division inline a render function? It is the same if I do it inline with JSX or in a separate function, no?

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
dohaki: Moved it

src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx
dohaki: Done 👍

Tulun (approved):

PR #263, opened 2022-09-07
feat: Add general maintenance banner.

Tulun:
Proposed solution for:  https://linear.app/uma/issue/ACX-260/prepare-banner-warning-users-that-bridge-will-be-offline-before-and

Make a generalized configurable message that can be changed in Vercel. We can't do formatting much this way but it might be useful to have something we can do without a PR in Vercel at a moment's notice.

To test, you should start this branch locally and add to your local .env, as I don't want to add this to environment variables until we are ready.

Screen Shot 2022-09-07 at 11 12 49 AM

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -65,3 +65,6 @@ REACT_APP_REWARDS_BANNER_WARNING=
 
 # Define Hardhat Infura ID for testing
 HARDHAT_INFURA_ID= 
+
+# Define a configurable message to display to users in environment.
+REACT_APP_GENERAL_MAINTENANCE_MESSAGE = Message to display in top banner.
--- a/.env.example
+++ b/.env.example
@@ -67,4 +67,4 @@ REACT_APP_REWARDS_BANNER_WARNING=
 HARDHAT_INFURA_ID= 
 
 # Define a configurable message to display to users in environment.
-REACT_APP_GENERAL_MAINTENANCE_MESSAGE = Message to display in top banner.
+REACT_APP_GENERAL_MAINTENANCE_MESSAGE=Message to display in top banner.
--- a/.env.example
+++ b/.env.example
@@ -67,4 +67,4 @@ REACT_APP_REWARDS_BANNER_WARNING=
 HARDHAT_INFURA_ID= 
 
 # Define a configurable message to display to users in environment.
-REACT_APP_GENERAL_MAINTENANCE_MESSAGE=Message to display in top banner.
+REACT_APP_GENERAL_MAINTENANCE_MESSAGE="Message to display in top banner."
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -18,6 +18,7 @@ import {
   enableMigration,
   WrongNetworkError,
   rewardsBannerWarning,
+  generalMaintenanceMessage,
 } from "utils";
 import { ReactComponent as InfoLogo } from "assets/icons/info-24.svg";
 import Toast from "components/Toast";
@@ -64,6 +65,9 @@ const Routes: React.FC = () => {
 
   return (
     <>
+      {generalMaintenanceMessage && (
+        <SuperHeader size="lg">{generalMaintenanceMessage}</SuperHeader>
+      )}
       {disableDeposits && (
         <SuperHeader>
           Across is experiencing issues. Deposits are currently disabled into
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -65,3 +65,6 @@ REACT_APP_REWARDS_BANNER_WARNING=
 
 # Define Hardhat Infura ID for testing
 HARDHAT_INFURA_ID= 
+
+# Define a configurable message to display to users in environment.
+REACT_APP_GENERAL_MAINTENANCE_MESSAGE = Message to display in top banner.
```
nicholaspai, line 70: Is there a limit to this message length we should comment about?
```suggestion
REACT_APP_GENERAL_MAINTENANCE_MESSAGE=Message to display in top banner.
```

nicholaspai: Also i think you're inserting an accidental " " char at the beginning of the string

src/Routes.tsx
```diff
@@ -64,6 +65,9 @@ const Routes: React.FC = () => {
 
   return (
     <>
+      {generalMaintenanceMessage && (
+        <SuperHeader size="lg">{generalMaintenanceMessage}</SuperHeader>
+      )}
       {disableDeposits && (
```
nicholaspai, line 71: How does this look when both `disableDeposits` and `generalMaintenanceMessage` (and some of the other banners below) are all turned on?

src/utils/constants.ts
```diff
@@ -574,6 +574,8 @@ export const hubPoolAddress = routeConfig.hubPoolAddress;
 export const migrationPoolV2Warning =
   process.env.REACT_APP_MIGRATION_POOL_V2_WARNING;
 export const enableMigration = process.env.REACT_APP_ENABLE_MIGRATION;
+export const generalMaintenanceMessage =
+  process.env.REACT_APP_GENERAL_MAINTENANCE_MESSAGE;
```
nicholaspai, line 578: Should we enforce a message length here?

nicholaspai (approved):
Have questions about:
- enforcing message length
- How stacked banners work and look

src/Routes.tsx
Tulun: They should stack. It's similar to the screenshot I've shown above.

src/utils/constants.ts
Tulun: Did you want to do an assert? It's probably not the best idea for the app to crash because we added a slightly too long message in Vercel. What would be the fallback if we added this?

And how many characters are you thinking? 256, 512?

.env.example
```diff
@@ -65,3 +65,6 @@ REACT_APP_REWARDS_BANNER_WARNING=
 
 # Define Hardhat Infura ID for testing
 HARDHAT_INFURA_ID= 
+
+# Define a configurable message to display to users in environment.
+REACT_APP_GENERAL_MAINTENANCE_MESSAGE=Message to display in top banner.
```
mrice32, line 70: Why does this one have a value while the others don't?

Also, doesn't this require quotes?

Tulun: Worked with the message I tested with but yes, good point, it wouldn't work always:

https://stackoverflow.com/questions/71538752/when-are-quotes-needed-in-env-file

src/utils/constants.ts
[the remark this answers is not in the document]
Tulun: Are you asking if we should assert existence first?

dohaki (approved):
LGTM 👍

daywiss (approved):
looks good. perhaps if u needed links or formatting in the message we could require it be specified as an html string in the env, but i think this is a fine solution

src/utils/constants.ts
Tulun: 500 or 1000 is probably fine. It would basically render ok but it would probably take up a *ton* of real estate on the UI. If you want, we could truncate the message if we think that'll be a better UX and note we should try to keep it under a certain character length.

nicholaspai: Lets set to 512 for now

nicholaspai: actually, i'd rather have the app not break. lets leave unrestricted for now and if there are issues we can shorten the message.

nicholaspai (approved):
https://github.com/across-protocol/frontend-v2/pull/263#discussion_r966169479

nicholaspai (approved):

PR #270, opened 2022-09-13
fix: Disable all routes in API

nicholaspai:
To prepare for merge, disable all API routes


Tulun:
Have a different proposed solution -- might suggest closing this.

nicholaspai:
> Have a different proposed solution -- might suggest closing this.

@Tulun this is primarily going to fix the API, but I'll change how this is done so it doesn't affect the front end

mrice32 (approved):
Looks great! I'm assuming we can't remove everything from the json because only the transfer page is supposed to be disabled in the FE? And removing all routes would disable the pool page drop down too?

PR #273, opened 2022-09-13
feat: Update time estimate for ETH2

Tulun:
Adjust returned estimate string.

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -165,7 +165,7 @@ export const getConfirmationDepositTime = (
 ) => {
   // Add this estimate for Ethereum 2.0
   if (toChain === ChainId.MAINNET) {
-    return "~4-7 minutes";
+    return "~6-7 minutes";
   }
   if (amount.lte(limits.maxDepositInstant)) {
     // 1 bot run, assuming it runs every 2 minutes.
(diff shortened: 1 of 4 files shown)


src/utils/bridge.ts
```diff
@@ -163,6 +163,10 @@ export const getConfirmationDepositTime = (
   limits: BridgeLimits,
   toChain: ChainId
 ) => {
+  // Add this estimate for Ethereum 2.0
+  if (toChain === ChainId.MAINNET) {
```
mrice32, line 167: I think this may be a bit more complicated. IIUC, the follow distance depends on the fromChain, not the toChain. So we are increasing the delay if you're coming _from_ mainnet, not to mainnet. This means:
1. We'll need to pull in fromChain in this method.
2. We'll probably need to rework this method such that each toChain has a time range (x-y) and then each fromChain has its own "delay" that it adds to that range.

Does that make sense? Thoughts?

Tulun: Sure Matt, what range should I put for each chain?

nicholaspai: So my suggestion is:
- Define a `DEPOSIT_DELAY` environment variable that is a string that can be parsed into type `[chainId]: number` similar to `MIN_DEPOSIT_CONFIRMATIONS` [here](https://github.com/across-protocol/frontend-v2/pull/272#discussion_r971245747.
- Instead of hard coding the time delay per type of deposit, determine it via a simple formula where there is some fixed delay that we can add the `DEPOSIT_DELAY` for the `fromChain` to.

PR #274, opened 2022-09-13
fix: pool removal crashing with small amounts

daywiss:

Motivation

UI was crashing when inputting very small numbers to withdraw from pool

Solution

We have to call tofixed on the decimal when passing into the preview. unfortunately for now the sdk specifies this is passed as a decimal, and probably should be sanitizing in there, but this will acheive the same effect easier by making the change on the FE


Tulun (approved):
LGTM

PR #276, opened 2022-09-13
chore(package): Bump Across SDK-v2 => v0.1.25

pxrl:
**Overview**
This introduces price request bundling and light caching to the Coingecko
class, which should help to save on repetitive price lookups from the
backend/API coingecko endpoint.

Price lookups initiated from the front end _might_ be completed slightly
faster than previously, owing to the removal of the synchronous price
lookup to CoinGecko.

**Acceptance Criteria**
Users should not notice any functional differences. All price lookups should
continue to work, and price-related lookup errors should not be seen in the
console logs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@across-protocol/contracts-v2": "^0.0.34",
-    "@across-protocol/sdk-v2": "^0.1.24",
+    "@across-protocol/sdk-v2": "^0.1.25",
     "@datapunt/matomo-tracker-js": "^0.5.1",
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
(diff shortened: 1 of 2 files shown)


pxrl:
FWIW, logs like these are evidence that the change is working as expected.

In this screenshot, the list of tokens included in the CoinGecko lookup increases as additional tokens are queried (i.e. we perform batch price lookups).

2022-09-13-223916_1828x103_scrot

In this screenshot, it's possible to see that the FE is suppressing additional price lookups. This is based on the observation that token prices are updated by CoinGecko much less frequently than every ~20s, so there's no point in escalating the price lookup each time. Instead, prices are held until their timestamp is > 300s (configurable).

2022-09-13-223745_1032x530_scrot

It's worth noting that the fact that this is visible via the FE is probably a pre-existing mistake - afaik all price lookups from the FE should be of the "cached" form (i.e. routed via the API). This mistake had previously been identified, but has apparently not been fixed yet. In any case, the primary benefit it in the backend/API implementation (which also uses this updated coingecko implementation).

mrice32 (approved):
LGTM!

PR #299, opened 2022-09-23
Add bundle analyse script

alter-eggo:
This pr adds analyse script to check final bundle size, chunks and it contents with the help of webpack-bundle-analyzer.
It helps heavily to understand how to optimise app.
Current bundle state: 
Screenshot 2022-09-23 at 21 27 40


alter-eggo:
ok, I see, thanks for the explanation here @james-a-morris

PR #304, opened 2022-09-27
feat:  Jason/acx 304 rewards api integration with api

Tulun:
1. Integration with dev API.
2. Need to connect Discord user *and* link a wallet in order to test.

### DIFF
--- a/src/utils/serverless-api/prod/rewards.ts
+++ b/src/utils/serverless-api/prod/rewards.ts
@@ -1,0 +1,21 @@
+import axios from "axios";
+import { RewardsApiInterface } from "../types";
+
+export default function rewardsApiCall(
+  address: string,
+  jwt?: string
+): Promise<RewardsApiInterface> {
+  return axios
+    .get("/rewards", {
+      headers: {
+        Authorization: `Bearer ${jwt}`,
+      },
+      params: {
+        address,
+      },
+    })
+    .then((res) => {
+      console.log("res", res);
+      return {} as RewardsApiInterface;
+    });
+}
--- a/src/utils/serverless-api/prod/rewards.ts
+++ b/src/utils/serverless-api/prod/rewards.ts
@@ -7,16 +7,19 @@ export default async function rewardsApiCall(
   jwt?: string
 ): Promise<RewardsApiInterface> {
   try {
-    const response = await axios.get(
-      `${rewardsApiUrl}/airdrop/rewards?address=${address}`,
-      {
-        headers: {
-          Authorization: `Bearer ${jwt}`,
-        },
-      }
-    );
-
-    return response.data;
+    const response = await axios.get(`${rewardsApiUrl}/airdrop/rewards`, {
+      headers: {
+        Authorization: `Bearer ${jwt}`,
+      },
+      params: {
+        address,
+      },
+    });
+    if (response.data) {
+      return response.data;
+    } else {
+      return {} as RewardsApiInterface;
+    }
   } catch (err) {
     console.error(err);
     return {} as RewardsApiInterface;
--- a/src/utils/serverless-api/prod/rewards.ts
+++ b/src/utils/serverless-api/prod/rewards.ts
@@ -6,15 +6,13 @@ export default function rewardsApiCall(
   address: string,
   jwt?: string
 ): Promise<RewardsApiInterface> {
-  console.log("jwt", jwt);
   return axios
-    .get(`${rewardsApiUrl}/rewards?address=${address}`, {
+    .get(`${rewardsApiUrl}/airdrop/rewards?address=${address}`, {
       headers: {
         Authorization: `Bearer ${jwt}`,
       },
     })
     .then((res) => {
-      console.log("res", res);
-      return {} as RewardsApiInterface;
+      return res.data as RewardsApiInterface;
     });
 }
--- a/src/utils/serverless-api/prod/rewards.ts
+++ b/src/utils/serverless-api/prod/rewards.ts
@@ -1,17 +1,16 @@
 import axios from "axios";
 import { RewardsApiInterface } from "../types";
+import { rewardsApiUrl } from "utils";
 
 export default function rewardsApiCall(
   address: string,
   jwt?: string
 ): Promise<RewardsApiInterface> {
+  console.log("jwt", jwt);
   return axios
-    .get("/rewards", {
+    .get(`${rewardsApiUrl}/rewards?address=${address}`, {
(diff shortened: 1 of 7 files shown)


src/utils/serverless-api/prod/rewards.ts
```diff
@@ -0,0 +1,18 @@
...
+): Promise<RewardsApiInterface> {
+  return axios
+    .get(`${rewardsApiUrl}/airdrop/rewards?address=${address}`, {
+      headers: {
+        Authorization: `Bearer ${jwt}`,
+      },
+    })
+    .then((res) => {
+      return res.data as RewardsApiInterface;
+    });
```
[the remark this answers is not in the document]
Tulun: Fair, made it async await.

src/views/PreLaunchAirdrop/api/getPrelaunchRewards.ts
```diff
@@ -1,40 +1,38 @@
...
-//   .setProtectedHeader({ alg: "dir", enc: "A256GCM" })
-//   .setIssuedAt()
-//   .setIssuer("urn:example:issuer")
-//   .setAudience("urn:example:audience")
-//   .setExpirationTime("2h")
-//   .encrypt(new Uint8Array(32));
+import { mockServerlessAPI } from "utils";
+
 const TEST_JWT =
   "eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0..IFJ1oL_VAa-nnhEX.HmM1L9g8uBlrSggZeQr8f-cpIslh8rTnVe1HVXxf0Kgkm6GO-TmdDxokln4vG1udByCTgTAOvTAWVDcvyJy4L_tTxSLwhoaEBIYnN_G3lZXg4JYen7__7qqdI2oZOLTJ0iSVGoL4Ro_TvWe_ZPLTLugx-Figkr20pabxUorw7yw8mqWCLDuWR56ZpH94RUGN5CjRTs0z1FXPac0TKo_Q-gtLcmnpwj383A.ZYkHwm56_n5GHvB-7y6Tag";
```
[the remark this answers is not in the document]
Tulun: simplified :)

src/utils/serverless-api/prod/rewards.ts
```diff
@@ -0,0 +1,24 @@
+import axios from "axios";
+import { RewardsApiInterface } from "../types";
+import { rewardsApiUrl } from "utils";
+
+export default async function rewardsApiCall(
+  address: string,
+  jwt?: string
+): Promise<RewardsApiInterface> {
+  try {
+    const response = await axios.get(
```
nicholaspai, line 10: Can't you use an axios params method? I think they expose some sort of way to set request url params in a more axios native way

src/utils/serverless-api/prod/rewards.ts
```diff
@@ -0,0 +1,24 @@
...
+        headers: {
+          Authorization: `Bearer ${jwt}`,
+        },
+      }
+    );
+
+    return response.data;
+  } catch (err) {
+    console.error(err);
+    return {} as RewardsApiInterface;
```
nicholaspai, line 22: It's a bit inconsistent to me that it .data doesn't exist above the user gets undefined but you're using {} here to indicate the same-ish failure state. Should you check for whether .data is defined?

src/views/PreLaunchAirdrop/api/getPrelaunchRewards.ts
```diff
@@ -1,40 +1,38 @@
...
-      payout: "10",
-    },
-  });
+import { mockServerlessAPI } from "utils";
+
+const TEST_JWT = "eyTESTING123";
+
+export default async function getPrelaunchRewards(
+  address: string,
+  jwt?: string
```
nicholaspai, line 8: How do you handle this being undefined?

nicholaspai (approved):
Feel free to merge when ready as it looks like this PR facilitates the next stage of QA but I pointed out some syntactical inconsistencies

src/utils/serverless-api/prod/rewards.ts
Tulun: Yup, can swap over to passing in the params object.

src/utils/serverless-api/prod/rewards.ts
Tulun: I can check for existence of response.data.

As it is right now, empty object would return a largely disconnected state. I'm not sure we fully have a "bad state" (from an API perspective) case for the user.

src/views/PreLaunchAirdrop/api/getPrelaunchRewards.ts
Tulun: I believe the JWT is isn't required to do this call, but only to get the community rewards, but I could be mistaken.

src/utils/serverless-api/prod/rewards.ts
nicholaspai: Ok I just think you should be consistent with you return the invalid response case

src/utils/serverless-api/prod/rewards.ts
```diff
@@ -0,0 +1,27 @@
...
+        Authorization: `Bearer ${jwt}`,
+      },
+      params: {
+        address,
+      },
+    });
+    if (response.data) {
+      return response.data;
+    } else {
+      return {} as RewardsApiInterface;
```
mrice32, line 21: In general, we should avoid casts like this, as the caller will not know to expect an empty object.

It might be better to return null or undefined in this case, add it to the return type `Promise<RewardsApiInterface | null>`. In that case, typescript can ensure that the caller handles the case where null is returned and the app doesn't blow up.

mrice32 (commented):
Looks great! Only minor comments around error/edge case handling.

src/utils/serverless-api/prod/rewards.ts
Tulun: Sounds good to me.

src/utils/serverless-api/prod/rewards.ts
```diff
@@ -0,0 +1,24 @@
+import axios from "axios";
+import { RewardsApiInterface } from "../types";
+import { rewardsApiUrl } from "utils";
+
+export default async function rewardsApiCall(
+  address: string,
+  jwt?: string
```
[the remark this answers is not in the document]
Tulun: I *think* so as I think the JWT may only be protecting the linked community rewards. but I'm not certain.

PR #342, opened 2022-10-03
feat: Add Stories for widgets.

Tulun:
1. Added Stories for new widgets
2. Started unit tests.

Note: Blocked on an import issue with ethers in Jest, will likely need to resolve that to tackle more substantial unit tests.


src/utils/tests/format.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  isValidString,
+  shortenAddress,
+  shortenString,
+  shortenTransactionHash,
+  smallNumberFormatter,
+} from "../format";
+const VALID_ADDRESS = "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828";
+const TX_HASH =
+  "0xe5a0c976ca4d09ce6ea034101e78b1a6a9536940cb8f246e7b54d1fe16b8c125";
+describe("#format functions", () => {
```
dohaki, line 11: NIT: Instead of wrapping all test cases inside one describe-block, it could be cleaner to have one describe-block for each function? Like
```js
describe("#isValidString()", ...)
describe("#shortenAddress()", ...)
```
This makes it easier to use test helper methods like `.only`, `.skip`, etc. for specific functions.

src/stories/Alert.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import Alert from "components/Alert";
+
+// More on default export: https://storybook.js.org/docs/react/writing-stories/introduction#default-export
```
dohaki, line 4: NIT: Do you prefer to have these boilerplate comments as part of the stories? Or can they be removed?

dohaki (commented):
LGTM 👍 Left two nits

src/utils/tests/format.test.ts
Tulun: Adjusted!

src/stories/Alert.stories.tsx
Tulun: Yeah fair. removed.

src/utils/tests/format.test.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  isValidString,
+  shortenAddress,
+  shortenString,
+  shortenTransactionHash,
+  smallNumberFormatter,
+} from "../format";
+const VALID_ADDRESS = "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828";
+const TX_HASH =
+  "0xe5a0c976ca4d09ce6ea034101e78b1a6a9536940cb8f246e7b54d1fe16b8c125";
```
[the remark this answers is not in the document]
Tulun: As we don't really care what these values are (in fact, we don't even check if they are valid), I don't think it matters. I just wanted one that "looked" real. I adjusted the Hash value slightly, and the address is the UMA token.

PR #343, opened 2022-10-04
test: add basic set of airdrop e2e tests

dohaki:
This adds some basic /airdrop route e2e tests. For the more refined ones (connected wallets with different eligibility states), I don't know yet which approach we should take, i.e. I will open separate PRs.


Tulun (approved):
LGTM. Looks like you need a rebase, though!

PR #352, opened 2022-10-10
feat: add a way to specify an override provider for the API

mrice32:
This adds a way to specify an override URL for a particular network in the API.

This is done by adding `OVERRIDE_PROVIDER_[chainId] = https://...` as an env.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -185,6 +185,21 @@ export const bobaProvider = (): providers.StaticJsonRpcProvider =>
   new ethers.providers.StaticJsonRpcProvider("https://mainnet.boba.network");
 
 /**
+ * Resolves a fixed Static RPC provider if an override url has been specified.
+ * @returns A provider or undefined if an override was not specified.
+ */
+export const overrideProvider = (
+  chainId: string
+): providers.StaticJsonRpcProvider | undefined => {
+  const url = process.env[`OVERRIDE_PROVIDER_${chainId}`];
+  if (url) {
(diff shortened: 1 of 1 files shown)


api/_utils.ts
```diff
@@ -184,6 +184,21 @@ export const infuraProvider = (name: string) => {
...
   new ethers.providers.StaticJsonRpcProvider("https://mainnet.boba.network");
 
+/**
+ * Resolves a fixed Static RPC provider if an override url has been specified.
+ * @returns A provider or undefined if an override was not specified.
+ */
+export const overrideProvider = (
+  chainId: string
+): providers.StaticJsonRpcProvider | undefined => {
+  const url = process.env[`OVERRIDE_PROVIDER_${chainId}`];
```
amateima, line 194: In Vercel we already have a couple of env variables used to override the ones from Infura (`REACT_APP_CHAIN_<ID>_PROVIDER_URL`). Can we use the same format here?
Screenshot 2022-10-10 at 16 12 38

mrice32: Don't we leave those defined all the time?

mrice32: The intention of this new env variable is to _override_ the default and be removed afterwards. If I understand correctly, those are more like alternate URLs (alchemy) for pages where we don't want to use infura, right? So we can never remove them without breaking the site, no?

amateima: You're right

amateima (approved):

PR #358, opened 2022-10-10
feat: Fix capital charge at 1bp  for USDC/DAI/ETH

nicholaspai:


cc7768 (approved):
lgtm

nicholaspai:
Reminder to self. Need to update this modal also https://github.com/across-protocol/frontend-v2/blob/6d489f7e242731767e23be322c548358515f2922/src/views/Transactions/components/SpeedUpModal/SpeedUpModal.tsx#L71

api/_constants.ts
```diff
@@ -8,14 +8,14 @@ export const disabledL1Tokens = [
 
 export const relayerFeeCapitalCostConfig = {
   ETH: {
-    lowerBound: ethers.utils.parseUnits("0.0003").toString(),
-    upperBound: ethers.utils.parseUnits("0.0006").toString(),
+    lowerBound: ethers.utils.parseUnits("0.0001").toString(),
```
mrice32, line 11: Should we make this configurable via an env? That way they're synchronized across the API and FE?

mrice32: Oh I forgot the FE calls the API instead of computing itself. But either way, might be better to use an ENV.

nicholaspai: The FE reads from API we ended up removing the same constants file from the FE folder

src/utils/constants.ts
```diff
@@ -636,7 +636,7 @@ export const wbtcLpCushion = process.env.REACT_APP_WBTC_LP_CUSHION || "0";
 export const daiLpCushion = process.env.REACT_APP_DAI_LP_CUSHION || "0";
 
 export const maxRelayFee = 0.25; // 25%
-export const minRelayFee = 0.0003; // 0.03%
+export const minRelayFee = 0.0001; // 0.03%
```
dohaki, line 639: nit
```suggestion
export const minRelayFee = 0.0001; // 0.01%
```

dohaki (approved):
LGTM 👍 I left one nit

api/_constants.ts
mrice32: LGTM! But I do think that we might want to have this set via an env.

mrice32 (approved):

api/_constants.ts
```diff
@@ -8,14 +8,14 @@ export const disabledL1Tokens = [
...
+    upperBound: ethers.utils.parseUnits("0.0001").toString(),
     cutoff: ethers.utils.parseUnits("750").toString(),
```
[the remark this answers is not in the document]
nicholaspai: Nope not if they are the same value

PR #371, opened 2022-10-17
feat: post launch airdrop splash and info

dohaki:
This mostly ports the pre-launch splash- and info-flows to the post-lauch airdrop page. I also added a generic `<Text />` component with different sizes and scales consistently if on mobile, because I saw myself repeatedly adding these styles.

Just as an FYI on the merging/porting strategy of the `PreLaunchAirdrop` page and the previous `Claim` page: I created a new folder `Airdrop` that will be the post-launch airdrop page. I will copy/paste and slim-down/modify the existing components from  the `PreLaunchAirdrop`/`Claim` folders and put them in the `Airdrop` folder. This leads to quite a bit of duplicated code, but I intend to clean-up by removing the `PreLaunchAirdrop`/`Claim` folders at the end.


dohaki:
> LGTM - but should we also remove the `PrelaunchAirdrop` view? This should turn many of the file additions into file renaming which can preserve the git history.

Yea good point. I will delete them in this PR then. Or at least the ones that moved.

PR #382, opened 2022-10-21
feat: airdrop claim scraper and contracts integration

dohaki:
Integrates the airdrop claim page with the correct Scraper API endpoint and the `MerkleDistributor` contract. I tested this with a local Scraper API instance and a contract owned by me.

~I will leave this as a draft until proper testing with the official Goerli contract, staging Scraper API and distributor scripts can be done.~

Eligible wallets should be able to test the claim flow by visiting the Goerli build [here](https://goerli-frontend-v2-git-claim-airdrop-integration-uma.vercel.app/).


src/utils/config.ts
```diff
@@ -88,8 +91,13 @@ export class ConfigClient {
     return this.config.hubPoolAddress;
   }
   getAcceleratingDistributorAddress(): string {
+    // TODO: replace hardcoded goerli address with official address from package
     return "0xbcfbCE9D92A516e3e7b0762AE218B4194adE34b4";
   }
+  getAcrossTokenAddress(): string {
+    // TODO: replace hardcoded goerli address with official address from package
+    return "0x40153DdFAd90C49dbE3F5c9F96f2a5B25ec67461";
+  }
```
amateima, line 100: In the future can this be done automatically based on the network that the user is connected to? 🤔 

Also, replacing hardcoded addresses will be the subject of another PR?

src/utils/constants.ts
```diff
@@ -433,6 +433,9 @@ assert(
 
 // PROCESS.ENV variables
 export const rewardsApiUrl = process.env.REACT_APP_REWARDS_API_URL;
+export const airdropWindowIndex = Number(
+  process.env.REACT_APP_AIRDROP_WINDOW_INDEX || 0
+);
```
amateima, line 438: In the future we'll have multiple windows

src/views/Airdrop/components/BreakdownStats.tsx
```diff
@@ -14,10 +14,10 @@ import { AmountBreakdown } from "../hooks/useAirdropRecipient";
 
 export type Props = {
   isLoading?: boolean;
-  discord?: {
+  discord?: Partial<{
     discordName: string;
     discordAvatar?: string;
-  };
+  }> | null;
```
amateima, line 20: Are `optional` and `Partial` needed in the same place?

src/views/Staking/hooks/useStakingActionsResolver.ts
```diff
@@ -12,6 +11,7 @@ import {
   safeDivide,
   switchChain,
   toWeiSafe,
+  notificationEmitter,
 } from "utils";
 import { useStakingPoolResolver } from "./useStakingPoolResolver";
```
amateima, line 16: 👍

amateima (commented):
Looks awesome!

src/utils/config.ts
dohaki: > In the future can this be done automatically based on the network that the user is connected to? 🤔

Not necessarily based on the network the user is connected to, but based on the configured hub pool chain id. Similar to here
https://github.com/across-protocol/frontend-v2/pull/382/files/9a44750bcddc39f0bdab0c0f76dcc9fb48504634#diff-285dea82a685fcd275e82141070c164a79a6650d46573cfd6303381afc47b28bR119-R126

> Also, replacing hardcoded addresses will be the subject of another PR?

Yea, I would tackle this in a separate PR because we need to add some things to the `@across-protocol/across-token` package I assume.

src/utils/constants.ts
dohaki: Yes, but I think they would be handled via a different UI and therefore a different env var?

src/views/Airdrop/components/BreakdownStats.tsx
dohaki: Yes, unfortunately, we need this... If not, TypeScript complains

amateima (approved):

PR #383, opened 2022-10-24
improve(api): Add quote timestamp buffer to suggested-fees route

nicholaspai:
This should be combined with a change to the relayer code [here](https://github.com/across-protocol/relayer-v2/pull/340) that will make relayer ignore any quote times that do not sufficiently lag HEAD

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -58,7 +58,7 @@ export const relayerFeeCapitalCostConfig = {
 };
 
 // If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
-export const QUOTE_TIMESTAMP_BUFFER = 12 * 60 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
+export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = 12 * 60 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
 
 export const BLOCK_TAG_LAG = -1;
 
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -58,7 +58,7 @@ export const relayerFeeCapitalCostConfig = {
 };
 
 // If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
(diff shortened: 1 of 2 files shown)


api/_constants.ts
```diff
@@ -57,6 +57,9 @@ export const relayerFeeCapitalCostConfig = {
   },
 };
 
+// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
+export const QUOTE_TIMESTAMP_BUFFER = 12 * 60 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
```
mrice32: nit: naming issue, looks like below you import DEFAULT_QUOTE_TIMESTAMP_BUFFER not QUOTE_TIMESTAMP_BUFFER.

api/suggested-fees.ts
```diff
@@ -81,8 +94,15 @@ const handler = async (
...
-    const blockTag = isString(timestamp) ? latestBlock : BLOCK_TAG_LAG;
+    // to be "latest". If we're not using the block associated with the `timestamp` passed as a
+    // request param, then we should set the lag approx. equal to the # of blocks that the
+    // quote time buffer lags HEAD.
+    // Note: Its ok if the BLOCK_TAG_LAG doesn't exactly correspond to the parsedTimestamp, since we are returning
+    // an indicative fee, not one passed to a smart contract function. The relayer will compute the actual fee for the
+    // quote time.
+    const blockTag = isString(timestamp)
+      ? latestBlock
+      : (-1 * quoteTimeBuffer) / 60 / 12;
```
mrice32, line 105: I think we can actually get rid of this code entirely and just use latestBlock directly (or change the name above of latestBlock to blockTag).

The latest block should already take into account the delay, right?

Thoughts?

api/_constants.ts
mrice32: Ah, looks like you fixed.

api/suggested-fees.ts
nicholaspai: yeah on second review, you're right

mrice32 (approved):
LGTM!

PR #385, opened 2022-10-24
improve: Add deposit quote time buffer to Bridge tab

nicholaspai:
Should have been included [in this PR](https://github.com/across-protocol/frontend-v2/pull/383) to keep FE aligned with API

### DIFF
--- a/src/hooks/useBridge.ts
+++ b/src/hooks/useBridge.ts
@@ -20,10 +20,10 @@ import {
   parseUnits,
   trackEvent,
   formatUnits,
-  QUOTE_TIMESTAMP_BUFFER,
 } from "utils";
 import useReferrer from "./useReferrer";
 import { useToast } from "components/Toast/useToast";
+import getApiEndpoint from "utils/serverless-api";
 
 enum SendStatus {
   IDLE = "idle",
@@ -94,7 +94,6 @@ export function useBridge() {
   });
 
   const hasToApprove = !!allowance && amount.gt(allowance);
-  const hubPool = config.getHubPool();
 
   const send = async () => {
     // NOTE: the `toAddress` check is redundant, as status won't be "ready" if `toAddress` is not set, but it's here to make TS happy. The same applies for `block` and `fees`.
@@ -120,8 +119,8 @@ export function useBridge() {
         toChain: selectedRoute.toChain,
         isNative: selectedRoute.isNative,
         relayerFeePct: fees.relayerFee.pct,
-        timestamp: (await hubPool.getCurrentTime()).sub(QUOTE_TIMESTAMP_BUFFER),
         referrer,
+        timestamp: fees.quoteTimestamp,
       });
       // matomo track bridge
       trackEvent({
--- a/src/hooks/useBridge.ts
+++ b/src/hooks/useBridge.ts
@@ -23,7 +23,6 @@ import {
 } from "utils";
 import useReferrer from "./useReferrer";
 import { useToast } from "components/Toast/useToast";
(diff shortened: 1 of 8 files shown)


src/hooks/useBridge.ts
```diff
@@ -119,7 +120,7 @@ export function useBridge() {
         toChain: selectedRoute.toChain,
         isNative: selectedRoute.isNative,
         relayerFeePct: fees.relayerFee.pct,
-        timestamp: await hubPool.getCurrentTime(),
+        timestamp: (await hubPool.getCurrentTime()).sub(QUOTE_TIMESTAMP_BUFFER),
```
mrice32, line 123: We don't use the quote timestamp returned by suggested-fees?

nicholaspai: hm good question looks like this is the only param we don't use from API. I'll make some adjustments to the data returned by `getSuggestedFees` even though the quote time is not a "fee", it would make sense to return with this other data

src/utils/constants.ts
```diff
@@ -662,3 +662,7 @@ export const QUERIESV2 = {
 };
 
 export const insideStorybookRuntime = Boolean(process.env.STORYBOOK);
+
+export const QUOTE_TIMESTAMP_BUFFER = process.env.QUOTE_TIMESTAMP_BUFFER
+  ? Number(process.env.QUOTE_TIMESTAMP_BUFFER)
+  : 12 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
```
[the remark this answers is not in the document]
nicholaspai: I ended up removing this constant in favor of pulling the quote time from the API, can you help verify that the time is being pulled correctly from the API?

nicholaspai:
I've just tested this on the staging site and the quote time looks correctly set to 5mins behind latest

src/utils/bridge.ts
```diff
@@ -51,6 +52,7 @@ export async function getRelayerFee(
   relayerGasFee: Fee;
   relayerCapitalFee: Fee;
   isAmountTooLow: boolean;
+  quoteTimestamp?: ethers.BigNumber;
```
mrice32, line 55: Why optional?

src/hooks/useBridge.ts
```diff
@@ -119,8 +119,8 @@ export function useBridge() {
         toChain: selectedRoute.toChain,
         isNative: selectedRoute.isNative,
         relayerFeePct: fees.relayerFee.pct,
-        timestamp: await hubPool.getCurrentTime(),
         referrer,
+        timestamp: fees.quoteTimestamp ?? (await hubPool.getCurrentTime()),
```
mrice32, line 123: Why do we need the defaulting behavior?

mrice32 (commented):
Just a few minor comments!

src/utils/bridge.ts
nicholaspai: For testing purposes, the mocked API hardcodes the fees but can't really hardcode the quote time

src/hooks/useBridge.ts
nicholaspai: For testing

src/utils/bridge.ts
mrice32: I see. It would be ideal if we could change this, but I'm guessing there's no way to do this without giving the mock api access to the test node, right? And that would be difficult?

Maybe we could add a TODO?

mrice32 (approved):
LGTM! Maybe we could add a TODO for making the mock API return a dynamic timestamp?

PR #426, opened 2022-11-02
feat(coingecko api): Allow hardcoding token prices for unlisted tokens

kevinuma:
Because some tokens are not listed on Coingecko

### DIFF
--- a/api/coingecko.ts
+++ b/api/coingecko.ts
@@ -9,7 +9,7 @@ import { coingecko, relayFeeCalculator } from "@across-protocol/sdk-v2";
 
 const { Coingecko } = coingecko;
 const { SymbolMapping } = relayFeeCalculator;
-const { REACT_APP_COINGECKO_PRO_API_KEY } = process.env;
+const { REACT_APP_COINGECKO_PRO_API_KEY, FIXED_TOKEN_PRICES } = process.env;
 
 const handler = async (
   { query: { l1Token, baseCurrency } }: CoinGeckoInputRequest,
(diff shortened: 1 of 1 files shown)


api/coingecko.ts
```diff
@@ -33,43 +33,52 @@ const handler = async (
...
-        // comparing to the l1Token value.
-        const [tokenPriceUsd, basePriceUsd] =
-          price1.address.toLowerCase() === l1Token.toLowerCase()
-            ? [price1.price, price2.price]
-            : [price2.price, price1.price];
-
-        // Drop any decimals beyond the number of decimals for this token.
-        price = Number(
-          (tokenPriceUsd / basePriceUsd).toFixed(baseCurrencyToken.decimals)
+      if (SUPPORTED_CG_BASE_CURRENCIES.has(baseCurrency)) {
```
pxrl, line 46: Could this maybe be an "else if" ? It should remove a level of indentation for all the relocated code.

api/coingecko.ts
```diff
@@ -33,43 +33,52 @@ const handler = async (
       REACT_APP_COINGECKO_PRO_API_KEY
     );
```
pxrl, line 35: Do we want to add a comment about why we support hardcoded prices?

api/coingecko.ts
kevinuma: Updated

api/coingecko.ts
nicholaspai: +1 and add a comment about whether we plan to revert this logic longer term

nicholaspai (commented):
I kind of think we should just hard code the ACX price and add a comment about reverting it once we can replace it with coingecko. The API is pretty quick to redeploy, what do you think?

api/coingecko.ts
```diff
@@ -33,9 +33,16 @@ const handler = async (
...
+      [token: string]: { [baseCurrency: string]: number };
+    } = FIXED_TOKEN_PRICES !== undefined ? JSON.parse(FIXED_TOKEN_PRICES) : {};
     let price: number;
-
-    if (SUPPORTED_CG_BASE_CURRENCIES.has(baseCurrency)) {
+    if (
+      fixedTokenPrices[l1Token] !== undefined &&
+      !isNaN(fixedTokenPrices[l1Token][baseCurrency])
+    ) {
+      price = fixedTokenPrices[l1Token][baseCurrency];
```
nicholaspai, line 44: Should we verify that `baseCurrency` is in `SymbolMapping[baseCurrency.toUpperCase()];` like we do on line 46?

nicholaspai: Or check: `SUPPORTED_CG_BASE_CURRENCIES.has(baseCurrency)`

nicholaspai (commented):
Left some comments to make sure the setter of the .env var doesn't shoot themselves in foot

nicholaspai (approved):

PR #437, opened 2022-11-08
fix(limits): Fix minDeposit derivation on Polygon

pxrl:
The limits API endpoint assumes that the L1 (bridged) token is priced in the native gas token. getCachedTokenPrice() defaults to returning prices in ETH terms when baseCurrency is not specified.

When pricing the bridged token in ETH terms on Polygon, it incorrectly devalues the bridged token and therefore produces an artificially high minimum deposit amount.

Issue initially reported by the curious minds over at Rango Exchange.

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -106,7 +106,9 @@ const handler = async (
       hubPool.interface.encodeFunctionData("pooledTokens", [l1Token]),
     ];
 
-    let tokenPrice = await getCachedTokenPrice(l1Token);
+    // @todo: Generalise the resolution of chainId => gasToken.
+    const baseCurrency = destinationChainId === "137" ? "matic" : "eth";
+    const tokenPrice = await getCachedTokenPrice(l1Token, baseCurrency);
 
     const [
       relayerFeeDetails,


pxrl:
Test results show that only minDeposit is impacted - all other fields (maxdeposit, maxDepositInstant and maxDepositShortDelay) remain unchanged.

```
# Before this change
$ curl "http://127.0.0.1:3000/api/limits?token=0xff970a61a04b1ca14834a43f5de4533ebddb5cc8&desti >
{"minDeposit":"62246582","maxDeposit":"5760837115267","maxDepositInstant":"5760837115267","maxDepositShortDelay":"5760837115267"}


# After this change
$ curl "http://127.0.0.1:3000/api/limits?token=0xff970a61a04b1ca14834a43f5de4533ebddb5cc8&destinationChainId=137&originChainId=42161"
{"minDeposit":"54329","maxDeposit":"5760837115267","maxDepositInstant":"5760837115267","maxDepositShortDelay":"5760837115267"}
```

api/limits.ts
```diff
@@ -106,7 +106,9 @@ const handler = async (
       hubPool.interface.encodeFunctionData("pooledTokens", [l1Token]),
     ];
 
-    let tokenPrice = await getCachedTokenPrice(l1Token);
+    // @todo: Generalise the resolution of chainId => gasToken.
```
pxrl, line 109: For context, I'd probably propose to do this as part of dropping in the PriceClient from sdk-v2.

mrice32 (approved):

PR #444, opened 2022-11-10
fix: base rewards apy calculation and test

dohaki:
This corrects the base rewards APR calculation by using a common unit (USD) and also re-adding the fixed point adjustment.


src/hooks/useStakingPool.ts
```diff
@@ -179,13 +187,10 @@ const fetchStakingPool = async (
...
     Promise.resolve((await lpTokenERC20.symbol()).slice(4)),
-    axios.get<{
-      estimatedApy: string;
-      totalPoolSize: BigNumberish;
-      exchangeRateCurrent: string;
-    }>(`/api/pools`, {
+    axios.get<PoolQueryData>(`/api/pools`, {
       params: { token: tokenAddress },
     }),
+    Promise.resolve(parseEtherLike("0.1")), // TODO: replace with coingecko
```
pxrl, line 193: You should be able to query the price via the Across API /coingecko endpoint:

https://across.to/api/coingecko?l1Token=0x44108f0223a3c3028f5fe7aec7f9bb2e66bef82f&baseCurrency=usd

It looks like the price is currently pegged at $10 :astonished:...but we can control it via env vars in Vercel and will need to tune it to something less ambitious. This will probably serve us until we get third-party feeds via CoinGecko (The Across API will automagically source from CoinGecko once we remove the fixed price).

resolveVercelEndpoint() might also be worth a look, but I'm not sure if it can distinguish between görli and mainnet deployments: https://github.com/across-protocol/frontend-v2/blob/48f820191ee0d578765c0aa00804b0dfc0187d57/api/_utils.ts#L112

dohaki: Awesome thanks for the hint! I changed it [here](https://github.com/across-protocol/frontend-v2/pull/444/commits/7b09013793df5bb62d8b6896cba58c12f3012750) I added the respective env var to the Goerli Vercel project with the ACX token address on Goerli.

PR #479, opened 2022-11-21
improve(limits): Reduce unnecessary getAddress() calls

pxrl:
Per the existing implementation, for DAI, UMA, BAL (and future ACX) , we've called getAddress()
4 times on the same token address.

The new code unfortunately got a bit squashed by the formatter...it'd be a lot more concise if
the lpCushions object declaration/assignment could be handled on one line per key/value.

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -158,47 +158,34 @@ const handler = async (
       multicallOutput[1]
     );
 
-    switch (ethers.utils.getAddress(l1Token)) {
-      case ethers.utils.getAddress(
-        "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
-      ):
-        // Add WETH cushion to LP liquidity.
-        liquidReserves = liquidReserves.sub(
-          ethers.utils.parseEther(REACT_APP_WETH_LP_CUSHION || "0")
-        );
-        break;
+    // Subtract by any env-defined cushion amount.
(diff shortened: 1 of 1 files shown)


api/limits.ts
```diff
@@ -158,41 +158,47 @@ const handler = async (
...
+        break;
+
+      case ethers.utils.getAddress(
+        "0x6B175474E89094C44Da98b954EedeAC495271d0F"
+      ):
+        // Add DAI cushion to LP liquidity.
+        liquidReserves = liquidReserves.sub(
+          ethers.utils.parseUnits(REACT_APP_DAI_LP_CUSHION || "0", 18)
+        );
+        break;
```
pxrl, line 196: FWIW I think I will replace this with an Object that maps address => LP cushion and handle it in about 1/3 of the lines.

api/limits.ts
```diff
@@ -158,41 +158,47 @@ const handler = async (
...
 
-    if (liquidReserves.lt(0)) liquidReserves = ethers.BigNumber.from(0);
+      default:
+        liquidReserves = ethers.BigNumber.from(0);
+        break;
```
pxrl, line 200: This was wrong anyway, since liquidReserves should only be overridden conditionally.

nicholaspai (approved):

pxrl:
FWIW I've compared this test deployment vs. the current prod for WETH, USDC, WBTC, DAI and UMA. The maxDeposit field in the response has been identical in each case.

PR #492, opened 2022-11-26
improve(limits): Dynamically source LP cushion

pxrl:
A follow-up commit needs to map symbol, address and decimals, so it makes sense to factor it out.

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -45,13 +45,6 @@ const handler = async (
       REACT_APP_PUBLIC_INFURA_ID,
       REACT_APP_FULL_RELAYERS, // These are relayers running a full auto-rebalancing strategy.
       REACT_APP_TRANSFER_RESTRICTED_RELAYERS, // These are relayers whose funds stay put.
-      REACT_APP_USDC_LP_CUSHION,
-      REACT_APP_WETH_LP_CUSHION,
-      REACT_APP_DAI_LP_CUSHION,
-      REACT_APP_WBTC_LP_CUSHION,
-      REACT_APP_BAL_LP_CUSHION,
-      REACT_APP_UMA_LP_CUSHION,
-      REACT_APP_BOBA_LP_CUSHION,
     } = process.env;
     const providerUrl = `https://mainnet.infura.io/v3/${REACT_APP_PUBLIC_INFURA_ID}`;
     const provider = new ethers.providers.StaticJsonRpcProvider(providerUrl);
@@ -86,6 +79,12 @@ const handler = async (
       token,
       originChainId
     );
+
+    const symbol = Object.keys(l1Tokens).find(
+      (symbol) => l1Tokens[symbol].address === l1Token
+    );
+    if (symbol === undefined)
+      throw new InputError(`Unsupported token address: ${token}`);
 
     const [tokenDetailsResult, routeEnabledResult] = await Promise.allSettled([
       getTokenDetails(provider, l1Token, undefined, destinationChainId),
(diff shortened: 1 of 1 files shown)


api/limits.ts
```diff
@@ -21,6 +21,17 @@ import {
...
+const l1Tokens: { [symbol: string]: { address: string; decimals: number } } = {
+  DAI: { address: "0x6B175474E89094C44Da98b954EedeAC495271d0F", decimals: 18 },
+  MATIC: {
+    address: "0x7d1afa7b718fb893db30a3abc0cfc608aacfebb0",
+    decimals: 18,
+  },
+  USDC: { address: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48", decimals: 6 },
+  WBTC: { address: "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599", decimals: 8 },
+  WETH: { address: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2", decimals: 18 },
+};
```
pxrl, line 33: FWIW we should really just define this centrally in the SDK so we don't have shadow copies popping up everywhere.

nicholaspai: Can we re-use this from the `src/` folder somehow? https://github.com/across-protocol/frontend-v2/blob/master/src/utils/constants.ts#L297

nicholaspai (commented):
What is the benefit of this PR?

pxrl:
> What is the benefit of this PR?

It's more obvious in the follow-up commit (https://github.com/across-protocol/frontend-v2/pull/493), where I also need a mapping of symbols to token addresses and decimals. But I also think this is an improvement in terms of "code space" - i.e. it takes up much less space to do the same thing, and it requires less copy/paste to add new LP cushions in future. This would therefore reduce the chance of introducing errors under pressure in the event that an LP cushion needs to be added quickly.

api/limits.ts
```diff
@@ -158,25 +169,23 @@ const handler = async (
...
+
+    // Subtract any env-defined cushion amount.
+    const lpCushions: { [address: string]: [string | undefined, number] } =
+      Object.fromEntries(
+        Object.entries(_lpCushions).map(([symbol, lpCushion]) => {
+          const { address, decimals } = l1Tokens[symbol];
+          return [address, [lpCushion, decimals]];
+        })
+      );
+
```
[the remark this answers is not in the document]
pxrl: This is a good point. FWIW I think the additional lookups are worth it given the reduction in code, and the fact that it's pretty insignificant compared to the overall execution time for this endpoint (we spend most time waiting).

If we really wanted to optimise it, we could actually run the object generation in parallel with the external lookups - then it'd be "free" in terms of time. However, I re-evaluated the entire concept and settled on something that is (hopefully) simpler. I've updated the PR now so that we infer the token symbol and lookup the var directly by generating the env var name. This results in much less code, and even less work to add new cushions.

wdyt?

pxrl:
BTW we absolutely need to consolidate these hardcoded token details somewhere else. I'd prefer to do that in a follow-up commit, since that work will likely end up touching many files (i.e. other places where we also hardcode token details internally within individual files).

api/limits.ts
```diff
@@ -21,6 +21,22 @@ import {
   handleErrorCondition,
 } from "./_utils";
 
+// Note: Addresses must be checksummed.
```
nicholaspai, line 24: I really don't like this reminder, can't we perform some sort of checksummjng check on this? Also again, we should make this constant importable by other modules and probably be placed in a constants or utils file

nicholaspai (commented):
One refactor suggestion

api/limits.ts
pxrl: FWIW I think it's preferable to require that the addresses are checksummed, since the alternative is that we must dynamically checksum them before performing any equality check. That means we iterate over all 40 hex characters to get the case right before then proceeding with checking for equality, which will only be right 1/n times.

The alternative is perhaps to do it the tokenList() way, where the object is defined using getAddress(). It still seems wasteful for something that we can define statically. Once we have these definitions centralised then it will be much less overload to make sure that the central definition is correct...

As for the comment about centralisation, I totally agree and have responded with that in a few places. I don't think that work belongs in this commit, though, so I propose to implement it as a follow-up.

api/limits.ts
nicholaspai: How much work would it be to implement that refactor right now? I think having multiple lists of tokens is bad, also if the list is not a useful format, I would recommend making common a useful util function `getTokenDetails(symbol)`

api/limits.ts
nicholaspai: Ok Ok I relent I'm ok with this assuming there is follow up work to refactor and we don't forget this for a few months

api/limits.ts
```diff
@@ -161,45 +176,11 @@ const handler = async (
...
-    };
-    const [lpCushion, decimals] = lpCushions[
-      ethers.utils.getAddress(l1Token)
-    ] ?? [0, 18];
-    if (lpCushion && decimals)
-      liquidReserves = liquidReserves.sub(
-        ethers.utils.parseUnits(lpCushion, decimals)
-      );
-
+    const lpCushion = ethers.utils.parseUnits(
```
mrice32, line 179: +1 way simpler.

mrice32 (approved):
LGTM!

PR #493, opened 2022-11-26
feat(limits): Support minimum deposit floor

pxrl:
Requested by a partner, this will allow a lower bound for the minimum
deposit to be specified in the environment via the
REACT_APP_MIN_DEPOSIT_USD env var. This env var is specified as a JSON
object, mapping chain ID to the floor amount in USD.
    
An initial proposed REACT_APP_MIN_DEPOSIT_USD config would be:
    
  `'{ "1": 10, "10": 5, "137": 2, "288": 5, "42161": 5 }'`

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -209,7 +209,7 @@ const handler = async (
       : ethers.utils
           .parseUnits(
             (minDeposits[destinationChainId] ?? 0).toString(),
-            decimals
+            l1Tokens[symbol].decimals
           )
           .mul(ethers.utils.parseUnits("1"))
           .div(tokenPriceUsd);
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -45,13 +45,6 @@ const handler = async (
       REACT_APP_PUBLIC_INFURA_ID,
       REACT_APP_FULL_RELAYERS, // These are relayers running a full auto-rebalancing strategy.
       REACT_APP_TRANSFER_RESTRICTED_RELAYERS, // These are relayers whose funds stay put.
-      REACT_APP_USDC_LP_CUSHION,
-      REACT_APP_WETH_LP_CUSHION,
-      REACT_APP_DAI_LP_CUSHION,
-      REACT_APP_WBTC_LP_CUSHION,
-      REACT_APP_BAL_LP_CUSHION,
-      REACT_APP_UMA_LP_CUSHION,
-      REACT_APP_BOBA_LP_CUSHION,
     } = process.env;
     const providerUrl = `https://mainnet.infura.io/v3/${REACT_APP_PUBLIC_INFURA_ID}`;
     const provider = new ethers.providers.StaticJsonRpcProvider(providerUrl);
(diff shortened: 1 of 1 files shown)


pxrl:
Note that this change builds on top of this PR: https://github.com/across-protocol/frontend-v2/pull/492

api/limits.ts
```diff
@@ -21,6 +21,17 @@ import {
   handleErrorCondition,
 } from "./_utils";
 
+const l1Tokens: { [symbol: string]: { address: string; decimals: number } } = {
```
nicholaspai, line 24: Is there a way to reuse the tokenList() from src/utile/constants.ts instead of defining these addresses here?

nicholaspai (commented):
Will review later just left one drive by comment!

nicholaspai (approved):

api/limits.ts
pxrl: Not directly; tokenList is an array of objects, so it's not very convenient to use.

I think the right solution is to make a separate object where we can either lookup token details based on symbol or contract address, and then use that to populate tokenList.

I'd prefer to do that in a follow-up commit, since it's not really related to the min deposit floor work, and it will end up touching a few different unrelated files.

kevinuma: +1 on using the token info from constants.ts

api/limits.ts
```diff
@@ -73,6 +87,12 @@ const handler = async (
       originChainId
     );
 
+    const symbol = Object.keys(l1Tokens).find(
```
kevinuma, line 90: Why not just use a map with token address as the key? address => token details. This is the case if we use the token map from constants.ts

api/limits.ts
```diff
@@ -161,60 +187,40 @@ const handler = async (
...
     const maxGasFee = ethers.utils
       .parseEther(maxRelayFeePct.toString())
       .sub(relayerFeeDetails.capitalFeePercent);
 
+    // Ensure a maximum ratio of gas fee / deposit amount, then apply a floor afterwards.
+    const minDeposit = ethers.BigNumber.from(relayerFeeDetails.gasFeeTotal)
+      .mul(ethers.utils.parseEther("1"))
+      .div(maxGasFee);
+
+    // Normalise the environment-set USD minimum to units of the token being bridged.
```
kevinuma, line 206: I'm not sure if I follow here. minDepositFloor would be tokenPriceUsd if tokenPriceUsd <= 0? Why not just make it 0?

pxrl: Oops, that's definitely an error!

api/limits.ts
pxrl: Will circle back on that one so that we can get this one in now :)

api/limits.ts
```diff
@@ -187,15 +198,29 @@ const handler = async (
...
+    const minDepositFloor = tokenPriceUsd.lte(0)
```
kevinuma, line 207: Can token price usd be < 0 ever?

kevinuma (approved):

api/limits.ts
pxrl: Practically...probably not. But if we don't protect against it then we're exposed to whatever the price feed decides to give us. My preference is for being defensive when dealing with data that comes from external sources.

api/limits.ts
pxrl: Should be fixed now.

PR #510, opened 2022-11-29
fix: gas limit multiplier stake and unstake + env var 

dohaki:
This adds a buffer to the stake and unstake actions. This also adds an env var to control the gas limit multiplier.


pxrl (approved):
Tested OK for me - got 120k gas on prod and 250k gas against this preview.

amateima (approved):
Looks good to me by only looking at the code

PR #512, opened 2022-11-29
fix: footer links old layout

dohaki:
Footer links for routes `/pool` and `/bridge` could not be clicked.


amateima (approved):

pxrl:
@dohaki Is the FAQ link working correctly in this deployment? It just redirects to the main page for me.

PR #518, opened 2022-12-01
feat(vercel-api): support `timestamp` query param cg price

dohaki:
This allows specifying a `timestamp` query param
in the `GET /api/coingecko` handler to retrieve a historic price using the `getHistoricContractPrices` method of the sdk `CoingeckoClient`.

Note, that this is preliminary work in order to retrieve historic `suggested-fees` for a given timestamp.

### DIFF
--- a/api/_types/routes/coingecko-routes.types.ts
+++ b/api/_types/routes/coingecko-routes.types.ts
@@ -3,4 +3,5 @@ import { TypedVercelRequest } from "../generic.types";
 export type CoinGeckoInputRequest = TypedVercelRequest<{
   l1Token: string;
   baseCurrencySymbol: string;
+  timestamp?: string;
 }>;
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -398,12 +398,13 @@ export const getRelayerFeeDetails = (
  */
 export const getCachedTokenPrice = async (
   l1Token: string,
-  baseCurrency: string = "eth"
+  baseCurrency: string = "eth",
+  timestamp?: string
 ): Promise<number> => {
   return Number(
     (
       await axios(`${resolveVercelEndpoint()}/api/coingecko`, {
-        params: { l1Token, baseCurrency },
+        params: { l1Token, baseCurrency, timestamp },
       })
     ).data.price
   );
--- a/api/coingecko.ts
+++ b/api/coingecko.ts
@@ -1,5 +1,6 @@
 import { VercelResponse } from "@vercel/node";
 import { ethers } from "ethers";
+import { DateTime, Duration } from "luxon";
 import { isString } from "./_typeguards";
 import { CoinGeckoInputRequest } from "./_types";
 import { getLogger, InputError, handleErrorCondition } from "./_utils";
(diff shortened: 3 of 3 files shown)


pxrl:
Is it possible to get a high-level overview of the process for determining suggested-fees quotes that have been supplied by the API?

Looking up historical prices gives the impression that we will be reconstructing the fee, rather than saving it. That's more efficient, but I'm not sure how well it will work in when we change the relayer capital fee (i.e. currently at 1bps, but subject to change in the future). Do we also plan to have a way to look up those changes over time?

dohaki:
> Is it possible to get a high-level overview of the process for determining suggested-fees quotes that have been supplied by the API?
> 
> Looking up historical prices gives the impression that we will be reconstructing the fee, rather than saving it. That's more efficient, but I'm not sure how well it will work when we change the relayer capital fee (i.e. currently at 1bps, but subject to change in the future). Do we also plan to have a way to look up those changes over time?

Yes, like you said the current high-level plan was to reconstruct the historic fee via a `timestamp` param:
- **Vercel API:** extend `GET /api/suggested-fees`  to allow query param `?timestamp` to retrieve historic fees under the assumption that the only time-dependent variables are `gasPrice` and `tokenPrice`
  - Possibility to get historic `tokenPrice` is covered in this PR
  - Possibility to get historic `gasPrice` would be done (for now) by using an external API
- **Scraper API:** would then use the above `timestamp` param to retrieve the suggested fee when indexing a deposit from the Vercel API

Drawbacks would be:
- Additional external dependency
- Historic suggested fees are not deterministic, due to
  - Possible discrepancies in the used and historic `gasPrice` values
  - Possible changing relayer capital fee as you mentioned (didn't know that this would change in the future 🤔)
 
The first discrepancy, we wanted to mitigate by allowing a large enough error margin when comparing the current relayer fee and the historic suggested fees. But not sure if this would also cover the changing of the capital fee...

We also discussed the approach to basically record the suggested fees in the Scraper API, every time the Vercel API endpoint is hit. But didn't want to increase the load of the Scraper too much.

pxrl:
> We also discussed the approach to basically record the suggested fees in the Scraper API, every time the Vercel API endpoint is hit. But didn't want to increase the load of the Scraper too much.

I agree that it seems heavy to record the quotes supplied, and that reconstructing an approximation of the fees is better if we can make it work. Recording averages might be an improvement, but that won't help much when there is short-term volatility.

> The first discrepancy, we wanted to mitigate by allowing a large enough error margin when comparing the current relayer fee and the historic suggested fees. But not sure if this would also cover the changing of the capital fee...

One thing we _could_ consider is pushing a nominal relayer capital fee to the ConfigStore contract. This would allow current and historical values to be queried dynamically by all dependent parties without introducing additional centralisation. It does come at the cost of some marginal overheads each time we want to update it, though...and any additions to the ConfigStore should be thoroughly justified. @nicholaspai @mrice32 any thoughts on that?

nicholaspai:
> > We also discussed the approach to basically record the suggested fees in the Scraper API, every time the Vercel API endpoint is hit. But didn't want to increase the load of the Scraper too much.
> 
> I agree that it seems heavy to record the quotes supplied, and that reconstructing an approximation of the fees is better if we can make it work. Recording averages might be an improvement, but that won't help much when there is short-term volatility.
> 
> > The first discrepancy, we wanted to mitigate by allowing a large enough error margin when comparing the current relayer fee and the historic suggested fees. But not sure if this would also cover the changing of the capital fee...
> 
> One thing we _could_ consider is pushing a nominal relayer capital fee to the ConfigStore contract. This would allow current and historical values to be queried dynamically by all dependent parties without introducing additional centralisation. It does come at the cost of some marginal overheads each time we want to update it, though...and any additions to the ConfigStore should be thoroughly justified. @nicholaspai @mrice32 any thoughts on that?

Just talked with @dohaki offline about this I agree there is no easy to way to query historical capital fee, although moving it to the ConfigStore would make the most sense.

The original motivation for this PR was for the Scraper-API to have access to the `/suggested-fees` returned for a deposit at the deposit's original timestamp. I think the better solution than this PR is to have the Scraper-API query the current `/suggested-fees` route as each deposit is queried for the first time and then save that result in the DB. This increases the Scraper-API's load but doesn't make us have to implement this `timestamp` query.

pxrl:
> The original motivation for this PR was for the Scraper-API to have access to the `/suggested-fees` returned for a deposit at the deposit's original timestamp. I think the better solution than this PR is to have the Scraper-API query the current `/suggested-fees` route as each deposit is queried for the first time and then save that result in the DB. This increases the Scraper-API's load but doesn't make us have to implement this `timestamp` query.

Any thoughts on how to handle the additional load from the proposed approach?

I think it's safe to expect that the suggestes-fees endpoint gets hit several times per Deposit that is actually placed. For example, aggregators will make this request very regularly, and there will often be at least 2 - 3 requests before a user proceeds with a Deposit (assuming they even choose Across). It seems like we might therefore record a _lot_ more quotes than we actually need.

nicholaspai:
@dohaki is this PR needed anymore?

dohaki:
@nicholaspai Closing this PR as we chose a different approach

PR #527, opened 2022-12-12
improve(api): Standardize relayer fee calculations from sdk-v2

nicholaspai:
- [This PR](https://github.com/across-protocol/sdk-v2/pull/145) improves SDK-v2 so that it exports `isAmountTooLow` and `minDeposit` which means that /limits and /suggested-fees can compute those values using the same math
- Independently, we should also set the capital costs percent from an env var in the same way that the FE does

Resolves ACX-480


api/_utils.ts
```diff
@@ -38,6 +39,10 @@ export const DEFAULT_GAS_MARKUP = 0;
 
 export const HUP_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID || 1);
 
+export const FLAT_RELAY_CAPITAL_FEE = REACT_APP_FLAT_RELAY_CAPITAL_FEE
+  ? Number(process.env.REACT_APP_FLAT_RELAY_CAPITAL_FEE)
+  : 0;
```
pxrl, line 44:
```suggestion
export const FLAT_RELAY_CAPITAL_FEE = Number(REACT_APP_FLAT_RELAY_CAPITAL_FEE ?? 0);
```

Is there a valid case for ever having a 0% relayer capital fee? How about we stick to something like 3bps and then override via REACT_APP_FLAT_RELAY_CAPITAL_FEE?

api/_utils.ts
```diff
@@ -360,10 +365,14 @@ export const getRelayerFeeCalculator = (destinationChainId: number) => {
...
   const relayerFeeCalculatorConfig = {
     feeLimitPercent: maxRelayFeePct * 100,
-    capitalCostsPercent: 0.04,
+    capitalCostsPercent: FLAT_RELAY_CAPITAL_FEE, // This is set same way in ./src/utils/bridge.ts
     queries: queryFn(),
     capitalCostsConfig: relayerFeeCapitalCostConfig,
   };
+  if (relayerFeeCalculatorConfig.feeLimitPercent < 1)
+    throw new Error(
+      "Setting fee limit % too low will produce nonsensical relay fee details"
```
pxrl, line 374: Consider specifying the actual limit as part of the exception?

api/_utils.ts
nicholaspai: i like that

pxrl (approved):
No major comments; only some suggestions around nits, for consideration.

pxrl (approved):
LGTM!

pxrl (approved):

api/_utils.ts
```diff
@@ -37,11 +36,11 @@ export const gasMarkup = GAS_MARKUP ? JSON.parse(GAS_MARKUP) : {};
 // Default to no markup.
 export const DEFAULT_GAS_MARKUP = 0;
 
-export const HUP_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID || 1);
+export const HUP_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID ?? 1);
```
pxrl, line 39: One observation here - this change permits a chain ID of 0:

```
> 0 || 1
1
> 0 ?? 1
0
```

nicholaspai: ah i see ill revert

api/_utils.ts
```diff
@@ -36,8 +36,14 @@ export const gasMarkup = GAS_MARKUP ? JSON.parse(GAS_MARKUP) : {};
 // Default to no markup.
 export const DEFAULT_GAS_MARKUP = 0;
 
+// Don't permit REACT_APP_FLAT_RELAY_CAPITAL_FEE=0
 export const HUP_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID || 1);
```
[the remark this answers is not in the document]
nicholaspai: no its ok to have capital fee be 0 but 0 as a default doesn't make the most sense

PR #542, opened 2022-12-17
refactor: refine ampli config

dohaki:
Some small adjustments to the amplitude config.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,12 +2,18 @@ import { BrowserRouter as Router } from "react-router-dom";
 import { Types } from "@amplitude/analytics-browser";
 import { ampli } from "./ampli";
 import Routes from "./Routes";
-import { amplitudeEnvironment, isProductionBuild } from "utils";
+import {
+  amplitudeEnvironment,
+  amplitudeAPIKey,
+  isProductionBuild,
+} from "utils";
 
 // Record an event when the application is loaded
 ampli.load({
   environment: amplitudeEnvironment,
+  disabled: !Boolean(amplitudeAPIKey),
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
     configuration: {
       disableCookies: true,
+      logLevel: isProductionBuild ? Types.LogLevel.Error : Types.LogLevel.Debug,
+      trackingOptions: {
+        ipAddress: false,
+        carrier: false,
+        city: false,
+        region: false,
+        dma: false, // designated market area
+      },
```
mrice32, line 26: +1

src/App.tsx
```diff
@@ -1,15 +1,29 @@
...
+  isProductionBuild,
+} from "utils";
 
 // Record an event when the application is loaded
 ampli.load({
   environment: amplitudeEnvironment,
+  disabled: !Boolean(amplitudeAPIKey),
   client: {
-    apiKey: amplitudeAPIKey || "",
+    apiKey: amplitudeAPIKey,
```
mrice32, line 16: Does the type matter here? I figured we would have to ensure this is a string before putting it here, but maybe not!

mrice32 (approved):
LGTM!

amateima (approved):

src/App.tsx
dohaki: It shouldn't matter much according to the code here https://github.com/across-protocol/frontend-v2/blob/507eb56695fec2e4d942cc71a3cd85069aa25774/src/ampli/index.ts#L1067-L1074

PR #546, opened 2022-12-22
feat: set user_id to wallet address

dohaki:
Sets the user_id to the wallet address.


amateima (approved):

dohaki:
> LGTM! Will this change need to be reflected in other `identify` calls to amplitude?

Yes, no additional changes needed.

PR #566, opened 2023-01-06
refactor: remove legacy pool code + redux deps

dohaki:
The old liquidity pool page was the only place where we used redux. With the new lp page, we don't need it anymore

Closes ACX-562


amateima (commented):
Are we ditching Redux entirely on the FE app?

dohaki:
> Are we ditching Redux entirely on the FE app?

Yes, there is no need to keep it anymore. Or do you have anything against this?

PR #574, opened 2023-01-10
fix(amplitude): update ampli and replace `TransferTransactionConfirmed` with `DepositTransactionConfirmed` event

dohaki:
Fixes ACX-583


amateima (commented):
I saw that `components/SendAction/useSendAction.ts` is removed in #578 PR

dohaki:
> I saw that `components/SendAction/useSendAction.ts` is removed in #578 PR

Yes, we need to re-add some amplitude tracking for the new bridge and pool page PR. But this should be part of the respective epic branch

PR #593, opened 2023-01-16
refactor: use superstruct in Vercel API for input validation

dohaki:
Closes ACX-498

Note, that this PR does not include a new dependency (e.g. zod) and instead uses the already existing validation lib superstruct because it is already used on the FE. If we decide to switch over to zod then we also need to adapt the validation on the FE.


api/_utils.ts
```diff
@@ -629,3 +630,34 @@ export function handleErrorCondition(
   }
   return response.status(status).send(error.message);
 }
+
+/* ------------------------- superstruct validators ------------------------- */
```
daywiss, line 634: ive never written these before, pretty cool

daywiss (approved):
nice

PR #607, opened 2023-01-19
fix: change ERC-20 to Ethereum wording

dohaki:


dohaki:
> Good point - this was how it was in Figma. Maybe we could look into something more general like Web3 wallet addresses?

Tim changed it to Ethereum after I pinged him in Figma

PR #609, opened 2023-01-19
fix: explicitly pass required chain to waitOnTx helper

dohaki:
For some reason, `txResponse.chainId` returns always `0`. So the `waitOnTransaction` helper falls back to `tx.wait` event for supported chains.

### DIFF
--- a/src/utils/notify.ts
+++ b/src/utils/notify.ts
@@ -50,12 +50,13 @@ export const notificationEmitter = async (
  * @param ignoreErrors An optional parameter to ignore tx failure and return successful
  **/
 export const waitOnTransaction = async (
+  requiredChainId: number,
   tx: ContractTransaction,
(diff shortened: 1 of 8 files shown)


src/utils/notify.ts
```diff
@@ -50,12 +50,13 @@ export const notificationEmitter = async (
  * @param ignoreErrors An optional parameter to ignore tx failure and return successful
  **/
 export const waitOnTransaction = async (
+  requiredChainId: number,
   tx: ContractTransaction,
   notify: NotifyAPI,
   timingBuffer?: number,
   ignoreErrors?: boolean
 ): Promise<void> => {
-  if (supportedNotifyChainIds.includes(tx.chainId)) {
```
dohaki, line 58: Somehow `tx.chainId` is always `0` 🤔

PR #617, opened 2023-01-24
fix: display wrong network banner on pool page

dohaki:
Displays super banner if incorrect network also on pool page.


daywiss:
this is still showing i can add liquidity on wrong network, though the button does send change network request to MM
image

dohaki:
> this is still showing i can add/remove liquidity on wrong network, though the button does send change network request to MM image

Yea this is intentional. Because the user doesn't necessarily need to scroll up to see the banner in order to know why the button is disabled.

daywiss (approved):
approving, you can fix button if needed in new pr

PR #655, opened 2023-02-13
feat: add sentry and error boundary

dohaki:
Closes ACX-564

This integrates Sentry to monitor errors. We also add an error boundary component that looks like this
image
in case of a fatal app crash.


amateima (approved):
Looks good 👍 I am not familiar with the design proposal, but does it make sense to display the website header too?

dohaki:
> Looks good 👍 I am not familiar with the design proposal, but does it make sense to display the website header too?

Yea, I initially wanted to do so. But the header has actually quite some logic (wallet button, sidebar, etc.) and I wanted to make the error boundary contain minimal possible error sources. Maybe I can wrap the header in a separate error boundary

PR #657, opened 2023-02-14
feat: include greyed-out selectors for all routes/tokens

james-a-morris:
This PR allows all chains/assets to be visible to the user when a user clicks a dropdown. However, the selectors now `disable` inconsistent/unavailable route combinations. 
image


dohaki (approved):
Nice job 👍 LGTM

PR #665, opened 2023-02-17
fix: retry lazy load on `ChunkLoadError`

dohaki:
Fixes ACX-851

Reload app on `ChunkLoadError`

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, lazy, Suspense } from "react";
+import { useState, useEffect, Suspense } from "react";
 import { Switch, Route, useLocation, useHistory } from "react-router-dom";
 import { Header, SuperHeader, Banner, Sidebar } from "components";
 import { useConnection } from "hooks";
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -13,6 +13,7 @@ import {
   stringValueInArray,
   getConfig,
 } from "utils";
+import lazyWithRetry from "utils/lazy-with-retry";
 import { ReactComponent as InfoLogo } from "assets/icons/info-24.svg";
 import Toast from "components/Toast";
 import BouncingDotsLoader from "components/BouncingDotsLoader";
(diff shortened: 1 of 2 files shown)


james-a-morris (commented):
Looks good! 

Question: should we be introducing stored state? Additionally, if the user experiences this issue in viewing session `T` and a storage element is created. If the user has another issue during viewing session `T+n` will the logic already have the `HAS_REFRESHED_KEY` set to true?

dohaki:
> Looks good!
> 
> Question: should we be introducing stored state? Additionally, if the user experiences this issue in viewing session `T` and a storage element is created. If the user has another issue during viewing session `T+n` will the logic already have the `HAS_REFRESHED_KEY` set to true?

> should we be introducing stored state? 

Hmm, you mean persisting app state on reload? I think you can achieve something like this react-query. 

> Additionally, if the user experiences this issue in viewing session `T` and a storage element is created. If the user has another issue during viewing session `T+n` will the logic already have the `HAS_REFRESHED_KEY` set to true?

We guarantee that if the chunk correctly loads, that the `HAS_REFRESHED_KEY` key is set to `false` here https://github.com/across-protocol/frontend-v2/pull/665/files#diff-b3d5dbd362214b55f13e3f893067e50783e9d8e4b2598622a3364f1b88711c85R17.

james-a-morris:
> > Looks good!
> > Question: should we be introducing stored state? Additionally, if the user experiences this issue in viewing session `T` and a storage element is created. If the user has another issue during viewing session `T+n` will the logic already have the `HAS_REFRESHED_KEY` set to true?
> 
> > should we be introducing stored state?
> 
> Hmm, you mean persisting app state on reload? I think you can achieve something like this react-query.
> 
> > Additionally, if the user experiences this issue in viewing session `T` and a storage element is created. If the user has another issue during viewing session `T+n` will the logic already have the `HAS_REFRESHED_KEY` set to true?
> 
> We guarantee that if the chunk correctly loads, that the `HAS_REFRESHED_KEY` key is set to `false` here https://github.com/across-protocol/frontend-v2/pull/665/files#diff-b3d5dbd362214b55f13e3f893067e50783e9d8e4b2598622a3364f1b88711c85R17.

Np! Thanks for clarifying.

james-a-morris (approved):

PR #666, opened 2023-02-20
fix: remove edge case

james-a-morris:
This PR removes the edge case outlined in ACX-852. Additionally, this PR includes a reset of the `finishedDate` state when a user begins to attempt another bridge request in one session.


dohaki (approved):
LGTM

PR #669, opened 2023-02-22
improve: properly set AllAssetsAllNetworksWalletCurrentBalanceUsd

james-a-morris:
## Motivation
During regular FE execution, the `AllAssetsAllNetworksWalletCurrentBalanceUsd` user property was not being correctly set. This error was caused by commit [`3c21999`](https://github.com/across-protocol/frontend-v2/commit/3c2199935fed0ea9f2d13d61a3aef0c7b21ca9fe), which removed the environment attribute for amplitude, leading to the incorrect setting of the user property.

Additionally, this PR also contains logic to retry the `AllAssetsAllNetworksWalletCurrentBalanceUsd` instrumentation every 30 seconds during a user's session until uploaded.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,6 @@ if (amplitudeAPIKey) {
   });
   ampli.load({
     client: { instance },
-    environment: isProductionBuild ? "production" : "development",
   });
 }
 // Record an event when the application is loaded
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,6 +30,7 @@ if (amplitudeAPIKey) {
   });
   ampli.load({
     client: { instance },
+    environment: isProductionBuild ? "production" : "development",
   });
 }
 // Record an event when the application is loaded
--- a/src/components/WalletBalanceTrace/useWalletBalanceTrace.ts
+++ b/src/components/WalletBalanceTrace/useWalletBalanceTrace.ts
@@ -1,6 +1,6 @@
 import { BigNumber, utils } from "ethers";
 import { useConnection } from "hooks";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import {
   ChainId,
   fixedPointAdjustment,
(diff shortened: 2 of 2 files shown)


mrice32 (approved):
LGTM!

src/App.tsx
```diff
@@ -30,6 +30,7 @@ if (amplitudeAPIKey) {
   });
   ampli.load({
     client: { instance },
+    environment: isProductionBuild ? "production" : "development",
```
dohaki, line 33: Hmm, I don't think that you need to pass that explicitly. Amplitude can derive the environment from the used API key AFAICT. Take a look at https://github.com/across-protocol/frontend-v2/blob/334011db6488b93a675bdc8efb4d0ddd5fa4a191/src/ampli/index.ts#L1295-L1309. The only time it uses the `environment` param, is when we don't pass an API key. That is the reason why I initially removed this from the config.

When you also take a look at the `this.amplitude.init` method, it doesn't use any `environment` value.

dohaki (changes requested):
I don't think that this fixes the issue tbh. The `environment` option is only used to determine an API key if we don't pass it explicitly.

src/components/WalletBalanceTrace/useWalletBalanceTrace.ts
```diff
@@ -17,96 +17,112 @@ import Sentry from "utils/sentry";
...
+                reportTotalWalletUsdBalance(balance);
+                setRecordingSuccessful(true);
+              })
+              .catch((e) => {
+                console.error("Failed to fetch balances:", e);
+                Sentry.captureException(e);
+              });
+          }, 30000)
+        : undefined;
+    return () => clearInterval(interval);
```
james-a-morris, line 38: The retry logic is performed here. This PR also includes a reformat to align with the eslint config.

james-a-morris:
@dohaki I've added additional logic to proactively retry uploading the user property during a session until complete.

src/components/WalletBalanceTrace/useWalletBalanceTrace.ts
dohaki: Can we add some max. retries here? I don't think that retrying indefinitely is a good practice. Alternatively, we could put that also into a `useQuery` hook where all the retry, etc. is already implemented.

dohaki (commented):
I couldn't verify that specifying the `environment` key actually fixes the issues, could you? If it doesn't, we should remove it from this PR IMO.

james-a-morris:
> I couldn't verify that specifying the `environment` key fixes the issues; could you? If it doesn't, we should remove it from this PR, IMO.

Although I believe that, in this case, explicitly defining the production scope of Amplitude is a benefit to the codebase, I have removed it in this commit. 🙂  The primary verification I performed was that without setting this value, the preview branch was not correctly uploading to the correct dataset. I can add the environment key back if the problem is not resolved by adding this retry logic.

Good callout with the `react-query` 🔥

src/components/WalletBalanceTrace/useWalletBalanceTrace.ts
```diff
@@ -14,99 +13,113 @@ import {
...
+      if (!account) return;
+      try {
+        const balance = await calculateUsdBalances(account);
+        reportTotalWalletUsdBalance(balance);
+        return balance;
+      } catch (e) {
+        console.error("Failed to fetch balances:", e);
+        Sentry.captureException(e);
+        throw e; // Throw the error so that react-query can retry
+      }
```
dohaki, line 33: NIT: We can actually remove the try/catch because we already log react-query failures to Sentry.

dohaki (approved):
Ok, nice. Yea weird when I tested it was correctly updating the user prop on Amplitude without specifying the environment key 🤔 Let's see. 

Looks good 👍 One last nit would be to remove the try/catch.

PR #673, opened 2023-03-07
feat: Add logs for request and response data

nicholaspai:
Our existing log providers (datadog, vercel) don't log request and response data so this uses existing logger module to do that


mrice32 (approved):

james-a-morris (approved):
Good call - it makes more sense that mainly tracking route configurations

PR #674, opened 2023-03-07
improve: Try to log errors better

nicholaspai:


api/_utils.ts
```diff
@@ -636,14 +636,23 @@ export function handleErrorCondition(
...
+    console.error("Error could not be defined.", error);
     return response.status(500).send("Error could not be defined.");
   }
   let status: number;
   if (error instanceof InputError || error instanceof StructError) {
-    logger.warn({ at: endpoint, message: "400 input error", error });
+    logger.warn({
+      at: endpoint,
+      message: "400 input error",
+      error: error.message,
```
james-a-morris, line 647: Nit: should we place the error inside the message itself?

james-a-morris (approved):
LGTM! This is a straightforward addition.

I left a nit if you'd like to consider it. Otherwise, it's good to me 🙂

PR #678, opened 2023-03-11
fix: add insufficient liquidity alert

mrice32:
This adds an error message to the button disable in #677.


amateima (approved):

james-a-morris:
Good call

PR #680, opened 2023-03-13
feat: :sparkles: add per-route lp cushions

james-a-morris:
This change adds per-route LP cushions. Additionally, this PR simplifies the logic in how we resolve, compute, and act on any given LP cushion.

Breaking Changes: The ENV variables will need to be modified to include a single `REACT_APP_LP_CUSHION_MAP` variable which replaces all other LP Cushion variables.

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -83,6 +83,25 @@ export const relayerFeeCapitalCostConfig: {
   )
 );
 
+/**
+ * The cushion applied to the total available liquidity when calculating the max amount of a token to relay.
+ * This is to account for slippage and other factors that may cause the relay to fail.
+ * This map takes the following format:
+ * {
+ *    "DAI": "100000",
+ *    "DAI:1:10": "100001",
+ * }
+ * The key is the token symbol, and the value is the cushion in wei. If the key is a token symbol followed by a colon,
+ * followed by an origin chain ID, followed by a colon, followed by a destination chain ID, then the cushion will only
+ * apply to that specific route. For example, the key "DAI:1:10" will only apply to the DAI route from mainnet to
+ * optimism.
+ */
+export const lpCushionMap: {
+  [symbol: string]: string;
+} = process.env.REACT_APP_LP_CUSHION_MAP
+  ? JSON.parse(process.env.REACT_APP_LP_CUSHION_MAP)
+  : {};
+
 // If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
 export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = 12 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
 
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -12,7 +12,11 @@ import { define, StructError } from "superstruct";
 import enabledMainnetRoutesAsJson from "../src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json";
 import enabledGoerliRoutesAsJson from "../src/data/routes_5_0xA44A832B994f796452e4FaF191a041F791AD8A0A.json";
 
-import { maxRelayFeePct, relayerFeeCapitalCostConfig } from "./_constants";
+import {
+  lpCushionMap,
+  maxRelayFeePct,
+  relayerFeeCapitalCostConfig,
+} from "./_constants";
 import { StaticJsonRpcProvider } from "@ethersproject/providers";
 import QueryBase from "@across-protocol/sdk-v2/dist/relayFeeCalculator/chain-queries/baseQuery";
 import { VercelResponse } from "@vercel/node";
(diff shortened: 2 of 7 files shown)


api/_utils.ts
```diff
@@ -668,3 +672,20 @@ export function boolStr() {
...
+ * @param fromChainId The origin chain ID
+ * @param toChainId The destination chain ID
+ * @returns The cushion in wei
+ */
+export function getLpCushion(
+  symbol: string,
+  fromChainId?: number,
+  toChainId?: number
+) {
+  const key = `${symbol}:${fromChainId}:${toChainId}`;
```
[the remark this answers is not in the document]
james-a-morris: This was just the first delimiter that came to mind. To add more thought, the `_` and `-` feel implicitly bi-directional. I also thought to use something like `->` but that felt like too complex in terms of key composition.

src/utils/bridge.ts
```diff
@@ -335,77 +336,19 @@ export default class LpFeeCalculator {
...
-    }
+    // Attempt to resolve the token from the token address.
+    const token = getTokenByAddress(tokenAddress);
+    // Resolve the LP cushion for the token. Determine if the token is a bridged token.
+    const lpCushion = getLpCushion(
+      token.symbol,
+      originChainId,
+      destinationChainId
+    );
+    console.log(lpCushion);
```
mrice32, line 347: Extra console log?

src/utils/bridge.ts
```diff
@@ -326,7 +325,9 @@ export default class LpFeeCalculator {
   }
   async isLiquidityInsufficient(
     tokenAddress: string,
-    amount: utils.BigNumberish
+    amount: utils.BigNumberish,
```
mrice32, line 328: I didn't realize we were computing this directly in the FE. Is there a reason we don't use /limits so we don't have to do this logic in both places?

src/utils/bridge.ts
```diff
@@ -326,7 +325,9 @@ export default class LpFeeCalculator {
...
+    originChainId?: number,
+    destinationChainId?: number
```
mrice32, line 330: I suspect that this function should have blocked the transfers, previously, but this logic didn't seem to be working, which is why [this fix](https://github.com/across-protocol/frontend-v2/commit/2017fd2c83bfd87978baf067b63b0df2b989dfb2) was added. Are we sure this function is currently being used to block transfers that are too large?

api/_constants.ts
```diff
@@ -83,6 +83,25 @@ export const relayerFeeCapitalCostConfig: {
...
+ * }
+ * The key is the token symbol, and the value is the cushion in wei. If the key is a token symbol followed by a colon,
+ * followed by an origin chain ID, followed by a colon, followed by a destination chain ID, then the cushion will only
+ * apply to that specific route. For example, the key "DAI:1:10" will only apply to the DAI route from mainnet to
+ * optimism.
+ */
+export const lpCushionMap: {
+  [symbol: string]: string;
+} = process.env.REACT_APP_LP_CUSHION_MAP
+  ? JSON.parse(process.env.REACT_APP_LP_CUSHION_MAP)
```
mrice32, line 102: I think these JSON envs are sometimes hard to configure (since you have to build the json and then format it all into a single string, which is hard to read in the vercel UI). Thoughts on breaking it up into separate override variables?

Something like (highest precedence first):
LP_CUSHION_DAI_1_10
LP_CUSHION_DAI_1
LP_CUSHION_DAI

Thoughts?

mrice32 (commented):
Looks great! Just a few minor questions/comments

src/utils/bridge.ts
james-a-morris: Yes - it is now removed. Thanks for the catch 🙂

src/utils/bridge.ts
james-a-morris: My understanding here is that the limits compute the true liquidity height, but the LP cushion is only a concept on the FE to pad the liquidity during bridges. An alternative here is that we could add an additional field in the `/limits` endpoint to create something like:
```
{
    "liquidity": ....,
    "preferred_liquidity": ....,
}
```

Where `preferred_liquidity` (or a more apt name) is essentially the padding that we'd prefer aggregators/the FE to use when giving quotes to the end user.

src/utils/bridge.ts
james-a-morris: This function primarily handles liquidity padding. The data that is used should be the logic that blocks high limits. My recommendation is that we create a separate PR that accounts for this. Alternatively, we can place it directly in this PR

mrice32 (commented):
@james-a-morris and I discussed this IRL and walked through three pre-existing issues with the code that we have the opportunity to fix in this PR :).
1. There is a validity check on the amount input field for the bridge that is using liquidReserves directly, so it is not considering the configurable cushion. This should be changed to consider maxDeposit from /limits.
2. This validity check does not produce an error message for insufficient liquidity, only for the case where the amount is too low.
3. The isLiquidityInsufficient method repeats some the logic in the API for computing the liquidReserves and subtracting the cushion from it. Rather than performing this logic in both places, we should just use the /limits API and remove some code!

@james-a-morris let me know if I missed anything!

api/_constants.ts
james-a-morris: This has been resolved

mrice32 (commented):
> The isLiquidityInsufficient method repeats some the logic in the API for computing the liquidReserves and subtracting the cushion from it. Rather than performing this logic in both places, we should just use the /limits API and remove some code!

@james-a-morris Did we intend to fix this? It looks like we still have repeated code in isLiquidityInsufficient (querying the HubPool directly rather than just using the maxDeposit value returned by /limits. Thoughts?

james-a-morris:
> > The isLiquidityInsufficient method repeats some the logic in the API for computing the liquidReserves and subtracting the cushion from it. Rather than performing this logic in both places, we should just use the /limits API and remove some code!
> 
> @james-a-morris Did we intend to modify this? It looks like we still have repeated code in isLiquidityInsufficient (querying the HubPool directly rather than just using the maxDeposit value returned by /limits. Thoughts?

My intuition is to put that in a separate PR as the logic for making on-chain requests is now only using the `/limits` API. Do you recommend we update this PR to include the refactor changes? 

Additionally, I've consulted with Tim and the new alerts are available for testing within this PR 🙂

mrice32 (approved):
LGTM!

mrice32 (approved):
Awesome, thank you for doing that. We've removed so much code in this PR -- really amazing work @james-a-morris!

PR #691, opened 2023-03-24
fix(ui): :ambulance: resolve autoconnect issue

james-a-morris:
This change corrects for the auto-connect functionality that was occasionally breaking the FE. This change adds a new option to the `customOnboardConnect` function to enable the autoconnect functionality. This is a correction from the previous logic because in production, currently the autoconnect tries to keep silentely failing causing some users to not be able to connect.


mrice32 (approved):
LGTM! I QA'd and it worked perfectly. No autoconnect when connecting with the button!

PR #693, opened 2023-03-27
fix: :art: set rounding to one decimal place

james-a-morris:
Sets the total fidelity of the volume on the splash page to use one decimal of rounding

Closes: ACX-998
Closes: ACX-994


dohaki (approved):
LGTM

PR #696, opened 2023-03-31
refactor(ampli): enforce initial user prop calls precedence

dohaki:
To some degree fixes ACX-955.

The problem with some charts in Amplitude is that user properties are always counted at the time of an event. So we need to make sure that we set relevant user properties before any event is triggered.

We can not fully enforce this because some user properties require a connected wallet. This PR therefore should minimize the `none` values for user properties on events that fire on auto-selects of wallets.


src/Routes.tsx
```diff
@@ -114,9 +112,7 @@ const Routes: React.FC = () => {
 
   return (
     <>
-      <RouteTrace />
-      <WalletTrace />
-      <WalletBalanceTrace />
+      <AmpliTrace />
```
james-a-morris, line 115: Good Refactor +1

dohaki:
> This LGTM - however, I wonder if we should buffer these events when the `areInitialUserPropsSet` is false. For example, if we have a route change event but the user props aren't loaded, should we add that to a queue of events to be pushed to amplitude for when `areInitialUserPropsSet` becomes true?

@james-a-morris Sorry for the late reply. Yes, you are right. I added a queuing mechanism that adds amplitude tracking requests to a queue and if the initial user props are set, it processes the tracking requests in the queue. I only added this approach to the page view and wallet events for now. But maybe it makes sense to wrap every tracking request with it?

james-a-morris (approved):
LGTM

PR #702, opened 2023-04-13
perf: improve tree-shaking typechain

dohaki:
While updating to `@across-protocol/contracts-v2@2.1.0-beta.1`, I had to circumvent some issues with the latest `@uma/common` package, which currently is not really compatible with a browser environment. The workaround is to import the typechain bindings via an explicit path. 

Then I noticed that our bundle size got smaller. It seems that the packages `@across-protocol/contracts-v2` and `@across-protocol/across-token` can not be tree-shaken if we import stuff directly from the main entry file. Until we fix this upstream, I decided to add a wrapper module `utils/typechain.ts` that re-exports the respective bindings. This results in a **300 kB (gzipped)** decrease in the bundle size.

### DIFF
--- a/src/hooks/useApprove.ts
+++ b/src/hooks/useApprove.ts
@@ -1,9 +1,9 @@
 import { useConnection } from "hooks";
 import { useMutation } from "react-query";
 import { BigNumberish } from "ethers";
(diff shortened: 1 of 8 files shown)


james-a-morris (approved):
This is a good PR - LGTM

PR #710, opened 2023-04-19
fix: fee query 400 code due to same origin + dest

dohaki:
Similar to #707 but related to the fees query. Example issue on Sentry https://risk-labs-m6.sentry.io/issues/3938438386/events/bbf1a146109f4ea1ac878ffdfa36f24c/


james-a-morris (approved):
LGTM

PR #711, opened 2023-04-21
fix(ampli): reset user props on account change

dohaki:
Fixes ACX-1044

The initial user props were not reset when the user connected a wallet by clicking the `Connect wallet` button or changed the account in the wallet directly.

This PR fixes the issue by re-running the `src/hooks/useInitialUserPropTraces.ts` hook if the connected account changes. We also make sure to wrap every tracking call with the `addToAmpliQueue` function, to avoid missing updated user properties.

### DIFF
--- a/src/views/Bridge/components/ChangeAccountModal.tsx
+++ b/src/views/Bridge/components/ChangeAccountModal.tsx
@@ -6,6 +6,7 @@ import { ReactComponent as CrossIcon } from "assets/icons/cross-16.svg";
 import { SecondaryButtonWithoutShadow as UnstyledButton } from "components/Buttons";
 import { ethers } from "ethers";
 import { ampli } from "ampli";
+import { useAmplitude } from "hooks";
 
 type ChangeAccountModalProps = {
   displayModal: boolean;
@@ -23,6 +24,8 @@ const ChangeAccountModal = ({
   const [userInput, setUserInput] = useState(currentAccount);
   const [validInput, setValidInput] = useState(false);
 
+  const { addToAmpliQueue } = useAmplitude();
+
   useEffect(() => {
     if (displayModal) {
       setUserInput(currentAccount);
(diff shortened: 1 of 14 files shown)


src/views/Bridge/components/ChangeAccountModal.tsx
```diff
@@ -38,8 +41,10 @@ const ChangeAccountModal = ({
   const onSaveHandler = () => {
     if (validInput) {
       changeAccountHandler(userInput);
-      ampli.toAccountChanged({
-        toWalletAddress: userInput,
+      addToAmpliQueue(async () => {
```
james-a-morris, line 44: Do we need this function to be asynchronous?

src/views/Bridge/components/CoinSelector.tsx
```diff
@@ -42,7 +49,9 @@ function useCoinSelector(
   const maxBalanceOnClick = () => {
     if (currentBalance) {
       setUserAmountInput(tokenFormatterFn(currentBalance));
-      trackMaxButtonClicked("bridgeForm");
+      addToAmpliQueue(async () => {
```
james-a-morris, line 52: Does this fn need to be async?

james-a-morris (commented):
Left a few comments

src/views/Bridge/components/ChangeAccountModal.tsx
james-a-morris: > Yea, I couldn't find a matching type for `addToAmpliQueue` that would cover all cases. It is set here
> 
> https://github.com/across-protocol/frontend-v2/blob/9a6896f93ea021020368687fb670ce82a95b6d7d/src/hooks/useAmpliTracking.ts#L5

Could we use `Promise<void> | void`?

dohaki: Let me check

dohaki: Thanks fixed here https://github.com/across-protocol/frontend-v2/pull/711/commits/66fc1b7aa4755c315f839ad330454ba726690c4c 👍

james-a-morris (approved):

PR #712, opened 2023-04-21
fix: prevent setting across.to as referring_domain

dohaki:


amateima (approved):
Can I get more context for this change? How does it impact amplitude?

dohaki:
> Can I get more context for this change? How does it impact amplitude?

On Amplitude we have some `referring_domain` values set to `across.to` when we do internal links via `<a href />`. This overrides the "real" referring_domain that we are interested in. So this PR just ignores internal links

james-a-morris (approved):
LGTM! Should the goerli be failing or is the error transient?

dohaki:
> LGTM! Should the goerli be failing or is the error transient?

It was transient 👍

PR #715, opened 2023-04-24
fix: :ambulance: mitigate vuln

james-a-morris:
This PR mitigates any vuln from interacting with user input

### DIFF
--- a/api/available-routes.ts
+++ b/api/available-routes.ts
@@ -34,7 +34,11 @@ const handler = async (
     query,
   });
   try {
-    assert(query, AvailableRoutesQueryParamsSchema);
+    try {
+      assert(query, AvailableRoutesQueryParamsSchema);
+    } catch (error) {
+      throw new Error("Invalid query parameters");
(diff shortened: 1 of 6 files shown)


dohaki (approved):
LGTM

api/available-routes.ts
```diff
@@ -34,7 +34,11 @@ const handler = async (
     query,
   });
   try {
```
nicholaspai, line 36: Do we still want this outside `try`? Maybe not the PR to work on it but I think its not great practice to have such a large try-catch block

nicholaspai (approved):
Good solution

api/available-routes.ts
james-a-morris: Fair point! I've removed it

james-a-morris:
> LGTM. API tests are failing though

I have now fixed the tests. I'll deploy once they pass

mrice32 (approved):
This LGTM!

PR #716, opened 2023-04-24
fix: :fire: disable the fees endpoints for 2.5 upgrade

james-a-morris:


dohaki (changes requested):
Looks good 👍 But can we wait before merging? I want to have a final Goerli test after updating to the latest packages/addresses

api/_utils.ts
```diff
@@ -537,7 +537,7 @@ export const isRouteEnabled = (
       toChainId === toChain &&
       fromToken.toLowerCase() === fromTokenAddress.toLowerCase()
   );
-  return enabled;
+  return enabled && false; // TODO: This will disable all routes until we want to enable them
```
mrice32, line 540: Just to make sure I'm understanding the logic, this will cause the suggested-fees and limits endpoints to always fail because every route is disabled, right? The other endpoints will continue to work correctly, right?

mrice32 (approved):
LGTM!

dohaki (approved):

PR #719, opened 2023-04-25
fix(api): throw 400 with query sanitization

dohaki:
The PR https://github.com/across-protocol/frontend-v2/pull/715 fixed the vulnerability ACX-1051 but caused the API to throw 500s even though it should throw 400s.

This fixes this behavior by sanitizing the query params instead of throwing a generic error.


api/coingecko.ts
```diff
@@ -100,11 +101,9 @@ const handler = async (
...
   });
   try {
-    try {
-      assert(query, CoingeckoQueryParamsSchema);
-    } catch (error) {
-      throw new Error("Invalid query parameters");
-    }
+    query = sanitizeQuery(query);
+    assert(query, CoingeckoQueryParamsSchema);
+
```
james-a-morris, line 106: Escaping arbitrary user input can be tricky to keep up with. We'll need to consistently ensure that lodash is fully up to date, and even then, we are susceptible to any in-library bugs/shortcomings. 

I would advise we catch the arbitrary assert or wrap the assert in a utility function like what is currently on prod so that the error message does not get propagated to the return statement in the event that a complex query is able to circumvent being escaped.

james-a-morris (changes requested):
LGTM - but I would maintain our try/catch around the asserts. I described in the comment below.

api/coingecko.ts
dohaki: Hmm, so lodash does essentially
```js
function escapeHtml(unsafe) {
    return unsafe
         .replace(/&/g, "&amp;")
         .replace(/</g, "&lt;")
         .replace(/>/g, "&gt;")
         .replace(/"/g, "&quot;")
         .replace(/'/g, "&#039;");
}
```
which in our case should suffice no? Because the attack surface is not that big anyways.

As an API consumer, I like rich input error messages. For example
```
GET https://frontend-v2-git-sanitize-query-instead-of-generic-error-uma.vercel.app/api/coingecko

Returns:
At path: l1Token -- Expected a value of type `validAddress`, but received: `undefined`
```
instead of a generic input error message 🤔 But yea I guess it would be a trade-off

dohaki: @james-a-morris Even though I think escaping the query should suffice in our case, I added a commit that sanitizes the error message by removing the input value instead of escaping. See here https://github.com/across-protocol/frontend-v2/pull/719/commits/75eb93dd0200dda3f417a0a5b67be05f10a5d9bf

james-a-morris (approved):
LGTM!

PR #726, opened 2023-04-27
fix: correctly handle disabled staking pools

dohaki:
Fixes ACX-915

I noticed the root cause while working/testing new tokens.

### DIFF
--- a/src/views/LiquidityPool/components/EarnByStakingInfoBox.tsx
+++ b/src/views/LiquidityPool/components/EarnByStakingInfoBox.tsx
@@ -27,7 +27,7 @@ export function EarnByStakingInfoBox({
   );
   const { isConnected } = useConnection();
 
-  const showValueOrDash = repeatableTernaryBuilder(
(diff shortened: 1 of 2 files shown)


james-a-morris (approved):

src/views/LiquidityPool/components/EarnByStakingInfoBox.tsx
```diff
@@ -38,6 +38,7 @@ export function EarnByStakingInfoBox({
     : data?.apyData.baseRewardsApy;
 
   const textColor = selectedPoolAction === "add" ? "aqua" : "warning";
+  const isStakingPoolEnabled = data?.poolEnabled;
```
james-a-morris, line 41: Nit: we may want to convert this into a boolean for the case that `data === undefined`.
```suggestion
  const isStakingPoolEnabled = !!data?.poolEnabled;
```

dohaki: Yes good point

PR #728, opened 2023-04-28
fix: :bug: correct error readout in bridge tab

james-a-morris:
This change modifies the error readouts to produce the correct alert to the user depending on the number of funds they request to bridge.


dohaki (approved):
LGTM

PR #729, opened 2023-05-01
fix: remove old google site verification

mrice32:
Leaving this here is a security risk because it could allow users who previously had access to regain access.


evaldofelipe (approved):

james-a-morris (approved):
LGTM!

I'll miss you, old google site verification 😢

nicholaspai (approved):

PR #731, opened 2023-05-02
feat: add config to disable bridge tokens in the ui

dohaki:
Closes ACX-1100

Allows us to enable USDT in the API but disable temporarily in the UI

### DIFF
--- a/src/views/Bridge/hooks/useBridge.ts
+++ b/src/views/Bridge/hooks/useBridge.ts
@@ -29,7 +29,7 @@ import { BridgeLimitInterface } from "utils/serverless-api/types";
 import { useBridgeAction } from "./useBridgeAction";
 import { useBridgeDepositTracking } from "./useBridgeDepositTracking";
 
-const enabledRoutes = getConfig().getRoutes();
+const enabledRoutes = getConfig().getEnabledRoutes();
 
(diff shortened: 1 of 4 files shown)


src/views/Bridge/hooks/useBridge.ts
```diff
@@ -29,7 +29,7 @@ import { BridgeLimitInterface } from "utils/serverless-api/types";
 import { useBridgeAction } from "./useBridgeAction";
 import { useBridgeDepositTracking } from "./useBridgeDepositTracking";
 
-const enabledRoutes = getConfig().getRoutes();
+const enabledRoutes = getConfig().getEnabledRoutes();
```
james-a-morris, line 32: Does this also apply to any pools?

james-a-morris (approved):
Left a comment. Feel free to disregard if it isn't actionable

src/views/Bridge/hooks/useBridge.ts
dohaki: Hmm, good point. Let me check with @nicholaspai

dohaki: So the config should only apply for bridging

nicholaspai (approved):

PR #736, opened 2023-05-03
fix: move up USDT in token list

dohaki:


james-a-morris (approved):
LGTM - out of curiosity, what's the reason for moving this in the list? Is there something that requires the order or is this for an organizational reason?

nicholaspai (approved):

dohaki:
> LGTM - out of curiosity, what's the reason for moving this in the list? Is there something that requires the order or is this for an organizational reason?

@james-a-morris Korpi requested this ad-hoc :) My guess is to group all stables together somehow

PR #737, opened 2023-05-04
chore: clean up deps 

dohaki:
Bumps @across-protocol/sdk-v2 to 0.5.4

Removes obsolete deps:
- @across-protocol/across-token (relies on version from SDK)
- @across-protocol/contracts-v2 (relies on version from SDK)
- @datapunt/matomo-tracker-js
- @uma/contracts-frontend

This saves us ~300 kB in our node_modules bundle size. But there are still a lot of opportunities to reduce the size which we can tackle in upcoming PRs.


james-a-morris (approved):
This looks good to me. I will miss the PLAAP 😞

PR #740, opened 2023-05-09
feat: required api endpoints for dodo

dohaki:
This PR adds some helper endpoints in an effort to be included in the DODO aggregator:

- `GET /api/build-deposit-tx`
  - Populates a raw deposit tx that can be signed. See https://github.com/DODOEX/dodo-bridge-aggregator#iv-buildtransactiondata
- `GET /api/token-list`
  - Returns a list of enabled tokens per enabled chain. See https://github.com/DODOEX/dodo-bridge-aggregator#iii-tokenlist

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -681,6 +681,16 @@ export function validAddress() {
   );
 }
 
+export function validAddressOrENS() {
+  return define<string>("validAddressOrENS", (value) => {
+    const ensDomainRegex =
+      /[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)?/;
+    return (
+      utils.isAddress(value as string) || ensDomainRegex.test(value as string)
+    );
+  });
+}
+
 export function positiveIntStr() {
   return define<string>("positiveIntStr", (value) => {
     return Number.isInteger(Number(value)) && Number(value) > 0;
@@ -716,9 +726,21 @@ export function getLpCushion(
   );
 }
 
-export function tagReferrer(dataHex: string, referrerAddress: string) {
-  if (!ethers.utils.isAddress(referrerAddress)) {
-    throw new Error("Data must be a valid address");
+export async function tagReferrer(
+  dataHex: string,
+  referrerAddressOrENS: string
+) {
+  let referrerAddress: string | null;
+
+  if (ethers.utils.isAddress(referrerAddressOrENS)) {
+    referrerAddress = referrerAddressOrENS;
+  } else {
+    const provider = infuraProvider(1);
+    referrerAddress = await provider.resolveName(referrerAddressOrENS);
+  }
+
+  if (!referrerAddress) {
+    throw new Error("Invalid referrer address or ENS name");
   }
 
   if (!ethers.utils.isHexString(dataHex)) {
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -486,34 +486,23 @@ export const getProvider = (_chainId: number): providers.Provider => {
  * @returns The corresponding SpokePool for the given `_chainId`
  */
 export const getSpokePool = (_chainId: number): SpokePool => {
+  const spokePoolAddress = getSpokePoolAddress(_chainId);
+  return SpokePool__factory.connect(spokePoolAddress, getProvider(_chainId));
+};
+
+export const getSpokePoolAddress = (_chainId: number): string => {
   const chainId = _chainId.toString();
-  const provider = getProvider(_chainId);
   switch (chainId.toString()) {
     case "1":
-      return SpokePool__factory.connect(
-        "0x5c7BCd6E7De5423a257D81B442095A1a6ced35C5",
-        provider
-      );
+      return "0x5c7BCd6E7De5423a257D81B442095A1a6ced35C5";
     case "10":
-      return SpokePool__factory.connect(
-        "0x6f26Bf09B1C792e3228e5467807a900A503c0281",
-        provider
-      );
+      return "0x6f26Bf09B1C792e3228e5467807a900A503c0281";
     case "137":
-      return SpokePool__factory.connect(
-        "0x9295ee1d8C5b022Be115A2AD3c30C72E34e7F096",
-        provider
-      );
+      return "0x9295ee1d8C5b022Be115A2AD3c30C72E34e7F096";
     case "288":
-      return SpokePool__factory.connect(
-        "0xBbc6009fEfFc27ce705322832Cb2068F8C1e0A58",
-        provider
-      );
+      return "0xBbc6009fEfFc27ce705322832Cb2068F8C1e0A58";
     case "42161":
-      return SpokePool__factory.connect(
-        "0xe35e9842fceaCA96570B734083f4a58e8F7C5f2A",
-        provider
-      );
+      return "0xe35e9842fceaCA96570B734083f4a58e8F7C5f2A";
     default:
       throw new Error("Invalid chainId provided");
   }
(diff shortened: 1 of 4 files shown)


api/build-deposit-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const handler = async (
+  { query }: TypedVercelRequest<BuildDepositTxQueryParams>,
+  response: VercelResponse
+) => {
+  const logger = getLogger();
+  logger.debug({
+    at: "BuildDepositTx",
+    message: "Query data",
+    query,
```
nicholaspai, line 41: Does this cause any DOS vectors where `query` contains some really long, unexpected bytes?

api/build-deposit-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+      amount: amountInput,
+      token,
+      destinationChainId: destinationChainIdInput,
+      originChainId: originChainIdInput,
+      recipient,
+      relayerFeePct: relayerFeePctInput,
+      quoteTimestamp,
+      message = "0x",
+      maxCount = ethers.constants.MaxUint256.toString(),
+      referrer,
```
nicholaspai, line 56: Just want to think about security, are there any ways in which these query params can be set such that the API gets dossed?

api/build-deposit-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const value =
+      token === ethers.constants.AddressZero ? amount : ethers.constants.Zero;
+    const tx = await spokePool.populateTransaction.deposit(
+      recipient,
+      token,
+      amount,
+      destinationChainId,
+      relayerFeePct,
+      quoteTimestamp,
```
nicholaspai, line 80: Can we `floor` the `quoteTimestamp` similar to [how we do it here](https://github.com/across-protocol/frontend-v2/blob/master/api/suggested-fees.ts#L93) to increase chance of the same `deposit` query params being requested so that we can try caching it?

api/build-deposit-tx.ts
```diff
@@ -0,0 +1,109 @@
...
+      value: tx.value?.toString(),
+    };
+
+    logger.debug({
+      at: "BuildDepositTx",
+      message: "Response data",
+      responseJson,
+    });
+
+    response.status(200).json(responseJson);
```
nicholaspai, line 103: should we try to cache the response at all?

api/token-list.ts
```diff
@@ -0,0 +1,85 @@
...
+    // that period — effectively hiding the latency penalty of revalidation from them.
+    // If no request happened during that period, the cache became stale and the next request will revalidate normally.
+    logger.debug({
+      at: "TokenList",
+      message: "Response data",
+      responseJson: enrichedTokensPerChain,
+    });
+    response.setHeader(
+      "Cache-Control",
+      "s-maxage=21600, stale-while-revalidate=21600"
```
nicholaspai, line 77: We could probbaly cache this even longer since token lists shouldn't change but I'm not even sure that gives big performance improvements

api/build-deposit-tx.ts
dohaki: AFAIK Node.js has some defaults set already which limits the size of a request header, i.e. URL query string.

api/build-deposit-tx.ts
dohaki: As mentioned above, one protection is the limit set by HTTP module in Node.js which I think is around 16kB. I could also add additional max length validation for specific query params. But this would then make sense to add to all API endpoints, no?

api/build-deposit-tx.ts
dohaki: Does this still make sense even though `quoteTimestamp` will be directly passed from the response from the `/suggested-fees` endpoint?

In the DODO config the implementation will be something like
1. `GET /suggested-fees`
  - Returns fees and timestamp which is already floored, etc.
2. Use above response directly and `GET /build-deposit-tx`
  - Will send request with floored timestamp

api/build-deposit-tx.ts
dohaki: Hmm yea I guess this could make sense. What caching policy / time you think makes sense here?

api/suggested-fees.ts
```diff
@@ -167,6 +168,7 @@ const handler = async (
       timestamp: parsedTimestamp.toString(),
       isAmountTooLow: relayerFeeDetails.isAmountTooLow,
       quoteBlock: blockTag.toString(),
+      spokePoolAddress: getSpokePoolAddress(Number(originChainId)),
```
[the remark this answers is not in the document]
dohaki: Yes, good catch! Thanks a lot. Fixed here https://github.com/across-protocol/frontend-v2/pull/740/commits/7606d78c571cdc00c157ebb837be407f34df5d1a

api/token-list.ts
dohaki: Yes true, this handler doesn't really do any complex computations. But setting a higher max age also shouldn't hurt. I changed this to a week here https://github.com/across-protocol/frontend-v2/pull/740/commits/cbc9f00c19bfd3bd0b5c81e37fca98bf99e837ef

api/build-deposit-tx.ts
dohaki: Does this make sense https://github.com/across-protocol/frontend-v2/pull/740/commits/be68b0398556b1be8076ff395477ab75166395d9?

api/build-deposit-tx.ts
nicholaspai: yeah good call, nothing to do then

api/token-list.ts
nicholaspai: I think this endpoint might actually just be a candidate for the classic cache header `response.setHeader("Cache-Control", "s-maxage= 21600")`

I think 6 hours is fine to be honest

api/build-deposit-tx.ts
dohaki: Yes this makes sense. Added here https://github.com/across-protocol/frontend-v2/pull/740/commits/e902d56cd4649ad3a0e72245ac2b1c4d5756c18a#diff-2e8adfa0e95cabb0fcb55d612df1124940549ca8ea68a31c9d7062bbede8cfcfR119-R122

api/build-deposit-tx.ts
dohaki: Added here https://github.com/across-protocol/frontend-v2/pull/740/commits/e902d56cd4649ad3a0e72245ac2b1c4d5756c18a#diff-2e8adfa0e95cabb0fcb55d612df1124940549ca8ea68a31c9d7062bbede8cfcfR53-R54

api/token-list.ts
dohaki: Changed here https://github.com/across-protocol/frontend-v2/pull/740/commits/07c2945a7c0a8653dad211f5733d01a43f51fccf

api/token-list.ts
```diff
@@ -0,0 +1,71 @@
...
+import { constants } from "@across-protocol/sdk-v2";
+import {
+  getLogger,
+  handleErrorCondition,
+  getFallbackTokenLogoURI,
+  ENABLED_ROUTES,
+} from "./_utils";
+import { TypedVercelRequest } from "./_types";
+
+const handler = async (_: TypedVercelRequest<{}>, response: VercelResponse) => {
```
mrice32, line 11: nit: this seems similar to available-routes. Can we reuse the logic there or use a common implementation that we put in utils to reduce bloat? If they're too dissimilar, no worries, not trying to squeeze a round peg in a square hole or anything, just wanted to mention it.

mrice32 (approved):
LGTM!

api/_utils.ts
```diff
@@ -726,3 +715,30 @@ export function getLpCushion(
...
   );
 }
+
+export function tagReferrer(dataHex: string, referrerAddress: string) {
+  if (!ethers.utils.isAddress(referrerAddress)) {
+    throw new Error("Data must be a valid address");
+  }
+
+  if (!ethers.utils.isHexString(dataHex)) {
+    throw new Error("Data must be a valid hex string");
```
james-a-morris, line 725: Will this function fail if an `ENS` URI is used?

api/_utils.ts
```diff
@@ -726,3 +715,30 @@ export function getLpCushion(
...
+
+export function getFallbackTokenLogoURI(l1TokenAddress: string) {
+  const isACX =
+    sdk.constants.TOKEN_SYMBOLS_MAP.ACX.addresses[1] === l1TokenAddress;
+
+  if (isACX) {
+    return "https://across.to/logo-small.png";
+  }
+
+  return `https://github.com/trustwallet/assets/blob/master/blockchains/ethereum/assets/${l1TokenAddress}/logo.png?raw=true`;
```
james-a-morris, line 743: Is this URI subject to rate limiting? We may want to host this in a dedicated location. Another thought: we are susceptible to 404's if trust wallet refactors their repo

james-a-morris (commented):
Left a few nits

api/_utils.ts
nicholaspai: +1 we should try to use similar logic to the FE UI to resolve ENS --> address

api/_utils.ts
nicholaspai: I think for common tokens DODO's backend will figure out the token URI itself without using our fallback

api/_utils.ts
dohaki: good point! added here https://github.com/across-protocol/frontend-v2/pull/740/commits/34162efd896d847f4c3e3cb0758b981d8725e7d8

james-a-morris (approved):
LGTM!

PR #751, opened 2023-05-25
chore(deps): batch bump deps flagged by dependabot

dohaki:
This batch updates most of the deps flagged by dependabot in multiple PRs. I only included non-breaking updates. We also still need to add some secrets to dependabot so that our test suites succeed. Until then, we need to manually bump deps so that the test suites can run on our CI with the "normal" secrets configured.


james-a-morris (approved):
All tests pass, but I think we should monitor this via the sentry error readouts.

PR #765, opened 2023-06-08
fix: weth/eth tooltip and warning

dohaki:
Fixes ACX-1119

This refines the tooltip and warning based on the behavior for ETH/WETH:
- If a bridge transfer is being sent to an EOA, the EOA will receive ETH (not WETH)
- If a bridge transfer is being sent to a contract or to Polygon, the contract will receive WETH (not ETH)

### DIFF
--- a/src/views/Bridge/components/BridgeForm.tsx
+++ b/src/views/Bridge/components/BridgeForm.tsx
@@ -19,8 +19,9 @@ import EstimatedTable from "./EstimatedTable";
 import QuickSwap from "./QuickSwap";
 import SlippageAlert from "./SlippageAlert";
 import { BigNumber } from "ethers";
-import { ToAccount } from "../hooks/useToAccount";
 import { getReceiveTokenSymbol } from "../utils";
+
+import type { ToAccount } from "../hooks/useToAccount";
 
 type BridgeFormProps = {
   availableTokens: TokenInfo[];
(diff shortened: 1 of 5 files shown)


src/views/Bridge/components/BridgeForm.tsx
```diff
@@ -20,6 +20,7 @@ import QuickSwap from "./QuickSwap";
 import SlippageAlert from "./SlippageAlert";
 import { BigNumber } from "ethers";
 import { ToAccount } from "../hooks/useToAccount";
```
james-a-morris, line 22: QQ: would the naming convention be `toAccount`?

src/views/Bridge/utils.ts
```diff
@@ -0,0 +1,20 @@
+export function getReceiveTokenSymbol(
```
james-a-morris, line 1: Should we include JSDoc comments?

james-a-morris (approved):
LGTM - left a few comments

src/views/Bridge/components/BridgeForm.tsx
dohaki: It is a type, that's why it starts with uppercase... I can make this more explicit by adding the `type` keyword I guess.

dohaki: Added here https://github.com/across-protocol/frontend-v2/pull/765/commits/cf05fe5df34a3ed86cc46ae7da23ca5f3d4d4aa9

PR #778, opened 2023-06-30
feat: :art: only reset bridge input on chain switch

james-a-morris:
This change disallows the input to be automatically reset when switching a chain. When a user attempts to change chains, their input token size will not be changed. However, this may cause invalid inputs if the user switches to a chain where the input token balance is insufficient to bridge.

Breaking Changes: none


nicholaspai (approved):
Nice fix!

PR #780, opened 2023-07-04
feat: migrate to vite + web3-onboard upgrades

dohaki:
While trying to upgrade `@web-onboard/core` to the latest version and also the respective wallet packages, I encountered several babel issues with our build tool (CRA aka react-scripts)... I tried different workarounds to customize the babel config (updating to v5, ejecting CRA, CRA config overrides like CRACO) but none of them really worked. they either led to other issues or made the configuration very messy.

I then tried migrating to a different tool [Vite](https://vitejs.dev/) which can be configured to our needs and is also way faster. As CRA is also [kinda officially deprecated](https://github.com/reactjs/react.dev/pull/5487), not maintained and even not recommended in the [new React docs anymore](https://react.dev/learn/start-a-new-react-project#production-grade-react-frameworks), I think now is actually a good time to switch. The required diff is also quite manageable as you can see in this PR. The configuration wasn't that straightforward as I hoped it to be but still not that bad.

This PR also makes use of WalletConnect V2 and fixes the TrustWallet connection by upgrading the packages.


package.json
```diff
@@ -40,22 +37,22 @@
     "react-dom": "^17.0.2",
     "react-feather": "^2.0.9",
     "react-lottie": "^1.2.3",
-    "react-popper": "^2.3.0",
```
dohaki, line 43: This lib was causing weird CSS issues with Vite. I replaced it with `react-tooltip` which also has a leaner API and smaller bundle size

package.json
```diff
@@ -123,16 +123,26 @@
...
     "cypress": "^10.4.0",
+    "eslint": "^8.44.0",
+    "eslint-config-react-app": "^7.0.1",
     "eslint-plugin-prettier": "^4.2.1",
     "hardhat": "^2.12.7",
     "husky": "^8.0.0",
+    "jest": "^29.5.0",
     "jest-environment-hardhat": "^1.1.8",
+    "jest-environment-jsdom": "^29.5.0",
+    "jest-transform-stub": "^2.0.0",
```
dohaki, line 134: Because we removed CRA/react-scripts, we need to explicitly add jest and eslint

src/components/WrongNetworkHeader/WrongNetworkHeader.tsx
```diff
@@ -1,5 +1,5 @@
 import { useIsWrongNetwork } from "hooks";
-import { SuperHeader } from "components";
+import SuperHeader from "components/SuperHeader/SuperHeader";
```
dohaki, line 2: Circular dependency was causing issues

tsconfig.json
```diff
@@ -9,13 +9,16 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
-    "module": "esnext",
+    "module": "CommonJS",
```
dohaki, line 12: This fixes https://github.com/orgs/vercel/discussions/1225. We removed the pinned `VERCEL_CLI_VERSION` but this change makes the serverless functions compatible with the latest versions

src/utils/onboard.ts
```diff
@@ -3,13 +3,16 @@ import injectedModule from "@web3-onboard/injected-wallets";
...
 import { providerUrlsTable } from "utils/providers";
 import logo from "assets/across-logo-v2.svg";
 
 const injected = injectedModule();
 const gnosis = gnosisModule();
-const walletConnect = walletConnectModule();
+const walletConnect = walletConnectModule({
+  projectId: walletConnectProjectId,
+  version: 2,
+});
```
james-a-morris, line 15: Should the `version` be hardcoded as a constant?

james-a-morris (approved):
This seems straightforward (albeit quite in-depth). I left a few nits, but otherwise LGTM

src/utils/constants.ts
```diff
@@ -296,28 +299,12 @@ export const tokenList: TokenInfoList = Object.entries(
...
-  "Missing process.env.REACT_APP_REWARDS_API_URL"
-);
-assert(
-  process.env.REACT_APP_CHAIN_137_PROVIDER_URL,
-  "REACT_APP_CHAIN_137_PROVIDER_URL must be defined."
-);
-assert(
-  process.env.REACT_APP_CHAIN_42161_PROVIDER_URL,
-  "REACT_APP_CHAIN_42161_PROVIDER_URL must be defined."
-);
```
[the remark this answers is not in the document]
dohaki: Yea, they were causing some issues in the production build. But I noticed that they don't have to be defined actually. Because we have default values so it made sense to remove them.

src/utils/onboard.ts
dohaki: I don't think we need to at this point because this is really the only place in the code where it is used and relevant

PR #782, opened 2023-07-06
fix: CSP for walletconnect v2

dohaki:
Walletconnect V2 uses different domains which we need to whitelist to suppress some noisy errors


evaldofelipe (approved):
LGTM!

PR #783, opened 2023-07-07
fix: pass in required chains wc2 init

dohaki:


james-a-morris (approved):

dohaki:
Unfortunately, this doesn't fix the issue with Safe via WalletConnect V2. Context on the underlying issue is discussed here https://github.com/orgs/WalletConnect/discussions/2733. A fix is already released here https://github.com/WalletConnect/walletconnect-monorepo/releases/tag/2.9.0 but still needs to be updated by the web3-onboard package. The respective PR is here https://github.com/blocknative/web3-onboard/pull/1810.

If the above PR is merged and a new version is released I will open a new PR.

PR #787, opened 2023-07-14
feat(uba): vercel `GET /uba-client-state` handler

dohaki:
Closes ACX-1295

Leaving as draft until testing is done

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -50,6 +50,8 @@ export const DEFAULT_GAS_MARKUP = 0;
 
 // Don't permit HUB_POOL_CHAIN_ID=0
 export const HUB_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID || 1);
+export const HUB_POOL_DEPLOYMENT_BLOCK =
+  HUB_POOL_CHAIN_ID === 1 ? 14819537 : 7370152;
 
 // Permit REACT_APP_FLAT_RELAY_CAPITAL_FEE=0
 export const FLAT_RELAY_CAPITAL_FEE = Number(
@@ -92,7 +94,6 @@ export const SUPPORTED_CHAIN_IDS =
     ? [
         sdk.constants.CHAIN_IDs.MAINNET,
         sdk.constants.CHAIN_IDs.ARBITRUM,
-        sdk.constants.CHAIN_IDs.BOBA,
         sdk.constants.CHAIN_IDs.OPTIMISM,
         sdk.constants.CHAIN_IDs.POLYGON,
       ]
@@ -186,7 +187,7 @@ export const getWinstonLogger = () => {
 
   const transports = [
     new winston.transports.Console({
-      level: "info",
+      level: "debug",
     }),
     isGoogleServiceAccountSet
       ? new LoggingWinston({
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1,18 +1,29 @@
 import {
   HubPool__factory,
+  HubPool,
   ERC20__factory,
   SpokePool__factory,
   SpokePool,
+  AcrossConfigStore__factory,
+  AcrossConfigStore,
 } from "@across-protocol/contracts-v2/dist/typechain";
 import axios from "axios";
 import * as sdk from "@across-protocol/sdk-v2";
 import { BigNumber, ethers, providers, utils } from "ethers";
 import { Log, Logging } from "@google-cloud/logging";
+import winston from "winston";
+import { LoggingWinston } from "@google-cloud/logging-winston";
 import { define, StructError } from "superstruct";
+
 import enabledMainnetRoutesAsJson from "../src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json";
 import enabledGoerliRoutesAsJson from "../src/data/routes_5_0x0e2817C49698cc0874204AeDf7c72Be2Bb7fCD5d.json";
 
-import { maxRelayFeePct, relayerFeeCapitalCostConfig } from "./_constants";
+import {
+  maxRelayFeePct,
+  relayerFeeCapitalCostConfig,
+  SPOKE_POOLS,
+} from "./_constants";
+
 import { StaticJsonRpcProvider } from "@ethersproject/providers";
 import QueryBase from "@across-protocol/sdk-v2/dist/relayFeeCalculator/chain-queries/baseQuery";
 import { VercelResponse } from "@vercel/node";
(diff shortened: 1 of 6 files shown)


api/_utils.ts
```diff
@@ -71,6 +82,21 @@ _ENABLED_ROUTES.routes = _ENABLED_ROUTES.routes.filter(
...
+  new Set(ENABLED_ROUTES.routes.map(({ fromTokenSymbol }) => fromTokenSymbol))
+);
+
+// Complete set of chain IDs ever supported by the HubPool
+export const SUPPORTED_CHAIN_IDS =
+  HUB_POOL_CHAIN_ID === 1
+    ? [
+        sdk.constants.CHAIN_IDs.MAINNET,
+        sdk.constants.CHAIN_IDs.ARBITRUM,
+        sdk.constants.CHAIN_IDs.BOBA,
```
dohaki, line 95: Not sure about this one 🤔 Can this be omitted or do we need to add this

api/_utils.ts
```diff
@@ -154,6 +180,34 @@ export const getLogger = (): LoggingUtility => {
   return logger;
 };
 
+export const getWinstonLogger = () => {
```
dohaki, line 183: I had add winston logger because there were type incompatibilities with the new clients

api/uba-client-state.ts
```diff
@@ -0,0 +1,132 @@
...
+        return spokePoolDeploymentBlock.number;
+      })
+    );
+
+    const configStoreClient = new clients.AcrossConfigStoreClient(
+      logger,
+      getAcrossConfigStore(),
+      {
+        fromBlock: HUB_POOL_DEPLOYMENT_BLOCK,
+        maxBlockLookBack: 10_000,
```
dohaki, line 59: Do these make sense?

api/uba-client-state.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const hubPoolClient = new clients.HubPoolClient(
+      logger,
+      getHubPool(),
+      configStoreClient,
+      HUB_POOL_DEPLOYMENT_BLOCK,
+      HUB_POOL_CHAIN_ID,
+      {
+        fromBlock: HUB_POOL_DEPLOYMENT_BLOCK,
+        maxBlockLookBack: 10_000,
```
dohaki, line 74: Do these make sense?

api/uba-client-state.ts
```diff
@@ -0,0 +1,132 @@
...
+          ...acc,
+          [chainId]: new clients.SpokePoolClient(
+            logger,
+            getSpokePool(chainId),
+            hubPoolClient,
+            chainId,
+            SPOKE_POOLS[chainId].deploymentBlock,
+            {
+              fromBlock: spokePoolFromBlocks[i],
+              maxBlockLookBack: 10_000,
```
dohaki, line 90: Do theses make sense?

api/uba-client-state.ts
```diff
@@ -0,0 +1,132 @@
...
+        };
+      },
+      {}
+    );
+
+    const ubaState = await clients.updateUBAClient(
+      hubPoolClient,
+      spokePoolClientsMap,
+      SUPPORTED_CHAIN_IDS,
+      ENABLED_TOKEN_SYMBOLS.filter((symbol) => symbol !== "ETH"),
```
dohaki, line 102: Does this make sense? I was getting an error `L1 token not found for ETH`

api/_utils.ts
```diff
@@ -39,6 +57,11 @@ export const DEFAULT_GAS_MARKUP = 0;
 
 // Don't permit HUB_POOL_CHAIN_ID=0
 export const HUB_POOL_CHAIN_ID = Number(REACT_APP_HUBPOOL_CHAINID || 1);
+export const HUB_POOL_DEPLOYMENT_BLOCK =
+  HUB_POOL_CHAIN_ID === 1 ? 14819537 : 7370152;
+
+export const CONFIG_STORE_DEPLOYMENT_BLOCK =
+  HUB_POOL_CHAIN_ID === 1 ? 14717196 : 14717196;
```
nicholaspai, line 64: Why are these the same block? Should it be a goerli block as the second one?

api/uba-client-state.ts
```diff
@@ -0,0 +1,92 @@
...
+  handleErrorCondition,
+  ENABLED_TOKEN_SYMBOLS,
+  SUPPORTED_CHAIN_IDS,
+  getWinstonLogger,
+  validTokenSymbol,
+  getHubAndSpokeClients,
+} from "./_utils";
+import { TypedVercelRequest } from "./_types";
+
+const SPOKE_POOL_FROM_BLOCK_MS_OFFSET = 24 * 60 * 60 * 1000; // 24 hours
```
nicholaspai, line 15: This seems reasonable offset if you're only loading one UBA bundle state. As a general rule FYI if I want to load X bundle states, I assume each bundle state spans 4 hours, so I'd want X * 4 * 2 hours of offset here. 2 is there for a buffer

api/uba-client-state.ts
```diff
@@ -0,0 +1,92 @@
...
+      hubPoolClient,
+      spokePoolClientsMap,
+      SUPPORTED_CHAIN_IDS,
+      tokenSymbols.filter((symbol) => symbol !== "ETH"),
+      true,
+      1
+    );
+    const serializedUBAState = JSON.parse(
+      clients.serializeUBAClientState(ubaClientState)
+    );
```
nicholaspai, line 67: FYI I think `suggested-fees` could take this `ubaClientState`, create a new `UBAClient` and pass it in to the `update` method, and then call [`getLatestFeesForDeposit`](https://github.com/across-protocol/sdk-v2/blob/738ca4d9d19bf32caf030c18226e6a6705bd57e3/src/clients/UBAClient/UBAClientWithRefresh.ts#L70)

api/uba-client-state.ts
```diff
@@ -0,0 +1,92 @@
...
+      // connections.
+      // Using multiple instances of `JsonRpcBatchProvider` instead makes the RPC requests more reliable.
+      {
+        useBatch: true,
+        useUncached: true,
+      }
+    );
+    await hubPoolClient.configStoreClient.update();
+
+    const ubaClientState = await clients.updateUBAClient(
```
nicholaspai, line 57: Can you add some timers around each of these `await` calls to figure out which is taking the longest?

nicholaspai (approved):
This is really easy to read code +1 

Now I think we need to optimize performance as much as we can.

dohaki:
Closing this one in favor of https://github.com/across-protocol/frontend-v2/pull/790

PR #790, opened 2023-07-19
feat: extract writing uba client state

dohaki:
This PR extracts the task of computing and storing the UBA client state into a separate script.


dohaki:
@james-a-morris sorry we can actually close this PR because this will moved out of this repo! so you concerns will be addressed

PR #792, opened 2023-07-21
feat: adapt suggested-fees to uba

dohaki:
WIP


james-a-morris:
I think we can close this out? cc @dohaki @nicholaspai

dohaki:
yep can be closed for now

PR #794, opened 2023-07-25
fix: info on usdc/usdc.e on arbitrum

dohaki:
Fixes ACX-1345

### DIFF
--- a/src/components/Selector/Selector.tsx
+++ b/src/components/Selector/Selector.tsx
@@ -16,7 +16,6 @@ export type SelectorElementType<Value> = {
     title: string;
     description: string | JSX.Element;
   };
-  hidden?: boolean;
 };
 
 export type SelectorPropType<Value> = {
@@ -88,7 +87,6 @@ const Selector = <ElementValue,>({
               }}
               active={!element.disabled && selectedIndex === idx}
               disabled={element.disabled}
-              hidden={element.hidden}
             >
               <ElementSection disabled={element.disabled}>
                 {element.element}
@@ -200,9 +198,8 @@ const ElementRowDivider = styled.div`
 const ElementRow = styled.div<{
   active: boolean;
   disabled?: boolean;
-  hidden?: boolean;
 }>`
-  display: ${({ hidden }) => (hidden ? "none" : "flex")};
+  display: flex;
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
--- a/src/components/Selector/Selector.tsx
+++ b/src/components/Selector/Selector.tsx
@@ -14,8 +14,9 @@ export type SelectorElementType<Value> = {
   disabled?: boolean;
   disabledTooltip?: {
     title: string;
-    description: string;
+    description: string | JSX.Element;
   };
+  hidden?: boolean;
 };
 
 export type SelectorPropType<Value> = {
@@ -87,6 +88,7 @@ const Selector = <ElementValue,>({
               }}
               active={!element.disabled && selectedIndex === idx}
               disabled={element.disabled}
+              hidden={element.hidden}
             >
               <ElementSection disabled={element.disabled}>
                 {element.element}
@@ -102,6 +104,7 @@ const Selector = <ElementValue,>({
                   )}
                   {element.disabled ? (
                     <Tooltip
+                      tooltipId={`selector-disabled-${idx}`}
                       title={element.disabledTooltip?.title ?? "Not supported."}
                       body={
                         element.disabledTooltip?.description ??
@@ -194,8 +197,12 @@ const ElementRowDivider = styled.div`
   width: calc(100% + 32px);
 `;
 
-const ElementRow = styled.div<{ active: boolean; disabled?: boolean }>`
-  display: flex;
+const ElementRow = styled.div<{
+  active: boolean;
+  disabled?: boolean;
+  hidden?: boolean;
+}>`
+  display: ${({ hidden }) => (hidden ? "none" : "flex")};
   flex-direction: row;
   justify-content: space-between;
   align-items: center;
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -258,6 +258,8 @@ export type TokenInfo = {
   logoURI: string;
   // tokens require a mainnet address to do price lookups on coingecko, not used for anything else.
   mainnetAddress?: string;
+  // optional display symbol for tokens that have a different symbol on the frontend
+  displaySymbol?: string;
 };
 // enforce weth to be first so we can use it as a guarantee in other parts of the app
 export type TokenInfoList = TokenInfo[];
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -271,6 +271,7 @@ export const orderedTokenSymbolLogoMap = {
   MATIC: polygonLogo,
   WMATIC: polygonLogo,
   USDC: usdcLogo,
+  "USDC.e": usdcLogo,
   USDT: usdtLogo,
   DAI: daiLogo,
   WBTC: wbtcLogo,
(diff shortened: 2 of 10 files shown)


src/components/Selector/Selector.tsx
```diff
@@ -14,8 +14,9 @@ export type SelectorElementType<Value> = {
   disabled?: boolean;
   disabledTooltip?: {
     title: string;
-    description: string;
+    description: string | JSX.Element;
   };
+  hidden?: boolean;
```
nicholaspai, line 19: Should `hidden` be different from `disabled`?

src/components/Selector/Selector.tsx
```diff
@@ -87,6 +88,7 @@ const Selector = <ElementValue,>({
               }}
               active={!element.disabled && selectedIndex === idx}
               disabled={element.disabled}
+              hidden={element.hidden}
```
james-a-morris, line 91: Should we disable the `onClick` if this element is also `element.hiden`

src/components/Selector/Selector.tsx
```diff
@@ -194,8 +197,12 @@ const ElementRowDivider = styled.div`
...
 `;
 
-const ElementRow = styled.div<{ active: boolean; disabled?: boolean }>`
-  display: flex;
+const ElementRow = styled.div<{
+  active: boolean;
+  disabled?: boolean;
+  hidden?: boolean;
+}>`
+  display: ${({ hidden }) => (hidden ? "none" : "flex")};
```
james-a-morris, line 205: Nit: should we do something like `visible` instead of `hidden`? My understanding is that `hidden` is a valid CS property and in this case we're using `hidden` to refer to the display type.

src/utils/constants.ts
```diff
@@ -283,9 +286,23 @@ export const orderedTokenSymbolLogoMap = {
...
+      symbol: "USDC.e",
+      displaySymbol: "USDC.e",
+      mainnetAddress: usdcTokenInfo.addresses[constants.CHAIN_IDs.MAINNET],
+    };
+  }
+
   const tokenInfo =
     constants.TOKEN_SYMBOLS_MAP[
-      symbol as keyof typeof orderedTokenSymbolLogoMap
+      symbol as keyof Omit<typeof orderedTokenSymbolLogoMap, "USDC.e">
```
james-a-morris, line 305: Just to confirm - USDC.e is really just a label for USDC for a specific route path?

src/views/Bridge/components/CoinSelector.tsx
```diff
@@ -252,7 +252,7 @@ const CoinSelector = ({
               {currentSelectedBalance && (displayBalance || userAmountInput) && (
                 <Text size="sm" color="grey-400">
                   Balance: {tokenFormatterFn(currentSelectedBalance)}{" "}
-                  {token.symbol.toUpperCase()}
+                  {token.symbol}
```
james-a-morris, line 255: Will this impact any additional symbol names?

src/views/Bridge/components/CoinSelector.tsx
```diff
@@ -297,17 +297,19 @@ const CoinSelector = ({
...
+                toChain={toChain}
+              />
+            ),
           },
           element: (
             <CoinIconTextWrapper>
               <CoinIcon src={t.logoURI} />
               <Text size="md" color="white-100">
-                {t.symbol.toUpperCase()}
+                {t.displaySymbol || t.symbol.toUpperCase()}
```
james-a-morris, line 312: Will we need to use `t.symbol` instead of `t.symbol.toUppercase()` like above

src/views/Bridge/components/CoinSelector.tsx
```diff
@@ -325,7 +327,7 @@ const CoinSelector = ({
           <CoinIconTextWrapper>
             <CoinIcon src={token.logoURI} />
             <Text size="md" color="white-100">
-              {token.symbol.toUpperCase()}
+              {token.displaySymbol || token.symbol.toUpperCase()}
```
james-a-morris, line 330: Similar to above ^

src/views/Bridge/components/RouteNotSupportedTooltipText.tsx
```diff
@@ -0,0 +1,51 @@
...
+  toChain: number;
+};
+
+export const RouteNotSupportedTooltipText = ({
+  symbol,
+  fromChain,
+  toChain,
+}: Props) => {
+  const isArbitrumNativeUSDC =
+    (fromChain === 42161 || toChain === 42161) && symbol === "USDC";
```
james-a-morris, line 16: If we are storing the hardcoded chain ID, should we store it at least in a constants file?

vite.config.js
```diff
@@ -47,7 +47,7 @@ export default defineConfig({
           process: true,
           buffer: true,
         }),
-        NodeModulesPolyfillPlugin(),
+        // NodeModulesPolyfillPlugin(),
```
james-a-morris, line 50: Is this meant to be part of prod?

vite.config.js
```diff
@@ -4,7 +4,7 @@ import react from "@vitejs/plugin-react";
 import tsconfigPaths from "vite-tsconfig-paths";
 import svgr from "vite-plugin-svgr";
 import { NodeGlobalsPolyfillPlugin } from "@esbuild-plugins/node-globals-polyfill";
-import { NodeModulesPolyfillPlugin } from "@esbuild-plugins/node-modules-polyfill";
+// import { NodeModulesPolyfillPlugin } from "@esbuild-plugins/node-modules-polyfill";
```
james-a-morris, line 7: Should this be removed?

james-a-morris (commented):
Left a few nits

src/components/Selector/Selector.tsx
dohaki: We can remove this `hidden` property... It's still a leftover from a previous approach actually

src/components/Selector/Selector.tsx
dohaki: We can remove this `hidden` property... It's still a leftover from a previous approach actually

src/utils/constants.ts
dohaki: Yea exactly... It is a bit confusing. Basically `USDC` is the symbol for native USDC but on Arbitrum we actually are currently only supporting the bridged `USDC.e` token. On other chains, though, it can be referred to as `USDC`.

src/views/Bridge/components/CoinSelector.tsx
dohaki: Not really... I added a `displaySymbol` property though. Maybe it's cleaner to use that instead.

src/views/Bridge/components/CoinSelector.tsx
dohaki: I think this approach is cleaner. We can assume that `symbol` is always uppercase. But the optional `displaySymbol` will never be modified. Wdyt?

src/views/Bridge/components/CoinSelector.tsx
dohaki: I think this approach is cleaner. We can assume that `symbol` is always uppercase. But the optional `displaySymbol` will never be modified. Wdyt?

vite.config.js
dohaki: yes will remove

vite.config.js
dohaki: Removed https://github.com/across-protocol/frontend-v2/pull/794/commits/124ab5933dadea32de9e7170c0ab40101a8aad89

vite.config.js
dohaki: removed https://github.com/across-protocol/frontend-v2/pull/794/commits/124ab5933dadea32de9e7170c0ab40101a8aad89

src/views/Bridge/components/RouteNotSupportedTooltipText.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/794/commits/b9a163aac775375c6be28384d0e540dc25dd4d6c

james-a-morris (approved):
LGTM - thanks for reviewing my comments

PR #797, opened 2023-08-02
feat(api): extend `/pools` to handle balancer pool tokens

dohaki:
Extends `/pools` API to allow
```
GET /pools?token=0x&externalPoolProvider=balancer
```

Adds support for external Balancer pool. The staking page will eventually call this endpoint on the frontend to determine the pool APY.

Part of ACX-1351

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -1,5 +1,5 @@
 import { ethers } from "ethers";
-import { constants, relayFeeCalculator } from "@across-protocol/sdk-v2";
+import { constants, relayFeeCalculator, utils } from "@across-protocol/sdk-v2";
 
 export const maxRelayFeePct = 0.25;
 
@@ -101,3 +101,6 @@ export const BLOCK_TAG_LAG = -1;
 // Note: this is a small subset of all the supported base currencies, but since we don't expect to use the others,
 // we've decided to keep this list small for now.
 export const SUPPORTED_CG_BASE_CURRENCIES = new Set(["eth", "usd"]);
+
+// 1:1 because we don't need to handle underlying tokens on FE
+export const EXTERNAL_POOL_TOKEN_EXCHANGE_RATE = utils.fixedPointAdjustment;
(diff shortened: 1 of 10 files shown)


api/_utils.ts
```diff
@@ -787,3 +789,64 @@ export function getFallbackTokenLogoURI(l1TokenAddress: string) {
...
+    throw new InputError(
+      `Balancer pool with address ${poolTokenAddress} not found`
+    );
+  }
+
+  const apr = await balancer.pools.apr(pool);
+
+  return {
+    estimatedApy: Number(apr.max / 1000).toFixed(2),
+    exchangeRateCurrent: utils.parseEther("1").toString(), // 1:1 because we don't need to handle underlying tokens on FE
```
pxrl, line 850: Since this is a constant it might be nice to instantiate it globally (in this file, or maybe import `fixedPoint` and rename it?), and then just reference it in this return statement.

dohaki: Good call. Changed here https://github.com/across-protocol/frontend-v2/pull/797/commits/85d82cec3c38f71168990ca52bbd2b21e8b58eca

nicholaspai (approved):
This lgtm

api/_constants.ts
```diff
@@ -101,3 +101,6 @@ export const BLOCK_TAG_LAG = -1;
 // Note: this is a small subset of all the supported base currencies, but since we don't expect to use the others,
 // we've decided to keep this list small for now.
 export const SUPPORTED_CG_BASE_CURRENCIES = new Set(["eth", "usd"]);
+
+// 1:1 because we don't need to handle underlying tokens on FE
+export const EXTERNAL_POOL_TOKEN_EXCHANGE_RATE = utils.fixedPointAdjustment;
```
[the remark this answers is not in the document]
dohaki: We use this exchange rate to calculate the USD value of a pool based on the underlying token for Across LP tokens. But if we assume that we always know the USD value of external LP tokens, then I think this will always be 1:1.

PR #799, opened 2023-08-02
feat: allow usage of override infura provider urls everywhere

dohaki:
This allows the usage of the env var `REACT_APP_CHAIN_<CHAIN_ID>_PROVIDER_URL` to override the default Infura provider URL. 

Setting the above env var will affect all RPC calls for the respective chain in all API handlers and on the frontend. Previously, we had the env var `OVERRIDE_PROVIDER_<CHAIN_ID>` which will be replaced by the shared one on the frontend.

### DIFF
--- a/src/utils/providers.ts
+++ b/src/utils/providers.ts
@@ -1,25 +1,48 @@
 import { ethers, providers } from "ethers";
 import { hubPoolChainId, ChainId, infuraId } from "./constants";
 
-function getProviderUrl(chainId: number) {
-  const overrideUrl = process.env[`REACT_APP_CHAIN_${chainId}_PROVIDER_URL`];
-  if (overrideUrl) {
-    return overrideUrl;
-  }
-
+function getInfuraProviderUrl(chainId: number) {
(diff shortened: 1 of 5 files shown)


nicholaspai (approved):

src/utils/providers.ts
```diff
@@ -1,35 +1,61 @@
...
+import { hubPoolChainId, ChainId, infuraId } from "./constants";
 
-const PolygonProviderUrl =
-  process.env.REACT_APP_CHAIN_137_PROVIDER_URL ||
-  `https://polygon-mainnet.infura.io/v3/${infuraId}`;
+function getInfuraProviderUrl(chainId: number) {
+  const infuraUrl = new providers.InfuraProvider(chainId, infuraId).connection
+    .url;
+  return infuraUrl;
+}
```
amateima, line 8: It's a bit weird to instantiate a provider just to get the URL out of it

amateima (approved):

src/utils/providers.ts
dohaki: Yea...  But seemed like a better option than hardcoding all Infura URLs to me.

PR #802, opened 2023-08-07
feat: zk sync era support

dohaki:
This adds support for bridging from/to zkSync Era mainnet and testnet.

Closes ACX-1335 and ACX-1350

**NOTE: Don't merge until testing is complete. Code can already be reviewed though.**


james-a-morris (approved):
LGTM - left some comments. My only worry is that the SVG's aren't ACX teal - but the CSS handles this fine.

api/_utils.ts
```diff
@@ -173,7 +173,7 @@ export const resolveVercelEndpoint = () => {
       return `https://${url}`;
     case "development":
     default:
-      return `http://localhost:3000`;
+      return `http://127.0.0.1:3000`;
```
james-a-morris, line 176: Is there a reason this needed to change?

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
       getLogger(),
       getGasMarkup(42161)
     ),
+  324: () =>
+    new sdk.relayFeeCalculator.ZkSyncQueries(
```
nicholaspai, line 368: @pxrl can you sanity check this is the correct way to construct this?

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
...
+      undefined,
+      REACT_APP_COINGECKO_PRO_API_KEY,
+      getLogger(),
+      getGasMarkup(324)
+    ),
   // testnets
   5: () =>
     new sdk.relayFeeCalculator.EthereumQueries(
-      getProvider(5),
+      getProvider(1),
```
nicholaspai, line 381: This seems like an error?

src/utils/constants.ts
```diff
@@ -202,6 +208,21 @@ export const chainInfoList: ChainInfoList = [
...
+    logoURI: zkSyncLogo,
+    grayscaleLogoURI: zkSyncLogoGrayscale,
+    rpcUrl: "https://mainnet.era.zksync.io",
+    explorerUrl: "https://explorer.zksync.io",
+    constructExplorerLink: defaultConstructExplorerLink(
+      "https://explorer.zksync.io"
+    ),
+    nativeCurrencySymbol: "ETH",
+    pollingInterval: defaultBlockPollingInterval,
+    earliestBlock: 10352565,
```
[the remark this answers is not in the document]
nicholaspai: Agreed i'm confused what this block is

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
...
+      undefined,
       undefined,
-      "0x063fFa6C9748e3f0b9bA8ee3bbbCEe98d92651f7",
       undefined,
       undefined,
       REACT_APP_COINGECKO_PRO_API_KEY,
       getLogger(),
-      getGasMarkup(5)
+      getGasMarkup(1)
```
james-a-morris, line 388: Similar to above @nicholaspai

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
...
-      getProvider(421613),
+    new sdk.relayFeeCalculator.ArbitrumQueries(
+      getProvider(42161),
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      REACT_APP_COINGECKO_PRO_API_KEY,
+      getLogger(),
+      getGasMarkup(42161)
```
james-a-morris, line 399: These chain IDs do not match

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
...
+    new sdk.relayFeeCalculator.ZkSyncQueries(
+      getProvider(324),
       undefined,
       undefined,
       undefined,
       undefined,
       REACT_APP_COINGECKO_PRO_API_KEY,
       getLogger(),
-      getGasMarkup(421613)
+      getGasMarkup(324)
```
james-a-morris, line 410: This is the wrong chain ID unless we are looking to mock 324

api/_utils.ts
pxrl: This seems correct to me :+1:

james-a-morris (changes requested):

api/_utils.ts
```diff
@@ -364,28 +364,50 @@ export const queries: Record<number, () => QueryBase> = {
...
+      undefined,
+      undefined,
+      undefined,
+      REACT_APP_COINGECKO_PRO_API_KEY,
+      getLogger(),
+      getGasMarkup(42161)
+    ),
+  280: () =>
+    new sdk.relayFeeCalculator.ZkSyncQueries(
+      getProvider(324),
```
pxrl, line 403: As discussed in the call yesterday, this seems intuitively wrong and I suspect it's in the class of "works by accident"...but it's a pre-existing pattern so I think we're OK to roll with it.

api/_utils.ts
dohaki: Yea, for some reason `localhost` doesn't work correctly with the new `vite` setup.

api/_utils.ts
dohaki: I changed these queries because they weren't working previously. Because the dummy "from" address that's used to estimate gas costs isn't loaded on these test chains I guess. @pxrl mentioned that we use a similar approach in the relayer, so I implemented the workaround here.

Maybe I should add a comment that explains this?

api/_utils.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/802/files#r1288744226

api/_utils.ts
dohaki: Yea, I used this workaround as explained in the call.

src/utils/constants.ts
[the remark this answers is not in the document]
dohaki: AFAICT these values are the earliest block we consider or care about on the respective chains. In this case, this would be the block of the zkSync SpokePool deployment. But I noticed that we no longer use that in the code. So I will remove this to prevent further confusion.

src/utils/constants.ts
```diff
@@ -243,6 +264,21 @@ export const chainInfoList: ChainInfoList = [
...
+    logoURI: zkSyncLogo,
+    grayscaleLogoURI: zkSyncLogoGrayscale,
+    rpcUrl: "https://testnet.era.zksync.dev",
+    explorerUrl: "https://goerli.explorer.zksync.io",
+    constructExplorerLink: defaultConstructExplorerLink(
+      "https://goerli.explorer.zksync.io"
+    ),
+    nativeCurrencySymbol: "ETH",
+    pollingInterval: defaultBlockPollingInterval,
+    earliestBlock: 10263886,
```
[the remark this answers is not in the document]
dohaki: Removed here https://github.com/across-protocol/frontend-v2/pull/802/commits/a055917bb7581a3fed6a1751d4dfa8c63c78b733

amateima (approved):

nicholaspai (approved):

PR #806, opened 2023-08-14
chore: remove duplicate available routes

james-a-morris:
Available routes currently contains a duplicate set of available routes from ZKSync to Optimism. This PR removes that.


dohaki (approved):
Good catch

PR #808, opened 2023-08-15
fix: only show ETH as receive token if dst not polygon

dohaki:
Fixes ACX-1460


pxrl (approved):
LGTM. Tooltip doesn't work, but I presume that's to be fixed separately.

img

PR #810, opened 2023-08-15
fix: add tooltip id to eth/weth info icon

dohaki:
Fixes ACX-1462


pxrl (approved):
LGTM - have confirmed this works in the preview link.

PR #813, opened 2023-08-16
fix: arbitrum logo animation

dohaki:
The animation on the deposit confirmation screen was broken for Abitrum


james-a-morris (approved):
LGTM - tested and works on prod preview link

PR #815, opened 2023-08-17
feat: base support

dohaki:
This adds support for Base.

I have added the chain id  `8453` to the env var `REACT_APP_DISABLED_CHAINS_FOR_AVAILABLE_ROUTES` so that it will be disabled on the production frontend and /available-routes endpoint. We can therefore merge this PR after it is approved.


src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
pxrl, line 1: Is this manually maintained or auto-generated? If it's manual then we probably want to call it out as something that needs careful review.

src/utils/constants.ts
```diff
@@ -40,11 +35,12 @@ export enum ChainId {
...
   ZK_SYNC = 324,
+  BASE = 8453,
   // testnets
   ARBITRUM_GOERLI = 421613,
   ZK_SYNC_GOERLI = 280,
+  BASE_GOERLI = 84531,
   GOERLI = 5,
-  // Polygon testnet
   MUMBAI = 80001,
 }
```
pxrl, line 45: Thoughts on using the CHAIN_IDs.<network> constants here?

src/utils/constants.ts
```diff
@@ -122,20 +118,21 @@ export const configStoreAddresses: Record<ChainId, string> = {
...
   [ChainId.ZK_SYNC]: ethers.constants.AddressZero,
+  [ChainId.BASE]: ethers.constants.AddressZero,
   [ChainId.ARBITRUM_GOERLI]: ethers.constants.AddressZero,
   [ChainId.GOERLI]: ethers.utils.getAddress(
     "0x3215e3C91f87081757d0c41EF0CB77738123Be83"
   ),
   [ChainId.MUMBAI]: ethers.constants.AddressZero,
   [ChainId.ZK_SYNC_GOERLI]: ethers.constants.AddressZero,
+  [ChainId.BASE_GOERLI]: ethers.constants.AddressZero,
 };
```
pxrl, line 129: For future, maybe we should have an array HUB_CHAIN_IDs -> i.e. chain IDs where we have a "Hub" deployment (HubPool + ConfigStore). Then we could probably skip the need to specify zero addresses for chains that don't have a deployment.

pxrl (commented):
LGTM - would be nice to know how closely we need to inspect the advertised routes though.

src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
dohaki: Good point. It is currently manually generated. So a careful review of these addresses would be highly appreciated.

src/utils/constants.ts
dohaki: Yea makes sense. I changed it here https://github.com/across-protocol/frontend-v2/pull/815/commits/d5866ea240bcf99157f8e1e3b21e2ab738caa982.

src/utils/constants.ts
dohaki: Thanks for pointing out that these mapped zero addresses are not ideal. I had a closer look where we actually use this constant and could replace it actually with our `getDeployedAddress` helper from the SDK. See https://github.com/across-protocol/frontend-v2/pull/815/commits/a69722d88519f1c8e68f6fe5577e97ac2549ec4b

src/utils/constants.ts
```diff
@@ -400,9 +394,12 @@ export function isSupportedChainId(chainId: number): chainId is ChainId {
...
   chainId: ChainId = hubPoolChainId
 ): string {
-  const configStoreAddress = configStoreAddresses[chainId];
+  const configStoreAddress = utils.getDeployedAddress(
+    "AcrossConfigStore",
+    chainId
+  );
   assert(
-    configStoreAddress !== AddressZero,
+    Boolean(configStoreAddress),
```
james-a-morris, line 402: Is this asserting that the `configStoreAddress` exists? if so, we may want to use `sdk.utils.isDefined(configStoreAddress)`
```suggestion
    sdk.utils.isDefined(configStoreAddress),
```

james-a-morris (approved):
Looks great 🔥 left one nit but otherwise the FE looks great and I look forward to drilling down to test this further from a relayer perspective

api/_utils.ts
```diff
@@ -310,109 +311,121 @@ export const getHubPoolClient = () => {
...
+      getProvider(CHAIN_IDS.BASE_GOERLI),
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      REACT_APP_COINGECKO_PRO_API_KEY,
+      getLogger(),
+      getGasMarkup(CHAIN_IDS.BASE_GOERLI)
     ),
 };
```
james-a-morris, line 430: Nit: I feel as though there's definitely a way to refactor this so that there's less code to write this out.

src/utils/constants.ts
dohaki: Yea, this should assert that the `configStoreAddress` is truthy. Hmm, what exactly does `isDefined` check though? Why is using that helper better than just casting to a boolean?

dohaki: Actually, I think checking whether it is a valid address makes the most sense. Changed here https://github.com/across-protocol/frontend-v2/pull/815/commits/38f86af85ce747f29cfee58fdab366f7d5a117d9

PR #816, opened 2023-08-18
feat: improve token and route selection

dohaki:
This PR fixes ACX-1429 and in general, improves the UX of token and route selection when bridging. The key UX improvements are:
- Improved behavior when selecting unavailable tokens or chains
  - The function `findNextBestRoute` ensures that the best matching route is selected
- Improved handling of USDC/USDC.e
  - The UI makes it clearer which token is sent/received
  
From a code perspective, this PR refactors quite a lot. Mainly moving logic from the `useBridge` hook into separate hooks. The umbrella `useBridge` hook grew in complexity over time and the implementation was hard to follow. I, therefore, tried to separate functional components into separate hooks. I also tried to reduce the extensive usage of `useEffect` in order to make things easier to reason about.

I know that this PR has quite a large diff and may be hard to review. If that is the case, I can try to open separate PRs for the refactorings. Even though, I still want to have everything on a single branch, so that we can test the UX improvements via a preview link.


src/components/Selector/Selector.tsx
```diff
@@ -79,14 +81,16 @@ const Selector = <ElementValue,>({
         <ElementRowWrapper enableScroll={elements.length > 7}>
           {elements.map((element, idx) => (
             <ElementRow
-              key={idx}
+              key={String(element.value)}
```
james-a-morris, line 84: Out of curiosity, if this breaks here should we not be using `idx` elsewhere?

src/views/Bridge/components/AmountInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  validationError,
+}: Props) {
+  const [displayBalance, setDisplayBalance] = useState(false);
+  const [didFocus, setDidFocus] = useState(false);
+
+  useEffect(() => {
+    setDidFocus(false);
+  }, [selectedRoute.fromTokenSymbol]);
+
+  const token = getToken(selectedRoute.fromTokenSymbol);
```
james-a-morris, line 36: Should we check if this token can be resolved?

src/views/Bridge/components/AmountInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+      </AmountWrapper>
+      {didFocus && !isAmountValid && (
+        <BridgeInputErrorAlert>
+          {validationError === "insufficientBalance" &&
+            "Insufficient balance to process this transfer."}
+          {validationError === "insufficientLiquidity" &&
+            "Insufficient bridge liquidity to process this transfer."}
+          {validationError === "invalid" &&
+            "Only positive numbers are allowed as an input."}
+          {validationError === "requestTooLow" &&
```
james-a-morris, line 85: Should we put these validation error types in an ENUM?

src/views/Bridge/components/AmountInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  align-items: flex-start;
+  padding: 0px;
+`;
+
+const AmountInnerInput = styled.input<IValidInput>`
+  font-family: "Barlow";
+  font-style: normal;
```
james-a-morris, line 180: These are defined globally I believe

src/views/Bridge/components/AmountInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  font-weight: 400;
+  font-size: 18px;
+  line-height: 26px;
+  color: #e0f3ff;
+
+  color: ${({ valid }) => (!valid ? "#f96c6c" : "#e0f3ff")};
```
james-a-morris, line 186: Nit
```suggestion
  color: ${({ valid }) => (valid ?  "#e0f3ff" : "#f96c6c")};
```

src/views/Bridge/components/TokenSelector.tsx
```diff
@@ -0,0 +1,128 @@
...
+            <Text size="md" color="grey-400">
+              {formatUnits(balances[i] ?? BigNumber.from(0), t.decimals)}
+            </Text>
+          ) : undefined,
+      }))}
+      displayElement={
+        <CoinIconTextWrapper>
+          <CoinIcon src={selectedToken.logoURI} />
+          <Text size="md" color="white-100">
+            {selectedToken.displaySymbol || selectedToken.symbol.toUpperCase()}
```
james-a-morris, line 88: We apply `toUpperCase()` or `toLowerCase()` extensively throughout the code. I wonder if we should append to `<Text >` an attribute for casing like `<Text casing={"upper" | "lower" | "normal" | "capital"} />` or if we should create util functions like `upper(str)`, `lower(str)`, `capital()`

james-a-morris (commented):
Code looks great - I have a few nits. Because this is a large refactor, do we have an exhaustive list of UI tests that should be tested against to ensure nothing has regressed?

src/components/Selector/Selector.tsx
dohaki: With our current usage, this doesn't necessarily break because we don't re-sort the token or chain options. But as soon as we would implement it, it would. AFAIK it is general best practice to not use the `index` of an item as the `key` prop. See the docs for details https://react.dev/learn/rendering-lists#rules-of-keys

src/views/Bridge/components/AmountInput.tsx
dohaki: Ideally yes... But we use this pattern in a lot of places and just assume that these tokens are known to us. I would tackle this in a separate PR if you don't mind.

src/views/Bridge/components/AmountInput.tsx
dohaki: Yes makes sense. Changed here https://github.com/across-protocol/frontend-v2/pull/816/commits/321856d8ffa91911ac5b823ab2f87816673770d6

src/views/Bridge/components/AmountInput.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/816/commits/c45414af38d3354559cb6a374e696c3f992f31bd

src/views/Bridge/components/AmountInput.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/816/commits/c45414af38d3354559cb6a374e696c3f992f31bd

src/views/Bridge/components/TokenSelector.tsx
dohaki: Yea, something like `<Text casing={"upper" | "lower" | "normal" | "capital"} />` could be handy. But not sure why a util function like `upper(str)` is better than using the native `str.toUpperCase()`?

In any case, we should tackle this in a separate PR for the whole codebase.

dohaki:
> Because this is a large refactor, do we have an exhaustive list of UI tests that should be tested against to ensure nothing has regressed?

@james-a-morris good point! Ideally, we would have e2e tests that ensure nothing has regressed 😅 But maybe this list can be a good starting point and be translated to cypress tests at some point.

- Input behavior
  - [ ] Max balance button works
  - [ ] Input validation works for invalid numbers, too small/large amount, etc.
  - [ ] Balance is shown and updated if wallet connected
- Token selector
  - [ ] Token balances are shown
  - [ ] Supported tokens are sorted on top
  - [ ] Selecting token works
  - [ ] Selecting unsupported token changes from/to chain
- From/To chain selectors
  - [ ] All chains can be selected
  - [ ] Selecting a chain results in sensible route selection
  - [ ] Switch button works
- Custom to address input
  - [ ] Able to change receiving address
- Fees table
  - [ ] Updates periodically
  - [ ] Updates when user enters amount
  - [ ] Does not update if transaction in progress
  - WETH/ETH behavior 
    - [ ] If destination chain is Polygon and bridge token is ETH, info on receiving WETH is shown
    - [ ] If receiving address is contract and bridge token is ETH, info on receiving WETH is shown
    - [ ] If receiving address is EOA and bridge token is WETH, info on receiving ETH is shown
   - USDC/USDC.e behavior
    - [ ] If destination chain is Arbitrum and bridge token is USDC, info on receiving USDC.e is shown
    - [ ] If source chain is Arbitrum and bridge token is USDC.e, info on receiving USDC is shown
- Deposit confirmation 
  - [ ] Confirming deposit results in deposit confirmation screen
  - [ ] During pending tx, quotes are not updated
  - [ ] Initiating a new tx, clears input

src/components/Selector/Selector.tsx
```diff
@@ -26,6 +26,7 @@ export type SelectorPropType<Value> = {
   displayElement?: JSX.Element;
   disabled?: boolean;
   "data-cy"?: string;
+  allowSelectDisabled?: boolean;
```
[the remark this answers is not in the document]
dohaki: Do you have any suggestions? This flag basically should allow disabled options to be selectable 😅 So I thought this name fits. But yea could be confusing without context I guess

src/hooks/useBridgeLimits.ts
```diff
@@ -18,18 +18,26 @@ export interface BridgeLimits {
...
-  token?: string,
+  tokenSymbol?: string,
   fromChainId?: ChainId,
   toChainId?: ChainId
 ) {
   const enabledQuery =
-    token !== undefined && toChainId !== undefined && fromChainId !== undefined;
+    tokenSymbol !== undefined &&
+    toChainId !== undefined &&
+    fromChainId !== undefined;
```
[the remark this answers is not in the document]
dohaki: Yea, that could be handy. I added a helper here https://github.com/across-protocol/frontend-v2/pull/816/commits/ea7a0b86e648d9a7e9765ed9480844ae68988fca. The implementation currently does not allow multiple arguments because I wanted to use type predicates. If you know how to achieve this with multiple values, let me know.

src/hooks/useBridgeLimits.ts
```diff
@@ -18,18 +18,26 @@ export interface BridgeLimits {
...
   const queryKey = enabledQuery
-    ? bridgeLimitsQueryKey(token, fromChainId, toChainId)
+    ? bridgeLimitsQueryKey(tokenSymbol, fromChainId, toChainId)
     : "DISABLED_BRIDGE_LIMITS_QUERY";
   const { data: limits, ...delegated } = useQuery(
     queryKey,
-    async () => getApiEndpoint().limits(token!, fromChainId!, toChainId!),
+    () => {
+      if (!tokenSymbol || !fromChainId || !toChainId) {
```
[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/frontend-v2/pull/816/commits/9011208e5c2cf6c9e42a6f2632bc209b307ce92b

src/views/Bridge/components/AmountInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+  &::placeholder {
+    color: #9daab3;
+  }
+
+  overflow-x: hidden;
+
+  @media ${QUERIESV2.sm.andDown} {
+    font-size: 16px;
+    line-height: 20px;
+  }
```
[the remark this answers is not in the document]
dohaki: Which xs query are you referring to? Most components and styles were not modified in this PR. Only extracted into separate components

src/views/Bridge/components/BridgeForm.tsx
```diff
@@ -1,214 +1,132 @@
...
+  amountInput: string;
+  amountToBridge?: BigNumber;
+  toAccount?: ToAccount;
+
+  onChangeAmountInput: (input: string) => void;
+  onClickMaxBalance: () => void;
+  onSelectToken: (token: string) => void;
+  onSelectFromChain: (chainId: number) => void;
+  onSelectToChain: (chainId: number) => void;
+  onClickQuickSwap: () => void;
```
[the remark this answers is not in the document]
dohaki: Makes sense! https://github.com/across-protocol/frontend-v2/pull/816/commits/2f9a9398dbcce48896cc8140c996618a9023529d

src/views/Bridge/components/TokenSelector.tsx
james-a-morris: Totally makes sense - I've added the initial precursor here: https://github.com/across-protocol/frontend-v2/pull/823

src/hooks/useBridgeLimits.ts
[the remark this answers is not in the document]
james-a-morris: How about an additional util that uses `isDefined<unknown>()` such as

```
export function areDefined(...array: unknown[]): boolean {
    return array.every(isDefined);
}
```

src/views/Bridge/components/AmountInput.tsx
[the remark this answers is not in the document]
james-a-morris: Ah I see - okay sounds good 🙂

james-a-morris (approved):
The code looks good to me! I left one additional comment but you can feel free to make it actionable or not.

This is great work

PR #822, opened 2023-08-26
fix: failed balance fetch usdbc

dohaki:
Fixes ACX-1498

It was a bit hard to track down the issue due to the side-effect of reporting token balances in the `getBalance` function. I decided to move it out therefore.

But the root cause was that the `getTokenBalance` function fetched the symbol from the erc20 contract and tried to resolve it. Because we handle USDbC on Base the same as USDC, this led to errors.


james-a-morris (approved):
LGTM - I left a comment on whether or not we should add retry logic. However, this is up to your discretion 🙂

PR #826, opened 2023-08-29
fix: add balancer token computation to API

mrice32:
**Overview**

This computes the price for a balancer v2 token correctly in the API. Previously, this would return 0 in the case of the ACX/wstETH bal pool.

Currently, this PR is in draft, as it has not been tested.

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -124,3 +124,59 @@ export const ENABLED_POOLS_UNDERLYING_TOKENS = [
 ];
 
 export const SECONDS_PER_YEAR = 365 * 24 * 60 * 60;
+
+export const MINIMAL_BALANCER_V2_POOL_ABI = [
+  {
+    inputs: [],
+    name: "getPoolId",
+    outputs: [{ internalType: "bytes32", name: "", type: "bytes32" }],
+    stateMutability: "view",
+    type: "function",
+  },
+  {
+    inputs: [],
+    name: "getVault",
+    outputs: [{ internalType: "contract IVault", name: "", type: "address" }],
+    stateMutability: "view",
+    type: "function",
+  },
+];
+
+export const MINIMAL_BALANCER_V2_VAULT_ABI = [
+  {
+    inputs: [{ internalType: "bytes32", name: "poolId", type: "bytes32" }],
+    name: "getPoolTokens",
+    outputs: [
+      { internalType: "contract IERC20[]", name: "tokens", type: "address[]" },
+      { internalType: "uint256[]", name: "balances", type: "uint256[]" },
+      { internalType: "uint256", name: "lastChangeBlock", type: "uint256" },
+    ],
+    stateMutability: "view",
+    type: "function",
+  },
+];
+
+export const MINIMAL_MULTICALL3_ABI = [
+  {
+    inputs: [
+      {
+        components: [
+          { internalType: "address", name: "target", type: "address" },
+          { internalType: "bytes", name: "callData", type: "bytes" },
+        ],
+        internalType: "struct Multicall3.Call[]",
+        name: "calls",
+        type: "tuple[]",
+      },
+    ],
+    name: "aggregate",
+    outputs: [
+      { internalType: "uint256", name: "blockNumber", type: "uint256" },
+      { internalType: "bytes[]", name: "returnData", type: "bytes[]" },
+    ],
+    stateMutability: "payable",
+    type: "function",
+  },
+];
+
+export const MULTICALL3_ADDRESS = "0xcA11bde05977b3631167028862bE2a173976CA11";
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -22,6 +22,10 @@ import {
   TOKEN_SYMBOLS_MAP,
   CHAIN_IDS,
   SECONDS_PER_YEAR,
+  MINIMAL_BALANCER_V2_POOL_ABI,
+  MINIMAL_BALANCER_V2_VAULT_ABI,
+  MINIMAL_MULTICALL3_ABI,
+  MULTICALL3_ADDRESS,
 } from "./_constants";
 import { StaticJsonRpcProvider } from "@ethersproject/providers";
 import QueryBase from "@across-protocol/sdk-v2/dist/relayFeeCalculator/chain-queries/baseQuery";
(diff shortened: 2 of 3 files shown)


api/_constants.ts
```diff
@@ -124,3 +124,59 @@ export const ENABLED_POOLS_UNDERLYING_TOKENS = [
...
+    outputs: [
+      { internalType: "uint256", name: "blockNumber", type: "uint256" },
+      { internalType: "bytes[]", name: "returnData", type: "bytes[]" },
+    ],
+    stateMutability: "payable",
+    type: "function",
+  },
+];
+
+export const MULTICALL3_ADDRESS = "0xcA11bde05977b3631167028862bE2a173976CA11";
```
james-a-morris, line 182: We may want to put this in a separate file so that we don't always load these ABIs if they're not necessary

api/_utils.ts
```diff
@@ -1022,3 +1026,107 @@ export function getBaseRewardsApr(
...
+    MULTICALL3_ADDRESS,
+    MINIMAL_MULTICALL3_ABI,
+    provider
+  );
+  const inputs = calls.map(({ contract, functionName, args }) => ({
+    target: contract.address,
+    callData: contract.interface.encodeFunctionData(functionName, args),
+  }));
+
+  const [_, results] = await multicall3.callStatic.aggregate(inputs);
```
james-a-morris, line 1055:
```suggestion
  const [, results] = await multicall3.callStatic.aggregate(inputs);
```

api/_utils.ts
```diff
@@ -1022,3 +1026,107 @@ export function getBaseRewardsApr(
...
+  );
+
+  const totalValue = tokenValues.reduce((a, b) => a + b, 0);
+
+  // Balancer v2 tokens have 18 decimals.
+  const floatTotalSupply = parseFloat(
+    ethers.utils.formatUnits(totalSupply, 18)
+  );
+
+  return totalValue / floatTotalSupply;
```
james-a-morris, line 1131: Should we do these computations with BigNumbers?

james-a-morris (commented):
Left some comments. It's a shame we can't exploit any additional speedups through parallelization. However, after reviewing the code, it's clear this is both a viable solution and one that lends to readability.

api/_utils.ts
mrice32: We can, but the previous computation used floats, so my intention was to avoid changing behavior. Do you think there would be any precision loss that we should be worried about? Would the downstream application be affected by precision loss?

api/_constants.ts
mrice32: Shouldn't they be tree-shaken if not imported in the process that's using them? I can move them to another file, though.

mrice32: Done.

api/_utils.ts
```diff
@@ -151,26 +160,26 @@ let logger: LoggingUtility;
...
+  //       private_key: GOOGLE_SERVICE_ACCOUNT.private_key,
+  //     },
+  //   }).log(VERCEL_ENV ?? "", { removeCircular: true });
+  //   logger = {
+  //     debug: (data: LogType) => log(gcpLogger, "DEBUG", data),
+  //     info: (data: LogType) => log(gcpLogger, "INFO", data),
+  //     warn: (data: LogType) => log(gcpLogger, "WARN", data),
+  //     error: (data: LogType) => log(gcpLogger, "ERROR", data),
+  //   };
+  // }
```
james-a-morris, line 182: Is this meant for prod?

api/_utils.ts
```diff
@@ -1048,3 +1057,109 @@ export function getBaseRewardsApr(
...
+    MINIMAL_MULTICALL3_ABI,
+    provider
+  );
+  const inputs = calls.map(({ contract, functionName, args }) => ({
+    target: contract.address,
+    callData: contract.interface.encodeFunctionData(functionName, args),
+  }));
+
+  const [, results] = await (multicall3.callStatic.aggregate(inputs) as Promise<
+    [BigNumber, string[]]
```
james-a-morris, line 1087: Nit: does `aggregate` have any generics that corral its output type? i.e. `aggregate<[BigNumber, string[]]>`

api/_utils.ts
```diff
@@ -1048,3 +1057,109 @@ export function getBaseRewardsApr(
...
+    provider,
+    [
+      {
+        contract: pool,
+        functionName: "getVault",
+      },
+      {
+        contract: pool,
+        functionName: "getPoolId",
+      },
```
james-a-morris, line 1124: Wouldn't we already have these values given a token address for our specific use case? If we assumed this was in a lookup table, we can parallelize the calls following this.

api/_utils.ts
james-a-morris: Fair enough! My only follow-up would be the maximum precision. We should limit the float result to a maximum decimal length of `18`. The FE expects `18` decimals when dealing with price-equivalents. 

i.e. `ethers.utils.parseUnits("1.123", 2)` <- throws a `fractional component exceeds decimals` error.

james-a-morris (commented):
LGTM - left some comments

api/_utils.ts
mrice32: It's just a normal ethers Contract type. I don't think there are any built-in generics to force the output type. But if you know of a better way to do this, I'm all ears!!

api/_utils.ts
mrice32: No! Nice catch. This code always breaks for me locally, so I have to comment it. Thank you for calling this out!

api/_utils.ts
mrice32: Fixed! PTAL

api/_utils.ts
```diff
@@ -151,7 +158,7 @@ let logger: LoggingUtility;
  * @returns A valid Logging utility that can be used throughout the runtime
  */
 export const getLogger = (): LoggingUtility => {
-  // Use the default logger which logs to console if no GCP service account is configured.
+  // // Use the default logger which logs to console if no GCP service account is configured.
```
james-a-morris, line 161:
```suggestion
  // Use the default logger which logs to console if no GCP service account is configured.
```

james-a-morris (approved):
Minor nit, otherwise LGTM

PR #829, opened 2023-09-04
improve: condense disabled chains env variables

james-a-morris:
The impetus for this PR is to reduce the number of environment variables that dictate which chains are available from two to one.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -121,5 +121,24 @@ DISABLED_ROUTE_TOKENS=
 # e.g. "1,10,137,288,42161"
 REACT_APP_DISABLED_CHAINS=
 
+# This comma-separated array represents the array of chains that are available for
+# the frontend to display in the /available-routes and /token-list API endpoints.
+# This list is whitelisted in that certain aggregators use this list to
+# which routes are available to serve bridge transactions by across. To ensure
+# that the available routes are explicitely defined, we use an opt-in approach
+# to adding chains to this list. If a chain is not in this list, it will not
+# be presented in the /available-routes and /token-list endpoints.
+#
+# NOTE:   If no chains are specified, then we default to the chains that are
+#         specified in the routes.json file.
+#
+# NOTE:   If a chain specified in this list is also specified in the
+#         `REACT_APP_DISABLED_CHAINS` list, then it will not be presented in the
+#         /available-routes and /token-list endpoints.
+#
+# USAGE: "1,10,137,288,42161"
+#
+REACT_APP_WHITELISTED_AVAILABLE_CHAINS=
+
 # Comma-separated list of wallet addresses to block from UI
 REACT_APP_WALLET_BLACKLIST=
--- a/.env.example
+++ b/.env.example
@@ -58,19 +58,19 @@ REACT_APP_DEFAULT_BLOCK_POLLING_INTERVAL_S=
 REACT_APP_FLAT_RELAY_CAPITAL_FEE=
 
 # Show rewards banner warning on /rewards page if set to `true`.
-REACT_APP_REWARDS_BANNER_WARNING= 
+REACT_APP_REWARDS_BANNER_WARNING=
 
 # Define Hardhat Infura ID for testing
-HARDHAT_INFURA_ID= 
+HARDHAT_INFURA_ID=
 
 # Define a configurable message to display to users in environment.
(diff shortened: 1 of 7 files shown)


.env.example
```diff
@@ -121,10 +121,5 @@ DISABLED_ROUTE_TOKENS=
 # e.g. "1,10,137,288,42161"
 REACT_APP_DISABLED_CHAINS=
 
-# Comma-separated list of chain ids to disable in the UI and in the following API handlers:
-# - /available-routes
-# - /token-list
-REACT_APP_DISABLED_CHAINS_FOR_AVAILABLE_ROUTES=
```
dohaki, line 127: The reason we had this second env var, was because we wanted to be able to enable certain chains for endpoints like `/suggested-feed` and `/limits` but disable them at the same time for other endpoints like `/available-routes`. Is this case something we won't need in the future or what is the motivation for this change?

dohaki (changes requested):
Refactoring looks good but left a question on the motivation behind this change

.env.example
james-a-morris: Thoughts on making this opt-in? I've added a change that includes the second variable but considers it a whitelist of chains we make visible via the available routes/token-list API endpoints.

I think have two thoughts that spurred my want for this change:

- Since the goal of this second var is to limit exposure of routes to aggregators - it may be beneficial to make it opt-in. 
- The naming conventions of the two are very similar, and we ran into an issue where the naming caused confusion & extra work.

dohaki: Isn't it already opt-in? We do not require a list of chains in the env var.

I agree that the previous name is a bit confusing but not sure if specifying an explicit list is also necessary.

Let's say, we add a new chain and want to test, then we need to add/remove all enabled chains from the new env var instead of adding/removing just the new chain id. So this could be more prone to errors because you need to keep track of multiple chain ids in contrast to a single one. 

I could also be a bit nitpicking here... Let me know what you think

mrice32: Previously, would putting something in disabled chains disable it everywhere (including the routes)? Or would you have to put the disabled chain in both env arrays? If putting it in disabled chains was all that was needed for a full shutdown of a chain, maybe all we need is to make sure that variable is well documented for that purpose?

Not super strongly opinionated, just want this to be as clear as possible for use in cases like zkSync where fast shutdown is needed.

mrice32 (commented):
https://github.com/across-protocol/frontend-v2/pull/829#discussion_r1316677860

.env.example
dohaki: > Previously, would putting something in disabled chains disable it everywhere (including the routes)?

Yes. So we already had the env var `REACT_APP_DISBALED_CHAINS` which would disable the respective chains everywhere. For disabling a chain only on the `/available-routes` endpoint, you need to use the env var `REACT_APP_DISABLED_CHAINS_FOR_AVAILABLE_ROUTES`.

src/utils/constants.ts
```diff
@@ -678,10 +678,6 @@ export const disabledChainIds = (
   process.env.REACT_APP_DISABLED_CHAINS || ""
 ).split(",");
 
-export const disabledChainIdsForAvailableRoutes = (
-  process.env.REACT_APP_DISABLED_CHAINS_FOR_AVAILABLE_ROUTES || ""
-).split(",");
```
dohaki, line 683: I think we still need to keep track of the whitelisted available chains if we go with your approach.

src/views/Transactions/utils.ts
```diff
@@ -17,9 +13,7 @@ export function getSupportedTxTuples(
       // Do not show transactions for disabled chains
       if (
         [tx.sourceChainId, tx.destinationChainId].some((chainId) =>
-          [...disabledChainIds, ...disabledChainIdsForAvailableRoutes].includes(
-            String(chainId)
-          )
```
dohaki, line 22: We need to consider only the whitelisted chains here.

.env.example
dohaki: After thinking a bit more about it, @james-a-morris approach seems actually better. Because we only need to update the whitelist env var only once when we want to enable the new chain. Instead of updating twice (adding the to be disabled chain and removing it again).

dohaki:
@james-a-morris I left only two comments on the transactions page but if these are addressed this looks good to me

.env.example
james-a-morris: Sounds good to me!

PR #839, opened 2023-09-05
feat: add balancer to base routes

james-a-morris:


dohaki (approved):

mrice32:
As discussed IRL, we're going to hold off on this until confusion over the correct address subsides!

nicholaspai (changes requested):
We might need to import contracts-v2 version 2.4.4 which includes the updated TOKEN_SYMBOLS_MAPPING with the new BAL address

But otherwise this LGTM when we want to enable BAL on Base

nicholaspai (approved):

PR #841, opened 2023-09-07
fix: use correct pool APY

james-a-morris:
Per a meeting today with leadership, we should be using `swap fees` + `tokenAprs` to compute the pool APY.


dohaki (approved):
LGTM

PR #843, opened 2023-09-19
feat: bump react-pro-sidebar and refactor

dohaki:
Updates sidebar lib hoping that this will fix https://linear.app/uma/issue/ACX-1275/landing-page-menu-button-doesnt-respond-on-click.

Initial tests confirm a performance boost

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "react-dom": "^17.0.2",
     "react-feather": "^2.0.9",
     "react-lottie": "^1.2.3",
-    "react-pro-sidebar": "^1.1.0-alpha.1",
+    "react-pro-sidebar": "1.0.0",
     "react-query": "^3.34.16",
     "react-router-dom": "^5.3.0",
     "react-router-sitemap": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "react-dom": "^17.0.2",
     "react-feather": "^2.0.9",
     "react-lottie": "^1.2.3",
-    "react-pro-sidebar": "^0.7.1",
+    "react-pro-sidebar": "^1.1.0-alpha.1",
     "react-query": "^3.34.16",
     "react-router-dom": "^5.3.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "react-dom": "^17.0.2",
     "react-feather": "^2.0.9",
     "react-lottie": "^1.2.3",
-    "react-pro-sidebar": "^0.7.1",
+    "react-pro-sidebar": "^1.1.0-alpha.1",
```
james-a-morris, line 41: Should we be bumping to an alpha build for prod?

dohaki: But yea I guess pinning to `1.0.0` makes sense.

dohaki: Ok, I would actually prefer the 1.1.0 alpha version 😅 It has a cleaner API and doesn't require us to wrap the app within another provider. Wdyt?

james-a-morris: Yeah no worries!

src/components/Sidebar/Sidebar.tsx
```diff
@@ -1,132 +1,166 @@
...
-              }
+    <ReactProSidebar
+      onBackdropClick={onClickOverlay}
+      toggled={openSidebar}
+      breakPoint="all"
+      width={sidebarWidth}
+      rootStyles={{
+        zIndex: "3000 !important",
+        // @ts-ignore
+        direction: "ltr !important", // hack to display on the right side
```
james-a-morris, line 70: Can we wrap this with a styles wrapper?

james-a-morris (approved):
LGTM - 2 nits. I ran it in preview and it looks good

src/components/Sidebar/Sidebar.tsx
dohaki: done https://github.com/across-protocol/frontend-v2/pull/843/commits/97870c4cbe8a38fa0bb4ac1edbb7c667f0390a78

PR #844, opened 2023-09-19
fix(bridge): allow from/to/token search params on bridge page

dohaki:
Allows setting query params on the /bridge page for pre-selected routes.

### DIFF
--- a/src/views/Bridge/utils.ts
+++ b/src/views/Bridge/utils.ts
@@ -276,7 +276,7 @@ export function getAllChains() {
     });
 }
 
-export function getInitialRouteFromQueryParams() {
+export function getRouteFromQueryParams() {
   const params = new URLSearchParams(window.location.search);
 
   const fromChain = Number(params.get("from"));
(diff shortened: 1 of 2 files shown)


pxrl (approved):
Tests OK for me :+1:

src/views/Bridge/utils.ts
```diff
@@ -273,3 +275,25 @@ export function getAllChains() {
...
+  };
+
+  let route = findNextBestRoute(["fromChain"], filter);
+
+  if (!route) {
+    route = getInitialRoute(filter);
+  }
+
+  return route;
+}
```
james-a-morris, line 299: Will this change if we call the function with parameters that have been changed?

I.e. If we call `across.to?param=test`, and then call this function after we route to `across.to?param=test1`, will it produce the same results?

I only ask because if it's possible to change the result of this call, we may want to not call it `getInitialRoute`

dohaki: Yea good point. I will rename to `getRouteFromQueryParams `

dohaki: Done https://github.com/across-protocol/frontend-v2/pull/844/commits/12e76a45b7ddbaf3b440ce50afc92c74262f91f7

PR #846, opened 2023-09-20
fix: do not clear amount input for interchangeable tokens

dohaki:
This PR changes the behavior of clearing the bridge amount input only if the newly selected token is truly different.


james-a-morris (approved):
LGTM - I haven't had an issue running on local

PR #849, opened 2023-09-21
fix: missing gas padding on native gas token deposits

dohaki:
This PR does two main things:
- Replace `gasMultiplier` for a single chain with `gasMultiplierPerChain` which is a mapping `chainId -> gasMultiplier`. This allows setting different buffers for different chains.
- Takes gas costs into account for max. balance value of native gas token deposits.


src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -0,0 +1,80 @@
...
+  );
+}
+
+async function estimateGasCostsForDeposit(
+  selectedRoute: Route,
+  signer: providers.JsonRpcSigner
+) {
+  const provider = getProvider(selectedRoute.fromChain);
+  const spokePool = config.getSpokePool(selectedRoute.fromChain);
+  const mockedArgs = {
```
james-a-morris, line 58: Nit: should we call this `mocked`? I feel like that term is usually reserved for tests.

src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+  const paddedGasEstimation = await getPaddedGasEstimation(
+    selectedRoute.fromChain,
+    spokePool,
+    "deposit",
+    ...Object.values(mockedArgs),
+    { value: selectedRoute.isNative ? mockedArgs.amount : 0 }
+  );
+  const gasPrice = await provider.getGasPrice();
+  return gasPrice.mul(paddedGasEstimation);
```
james-a-morris, line 79: Will we need to account for any decimals? i.e. with a div by `fixedPointAdjustment`

src/utils/transactions.ts
```diff
@@ -26,34 +30,53 @@ export async function sendSignedTransaction(
...
+  const gasMultiplier = gasMultiplierPerChain[chainId];
+  /* If the gas multiplier hasn't been set, run this function as a normal tx */
+  if (!gasMultiplier) {
+    return contract.estimateGas[method](...args);
+  } else {
+    // Estimate the gas with the provided estimateGas logic
+    const gasEstimation = await contract.estimateGas[method](...args);
+    // Factor in the padding
+    const gasToRecommend = gasEstimation
+      .mul(parseEther(String(gasMultiplier)))
```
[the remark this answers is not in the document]
dohaki: Yea, we could. But I am not sure if this makes things more predictable or intuitive tbh. So if I import this constant somewhere, I would not expect this to be a BigNumber.

src/views/Bridge/hooks/useMaxBalance.ts
dohaki: Hmm, yea feel free to provide an alternate name :)

src/views/Bridge/hooks/useMaxBalance.ts
dohaki: Nope, because the gas estimation is returned in `gas` and the gas price is returned in `wei per gas`

src/utils/transactions.ts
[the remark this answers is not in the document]
james-a-morris: Fair enough!

src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -0,0 +1,80 @@
...
+          selectedRoute.fromTokenSymbol !== "ETH"
+            ? balance
+            : // For ETH, we need to take the gas costs into account before setting the max. bridgable amount
+              await estimateGasCostsForDeposit(selectedRoute, signer)
+                .then((estimatedGasCosts) =>
+                  max(balance.sub(estimatedGasCosts), 0)
+                )
+                .catch((err) => {
+                  console.error(err);
+                  return max(balance.sub(utils.parseEther("0.01")), 0);
```
james-a-morris, line 38: Should we encode this somewhere as a constant? This way, we can modify `0.01` without having to grep for it?

src/views/Bridge/hooks/useMaxBalance.ts
james-a-morris: Maybe: `argsForEstimation`

src/views/Bridge/hooks/useMaxBalance.ts
james-a-morris: Perfect

james-a-morris (approved):
LGTM - left a comment

pxrl (commented):
One high-level comment - we also have gas padding that occurs in the API for estimating the fill cost of the relayer, and there's some potential for them to be confused. The main hint here is that this new env var is prefixed with REACT_APP.

Not sure exactly what to do about it, but I just wanted to note it.

src/views/Bridge/hooks/useMaxBalance.ts
dohaki: Good point! Will change

dohaki:
> One high-level comment - we also have gas padding that occurs in the API for estimating the fill cost of the relayer, and there's some potential for them to be confused. The main hint here is that this new env var is prefixed with REACT_APP.
> 
> Not sure exactly what to do about it, but I just wanted to note it.

@pxrl Thanks for flagging. Yea I also noticed that. You are referring to `GAS_MARLUP`, right? Do you think it would make sense to somehow consolidate those into a single one?

PR #852, opened 2023-09-22
chore: bump cypress to latest

dohaki:
This bumps cypress to the latest version. This should also fix our CI. Related https://github.com/cypress-io/cypress/issues/27804


james-a-morris (approved):
Great call!

PR #854, opened 2023-09-28
fix: bump web3 onboard + fix for mobile

dohaki:
Fixes ACX-1531


patches/@web3-onboard+walletconnect+2.4.1.patch
james-a-morris, line 1: Is this removed because we are at a later version of web3-onboard & walletconnect?

src/components/Layout/index.ts
james-a-morris, line 1: If this is removed in favor of `LayoutV2`, should we refactor `LayoutV2 -> Layout` since `Layout` is now a free namespace?

yarn.lock
james-a-morris, line 1: Should we enable dependency review for this repo?

james-a-morris (commented):
Two comments + one outstanding question about the repo. Otherwise, looks great!

patches/@web3-onboard+walletconnect+2.4.1.patch
dohaki: Yes, the newest version has the fix so we don't need to manually apply our patch anymore.

src/components/Layout/index.ts
dohaki: Yes, this makes sense. I would tackle this in a separate PR though.

yarn.lock
dohaki: It is already enabled https://github.com/across-protocol/frontend-v2/actions/runs/6338448518/job/17215577629

src/components/Layout/index.ts
james-a-morris: SGTM! Added ACX-1602 so it's not forgotten.

james-a-morris (approved):
LGTM

PR #855, opened 2023-09-28
fix: table styles mobile rewards page

dohaki:
Fixes ACX-1277, fixes ACX-1284 and ACX-1276


james-a-morris (approved):
Good change to use the more updated V2 QUERY strings. LGTM

PR #856, opened 2023-09-29
fix: transactions page mobile fixes

dohaki:
Fixes various issues on the transactions page for mobile. Fixes ACX-1279, ACX-1280 and ACX-1281


james-a-morris (approved):
Nice work on migrating to QUERIESV2. This looks good on mobile

PR #858, opened 2023-10-02
fix: update quote on error

dohaki:
Hopefully fixes ACX-1565

I could reproduce the above issue when I rejected the transaction within MetaMask and afterward changed the route or amount.

The root causes were:
- Fees were not updating when the action mutation errored by user rejecting a tx
- We did not let these kinds of errors bubble up and suppressed them.

### DIFF
--- a/src/views/Bridge/hooks/useBridgeAction.ts
+++ b/src/views/Bridge/hooks/useBridgeAction.ts
@@ -90,21 +90,17 @@ export function useBridgeAction(
     let succeeded = false;
     let timeSigned: number | undefined = undefined;
     let tx: ContractTransaction | undefined = undefined;
+    addToAmpliQueue(() => {
+      // Instrument amplitude before sending the transaction for the submit button.
+      ampli.transferSubmitted(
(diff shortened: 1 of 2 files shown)


src/views/Bridge/hooks/useBridgeAction.ts
```diff
@@ -90,21 +90,17 @@ export function useBridgeAction(
...
-          )
-        );
-      });
-      const timeSubmitted = Date.now();
+    addToAmpliQueue(() => {
+      // Instrument amplitude before sending the transaction for the submit button.
+      ampli.transferSubmitted(
+        generateTransferSubmitted(frozenQuote, referrer, frozenInitialQuoteTime)
+      );
+    });
```
pxrl, line 98: +1 this seems very helpful!

james-a-morris: Good call, @dohaki

james-a-morris (commented):
LGTM - I can't emulate the error anymore on this staging branch.

pxrl (approved):

PR #860, opened 2023-10-04
fix(sec): remove gcp key exposure

james-a-morris:
We shouldn't be exposing keys with the `REACT_APP` prefix unless they're designed to be exposed publicly (i.e. our RPC keys). 

We will need to modify our prod ENV to reflect this change.


nicholaspai (approved):
LGTM but make sure we have production env keys in before we deploy this

dohaki (approved):

PR #861, opened 2023-10-05
feat: support `SpokePoolVerifier` contract for native and enabled routes

dohaki:
This PR uses the `deposit` of a `SpokePoolVerifier` contract if enabled via the `routes.json` config file and if the sending token is a native one.

Closes ACX-1589

Needs https://github.com/across-protocol/scraper-api/pull/266 before it can be merged.


src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
```diff
@@ -1708,5 +1708,27 @@
...
+      "address": "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D"
+    },
+    {
+      "chain": 8453,
+      "address": "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D"
+    },
+    {
+      "chain": 42161,
+      "address": "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D"
+    }
```
james-a-morris, line 1732: Per the contract docs, the `SpokePoolVerifier` is deployed via `Create2` to the same address on all chains. We can simplify the structure to something like the following:
```
"spokePoolVerifier": {
    "address": "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D",
    "chains": [1, 10, 137, 8453, 42161]
}

```

Doc reference: https://github.com/across-protocol/contracts-v2/blob/99f904a1e813bbc69e410cb30ab1ed721d137af2/contracts/SpokePoolVerifier.sol#L7-L13

src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
james-a-morris, line 1: We should reflect this addition to the Goerli route for consistency

src/utils/bridge.ts
```diff
@@ -178,14 +178,34 @@ export async function sendAcrossDeposit(
...
-  if (!code) {
+
+  // We send native tokens to the SpokePoolVerifier contract if enabled for the chain.
+  const shouldUseSpokePoolVerifier =
+    config.enabledSpokePoolVerifierChains.has(fromChain) && isNative;
+
+  const spokePool = config.getSpokePool(fromChain);
+  const spokePoolVerifier = shouldUseSpokePoolVerifier
+    ? config.getSpokePoolVerifier(fromChain)
+    : undefined;
```
james-a-morris, line 190: Could we pack this logic into the `config.getSpokePoolVerifier`? It could return `undefined` if the chain isn't available.

src/utils/bridge.ts
```diff
@@ -178,14 +178,34 @@ export async function sendAcrossDeposit(
...
+      : Promise.resolve(true),
+  ]);
+  if (!spokePoolCode) {
     throw new Error(`SpokePool not deployed at ${spokePool.address}`);
   }
+  if (!spokePoolVerifierCode) {
+    throw new Error(
+      `SpokePoolVerifier not deployed at ${spokePoolVerifier?.address}`
+    );
+  }
```
james-a-morris, line 205: Do these conditions check that any contract is deployed to these addresses or that the `SpokePool` and `SpokePoolVerifier` contracts are deployed to these addresses? If the former, we may want to revise these errors to indicate that no contract is deployed to the address.

src/utils/bridge.ts
```diff
@@ -178,14 +178,34 @@ export async function sendAcrossDeposit(
...
+
+  const spokePool = config.getSpokePool(fromChain);
+  const spokePoolVerifier = shouldUseSpokePoolVerifier
+    ? config.getSpokePoolVerifier(fromChain)
+    : undefined;
+
+  const [spokePoolCode, spokePoolVerifierCode] = await Promise.all([
+    provider.getCode(spokePool.address),
+    spokePoolVerifier
+      ? await provider.getCode(spokePoolVerifier.address)
```
james-a-morris, line 195: Do we need this additional await? Should the promise be resolved in the `Promise.all( )`?
```suggestion
      ? provider.getCode(spokePoolVerifier.address)
```

src/utils/bridge.ts
```diff
@@ -178,14 +178,34 @@ export async function sendAcrossDeposit(
...
+      : Promise.resolve(true),
+  ]);
+  if (!spokePoolCode) {
     throw new Error(`SpokePool not deployed at ${spokePool.address}`);
   }
+  if (!spokePoolVerifierCode) {
```
james-a-morris, line 201: If I recall correctly, would we need to check that `!spokePoolCode || spokePoolCode === "0x"` and `!spokePoolVerifierCode || spokePoolVerifierCode === "0x"`

james-a-morris (commented):
Left some comments

src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
dohaki: Yea you are right. I somehow thought that it would make sense to keep the granularity in case the address would differ (like on zkSync) but then I realized it doesn't add the desired protection if the address differs. Will change to your suggested structure.

src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json
dohaki: Changed here https://github.com/across-protocol/frontend-v2/pull/861/commits/33a177c3d706aace715e0e5be10d97b3f413bdc3

src/utils/bridge.ts
dohaki: Changed here https://github.com/across-protocol/frontend-v2/pull/861/commits/79f059456f0ff72f00507398365b291ca10a520e

src/utils/bridge.ts
dohaki: Hmm yea, it is actually the former... So, do you mean changing the error message to
```
No contract deployed at XXX
```
is better?

src/utils/bridge.ts
dohaki: Changed https://github.com/across-protocol/frontend-v2/pull/861/commits/720e5fbdd7c67bc5bda8afaf6435040a16b2f898

src/utils/bridge.ts
dohaki: Yup that's right! Changed

james-a-morris (approved):
LGTM! Thanks for reviewing my comments

PR #862, opened 2023-10-05
improve: simulate relay gas fees for messages

james-a-morris:
We will require the following PR to be merged first: https://github.com/across-protocol/sdk-v2/pull/427

Fixes ACX-1529, ACX-1575, and ACX-1574


api/suggested-fees.ts
```diff
@@ -78,6 +84,31 @@ const handler = async (
...
+    }
+
+    const messagePayload =
+      sdk.utils.isDefined(message) && sdk.utils.isDefined(recipientAddress)
+        ? {
+            message,
+            recipientAddress,
+            relayerAddress,
+          }
+        : undefined;
```
mrice32, line 110: As discussed in the other PR, it would be nice to be able to carry forward the recipient address even if the message isn't defined.

api/_utils.ts
```diff
@@ -492,22 +482,43 @@ export const getTokenSymbol = (tokenAddress: string): string => {
...
+    } else {
+      // Otherwise, we throw a 500.
+      throw new Error(
+        `Could not process relayer fee details.
+      ${
+        sdk.utils.isDefined(messagePayload)
+          ? " The likely cause is related to message payload."
+          : ""
+      }`
+      );
```
mrice32, line 519: nit: I think it's (ironically) error prone to have switching logic based on the text of a thrown error.

Prefer returning a distinct error type to indicate the _type_ of issue that came up during processing. Each type of error can have its own error type or you can always return your custom error type and have that error type have a readable property (an enum for instance) that will tell you what type of issue occured (input error vs processing error).

mrice32 (commented):
Looks good, just a few comments!

api/suggested-fees.ts
james-a-morris: Done: https://github.com/across-protocol/frontend-v2/pull/862/commits/80cb76f5159902db393a33ce2fc5155571e4ba56

api/_utils.ts
james-a-morris: Fair point - most of the errors are now being checked for in the FE itself. So, this is now resolved per https://github.com/across-protocol/frontend-v2/pull/862/commits/1abe39e004dd4fde3782cfc4e860d02e1fa950c7

api/suggested-fees.ts
```diff
@@ -78,6 +85,46 @@ const handler = async (
     ]);
     const { l1Token, hubPool, chainId: computedOriginChainId } = tokenDetails;
 
+    if (sdk.utils.isDefined(message) && !sdk.utils.isMessageEmpty(message)) {
+      if (message.length % 2 !== 0) {
+        // Our message encoding is a hex string, so we need to check that the length is even.
+        throw new InputError("Message must be an even hex string");
```
nicholaspai, line 91: nit: Is there no helper method we can use from `ethers` to more accurately determine if its a hex encoded?

api/suggested-fees.ts
```diff
@@ -78,6 +85,46 @@ const handler = async (
...
+      }
+      const isRecipientAContract = await sdk.utils.isContractDeployedToAddress(
+        recipientAddress,
+        provider
+      );
+      if (!isRecipientAContract) {
+        throw new InputError(
+          "Recipient address with a contract deployed to it is required when a message is provided"
```
nicholaspai, line 99:
```suggestion
          "Recipient must be a contract when a message is provided"
```

This is needlessly complicated IMO

api/suggested-fees.ts
```diff
@@ -78,6 +85,46 @@ const handler = async (
...
+        // of the `handleAcrossMessage` we will check that the balance of the relayer is sufficient to
+        // support this deposit.
+        const destinationToken =
+          sdk.utils.getL2TokenAddresses(l1Token)?.[Number(destinationChainId)];
+        if (!sdk.utils.isDefined(destinationToken)) {
+          throw new InputError(
+            `Could not resolve token address on ${destinationChainId} for ${l1Token}`
+          );
+        }
+        const balanceOfToken = await sdk.utils.getTokenBalance(
```
nicholaspai, line 115: so this is where I think we might want to add an endpoint `get-balance` where we can cache the balance for the `relayerAddress`, `destinationToken` and chain where `provider` to get some speedup on average.

api/suggested-fees.ts
```diff
@@ -78,6 +85,46 @@ const handler = async (
...
+          relayerAddress,
+          destinationToken,
+          provider
+        );
+        if (balanceOfToken.lt(amountInput)) {
+          throw new InputError(
+            `Quotes for partial filling when message is defined are unavailable.`
```
nicholaspai, line 122: I don't think this is going to be a helpful error message to the user. The error should instead be indicating that the `relayerAddress` doesn't have enough funds to fully fill the message and the user should contact our support team for help.

nicholaspai (commented):
Nice, just some nits

api/suggested-fees.ts
james-a-morris: Not really. This is an issue of the representation of the byte representation being off since we have a byte array and bytes are stored with 2 hex decimals (2 4-bit hex digits).

We could pad for these edge cases by adding a `0x0${oddLengthHex}` but we could also consider this to be an error path.

james-a-morris: That said - I've added a check to make sure the data takes the form of `0x[0-9a-fA-F]` with `ethers.utils.isHexString()` but this won't check for the uneven representation because even though you can't accurately represent a byte array with `0xa` it's still a valid hex string.

api/suggested-fees.ts
james-a-morris: https://github.com/across-protocol/frontend-v2/pull/862/commits/a8e7c84a9c751734ff705059e7bd2c31e570f68c

james-a-morris: https://github.com/across-protocol/frontend-v2/pull/862/commits/703312bb7db53f7766a41cc3ac2db9e4feb781ca

api/_utils.ts
```diff
@@ -589,23 +609,63 @@ export const isRouteEnabled = (
...
   account: string,
+  token?: string,
   blockTag: number | "latest" = "latest"
 ): Promise<BigNumber> => {
-  return ERC20__factory.connect(token, getProvider(Number(chainId))).balanceOf(
+  const provider = getProvider(Number(chainId));
+  if (sdk.utils.isDefined(token)) {
+    return sdk.utils.getTokenBalance(account, token, provider, blockTag);
+  } else {
+    return provider.getBalance(account, blockTag);
```
nicholaspai, line 630: Do we expect that `token` is undefined when the caller is intending to get balance of native token..? That seems unintuitive for me, I'd rather `token` be defined and then check if token is native, and then do the `sdk.utils.getTokenBalance`/`provider.getBalance` swap

api/account-balance.ts
```diff
@@ -0,0 +1,70 @@
...
+      account: account,
+      token: token,
+      isNative: token === undefined,
+      tag: blockTag,
+    };
+    // Determine the age of the caching of the response
+    // 1. If the blockTag is specified, then we can cache for 5 minutes (300 seconds)
+    // 2. If the blockTag is not specified, the "latest" block is used, and we should
+    //    only cache for 10 seconds.
+    const cachingTime = blockTag ? 300 : 10;
```
nicholaspai, line 50: won't `blockTag` mostly be unspecified since we usually want latest? Why only 10 seconds for latest?

api/limits.ts
```diff
@@ -149,12 +150,13 @@ const handler = async (
         ethers.BigNumber.from("10").pow(18),
         computedOriginChainId,
         Number(destinationChainId),
+        utils.randomAddress(),
```
nicholaspai, line 153: should this really be random? Would rather this be a known dummy address, just incase some real address collision on the `destinationChainId` causes issues

api/account-balance.ts
```diff
@@ -0,0 +1,70 @@
...
+  logger.debug({
+    at: "AccountBalance",
+    message: "Query data",
+    query,
+  });
+  try {
+    // Validate the query parameters
+    assert(query, AccountBalanceQueryParamsSchema);
+    // Deconstruct the query parameters
+    let { token, account, chainId, blockTag } = query;
```
nicholaspai, line 34: Why do we want to support the `blockTag` to be specified? Isn't the main use case to grab balance at the "latest" block? Then we can cache that "latest" result for 5 mins

nicholaspai (commented):
Great start, left some comments

api/account-balance.ts
james-a-morris: Removed: https://github.com/across-protocol/frontend-v2/pull/862/commits/bc6c8f73b975ca0f6a1025a16ca36003ac8730c9

api/account-balance.ts
james-a-morris: https://github.com/across-protocol/frontend-v2/pull/862/commits/bc6c8f73b975ca0f6a1025a16ca36003ac8730c9

Updated & resolved timing offline

api/_utils.ts
james-a-morris: `token` in this case is the address for the associated `ERC20` token. Should we not let it be an address and instead make it a symbol?

api/limits.ts
james-a-morris: The likelihood here would be very low assuming that `ethers` is using a regular implementation of a PRNG. However, I have changed this to use `0x893d0d70ad97717052e3aa8903d9615804167759` here: https://github.com/across-protocol/frontend-v2/pull/862/commits/682126802e21201ca5aed8a4aff88f54ca05101c

api/_utils.ts
```diff
@@ -589,23 +612,63 @@ export const isRouteEnabled = (
...
- * @param blockTag A blockTag to specify a historical balance date
+ * @param blockTag A blockTag to specify a historical balance date or the latest balance
  * @returns A promise that resolves to the BigNumber of the balance
  */
 export const getBalance = (
   chainId: string | number,
-  token: string,
   account: string,
+  token?: string,
   blockTag: number | "latest" = "latest"
```
nicholaspai, line 627: Why even support `blockTag` versus always querying latest?

api/_utils.ts
nicholaspai: It's fine the way it is, I just don't love that `undefined` defaults to native gas token, that doesn't seem like a natural logical jump

api/_utils.ts
```diff
@@ -589,23 +612,63 @@ export const isRouteEnabled = (
...
+ * @param token The valid ERC20 token address on the given `chainId`. If undefined, the native currency will be used
+ * @param account A valid Web3 wallet address
+ * @param blockTag A blockTag to specify a historical balance date or the latest balance
+ * @returns A promise that resolves to the BigNumber of the balance
+ */
+export const getCachedTokenBalance = async (
+  chainId: string | number,
+  account: string,
+  token?: string,
+  blockTag: number | "latest" = "latest"
```
nicholaspai, line 650: same question as above, why even support `blockTag` overriding instead of just using latest

james-a-morris: https://github.com/across-protocol/frontend-v2/pull/862/commits/fef8c7a170bb8b4c89673a51c1867147de121c35

api/_utils.ts
james-a-morris: https://github.com/across-protocol/frontend-v2/pull/862/commits/fef8c7a170bb8b4c89673a51c1867147de121c35

api/account-balance.ts
```diff
@@ -0,0 +1,61 @@
...
+    at: "AccountBalance",
+    message: "Query data",
+    query,
+  });
+  try {
+    // Validate the query parameters
+    assert(query, AccountBalanceQueryParamsSchema);
+    // Deconstruct the query parameters
+    let { token, account, chainId } = query;
+    // Rely on the utils to query the balance of either the native
```
nicholaspai, line 34: need to change this comment

api/_utils.ts
```diff
@@ -487,25 +477,58 @@ export const getTokenSymbol = (tokenAddress: string): string => {
...
  * @returns The a promise to the relayer fee for the given `amount` of transferring `l1Token` to `destinationChainId`
  */
-export const getRelayerFeeDetails = (
+export const getRelayerFeeDetails = async (
   l1Token: string,
   amount: sdk.utils.BigNumberish,
   originChainId: number,
   destinationChainId: number,
-  tokenPrice?: number
+  recipientAddress = "0x893d0d70ad97717052e3aa8903d9615804167759", // A known dummy address
```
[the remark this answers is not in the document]
james-a-morris: The only usage of `constants` in the SDK is for the RelayerAddress. This is the recipient & it was added so that we only had to specify the dummy address in one location. 

I'm not sure adding a recipient address to the `/limits` would benefit the endpoint as this recipient is just needed to resolve the minimum deposit for a dummy fill request.

api/limits.ts
```diff
@@ -149,34 +150,25 @@ const handler = async (
...
+          getCachedTokenBalance(destinationChainId!, destinationToken, relayer)
         )
       ),
       Promise.all(
         transferRestrictedRelayers.map((relayer) =>
-          getBalance(
-            destinationChainId!,
-            destinationToken,
-            relayer,
-            BLOCK_TAG_LAG
```
nicholaspai, line 171: How was `BLOCK_TAG_LAG` being used before? IIUC i don't see this parameter passed to `getBalance` in `_utils`..?

nicholaspai (commented):
One question https://github.com/across-protocol/frontend-v2/pull/862/files#r1372068764

api/limits.ts
james-a-morris: It was passed to all `getBalance()` calls in the `/limits` https://github.com/across-protocol/frontend-v2/blob/33ddbde87d77df52144bed72bd2c9751973fd38b/api/limits.ts#L155-L182

api/_utils.ts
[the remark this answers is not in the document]
james-a-morris: Resolved offline ^

api/limits.ts
james-a-morris: The `_utils.ts` `getBalance()` had an optional `blockTag` parameter:

```
export const getBalance = (
  chainId: string | number,
  token: string,
  account: string,
  blockTag: number | "latest" = "latest"
): Promise<BigNumber> => {
  return ERC20__factory.connect(token, getProvider(Number(chainId))).balanceOf(
    account,
    { blockTag }
  );
};
```

james-a-morris:
@nicholaspai rerequested due to needing a UI change to accommodate the new request.

nicholaspai (approved):
LGTM but do we need to make changes to the relayer or let other callers of the API know before we merge and deploy?

PR #863, opened 2023-10-05
feat: add vercel as dev dependency

james-a-morris:
Vercel is a required part of our development cycle. This enforces that vercel is available locally and includes a script to easily start the vercel dev command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     "ts-jest": "^29.1.1",
     "ts-node": "^10.9.1",
     "tsconfig-paths-webpack-plugin": "^4.0.0",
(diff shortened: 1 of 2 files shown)


mrice32 (approved):

dohaki (commented):
LGTM but there seem to be some vulnerabilities with this dependency 🤔

james-a-morris:
This is something that we should put in the tech debt cycle. It's no longer an easy lift item as vulnerabilities have been detected. Thoughts on taking this PR down @nicholaspai @dohaki ?

james-a-morris:
Closing for now - will re-open during the next tech debt cycle.

PR #864, opened 2023-10-06
fix: make quotes fees less prone to race conditions

dohaki:
Try to fix ACX-1583

This PR removes some of `useEffect`s prone to race conditions and replaces them with a `useQuery`.

### DIFF
--- a/src/utils/amplitude.ts
+++ b/src/utils/amplitude.ts
@@ -248,11 +248,11 @@ export function generateTransferQuote(
   tokenInfo: TokenInfo,
   fromChainInfo: ChainInfo,
   toChainInfo: ChainInfo,
-  toAddress: string,
-  account: string,
   tokenPrice: BigNumber,
   estimatedTimeToRelayObject: ConfirmationDepositTimeType,
-  amount: BigNumber
+  amount: BigNumber,
+  account?: string,
+  toAddress?: string
 ): TransferQuoteReceivedProperties {
   // Create a function that converts a wei amount into a formatted token amount
   const formatTokens = (wei: BigNumber) =>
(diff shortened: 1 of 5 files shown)


src/utils/constants.ts
```diff
@@ -126,7 +126,7 @@ export type ChainInfoList = ChainInfo[];
 export type ChainInfoTable = Record<number, ChainInfo>;
 
 export const defaultBlockPollingInterval =
-  Number(process.env.REACT_APP_DEFAULT_BLOCK_POLLING_INTERVAL_S || 30) * 1000;
+  Number(process.env.REACT_APP_DEFAULT_BLOCK_POLLING_INTERVAL_S || 15) * 1000;
```
james-a-morris, line 129: Should we be reflecting this change in the ENV variable? We could make it required so that the user will always need to supply it.

src/utils/amplitude.ts
```diff
@@ -248,11 +248,11 @@ export function generateTransferQuote(
...
   fromChainInfo: ChainInfo,
   toChainInfo: ChainInfo,
-  toAddress: string,
-  account: string,
   tokenPrice: BigNumber,
   estimatedTimeToRelayObject: ConfirmationDepositTimeType,
-  amount: BigNumber
+  amount: BigNumber,
+  account?: string,
+  toAddress?: string
```
james-a-morris, line 255: Is this now an optional field?

src/views/Bridge/hooks/useBridge.ts
```diff
@@ -46,22 +51,22 @@ export function useBridge() {
...
-    quote,
-    initialQuoteTime,
-    quotedFees,
-    quotedLimits,
-    quotePriceUSD,
-    isQuoteLoading,
-  } = useTransferQuote(
+  const { data: transferQuote, isLoading: isQuoteLoading } = useTransferQuote(
     selectedRoute,
     parsedAmount?.gt(0) ? parsedAmount : BigNumber.from(0),
```
james-a-morris, line 56: I believe we import `bnZero` in the SDK

src/views/Bridge/hooks/useTransferQuote.ts
```diff
@@ -54,91 +35,69 @@ export function useTransferQuote(
...
-  }, [
-    amountToBridge,
-    quotedLimits,
-    selectedRoute.fromChain,
-    selectedRoute.toChain,
-  ]);
+  return useQuery({
+    queryKey: [
+      "quote",
+      feesQuery.fees?.quoteBlock.toString(),
```
james-a-morris, line 41: Should we allow this undefined key? Would it make sense to disable this query if this is undefined and opt for a default return value

james-a-morris (commented):
Left some comments

src/utils/constants.ts
dohaki: I actually think in this case changing the default makes more sense. Because the average block time on mainnet is 15 seconds. With that in mind, does that also make sense to you?

src/utils/amplitude.ts
dohaki: Yep, this makes sense IMO because we also generate quotes if no wallet is connected.

src/views/Bridge/hooks/useBridge.ts
dohaki: Ok replace here https://github.com/across-protocol/frontend-v2/pull/864/commits/697cc68b1f592b312c9295cdd3e513778c0534e7

src/views/Bridge/hooks/useTransferQuote.ts
dohaki: The query already gets disabled if `feesQuery.fees` is `undefined`. See https://github.com/across-protocol/frontend-v2/pull/864/files#diff-f3e48b2503b68de3bbb63a4f53558ba42ff4b2aef87b72fc46e19bf3dddcd554R48

src/utils/amplitude.ts
dohaki: Yes, this is now optional. Because we also generate quotes when the user did not connect a wallet, i.e. the `toAddress` is `undefined`.

PR #866, opened 2023-10-12
fix(bridge): initial quote time issue

dohaki:
In some edge cases the quote time wasn't set properly which led to a disabled bridge action button.

### DIFF
--- a/src/views/Bridge/hooks/useTransferQuote.ts
+++ b/src/views/Bridge/hooks/useTransferQuote.ts
@@ -86,14 +86,16 @@ export function useTransferQuote(
         toAddress
       );
 
+      let initialQuoteTimeToUse = initialQuoteTime || Date.now();
+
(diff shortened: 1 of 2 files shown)


src/views/Bridge/hooks/useTransferQuote.ts
```diff
@@ -86,14 +86,16 @@ export function useTransferQuote(
         toAddress
       );
 
+      let initialQuoteTimeToUse = initialQuoteTime || Date.now();
```
pxrl, line 89: I'm a little ignorant of how `initialQuoteTime` is set - does it come from the FE itself, or is it read from `suggested-fees`?

dohaki: This is just used for Amplitude and set on the FE itself

pxrl (approved):
LGTM

PR #868, opened 2023-10-12
fix: MetaMask browser app crash on wallet change

dohaki:
Fixes ACX-1532

When changing the account in the MetaMask mobile browser, the `wallet.accounts` array is empty for a short period of time. This led to a fatal crash of the app.


james-a-morris (commented):
AFAICT this is both resolving the fix & setting the failure message more accurately?

dohaki:
> AFAICT this is both resolving the fix & setting the failure message more accurately?

@james-a-morris Yep, exactly!

james-a-morris (approved):
Perfect. thanks @dohaki

PR #871, opened 2023-10-13
fix: spread args in send with padded gas utils

dohaki:
The PR https://github.com/across-protocol/frontend-v2/pull/849 caused an issue with on-chain actions that were wrapped by `sendWithPaddedGas` because we didn't spread the args properly.


pxrl (approved):
Worked OK for me to unstake LP tokens: https://etherscan.io/tx/0x84d071229a1f4d2816db7eb44444e3cb7ac792f75377dbcc9d3e09c325578846

PR #872, opened 2023-10-13
feat: add caching to suggested fees

james-a-morris:


api/suggested-fees.ts
```diff
@@ -179,6 +179,7 @@ const handler = async (
       responseJson,
     });
 
+    response.setHeader("Cache-Control", "s-maxage=300");
```
dohaki, line 182: Is this a safe cache time to set? Referring to the timestamp / latest block logic we do here https://github.com/across-protocol/frontend-v2/blob/019f6198192514703d31f8e8104e5b032e7c1859/api/suggested-fees.ts#L75

PR #873, opened 2023-10-16
fix: use pre-computed gas usage for deposit max amounts

dohaki:
This PR makes setting the max. bridgable amount for native token deposits more ~accurate and~ reliable by:
- Removing the usage of a fallback value gas costs and using infinite retry
- Using a pre-computed value for gas expenditure instead of querying via an additional RPC call

### DIFF
--- a/src/views/Bridge/hooks/useMaxBalance.ts
+++ b/src/views/Bridge/hooks/useMaxBalance.ts
@@ -1,17 +1,8 @@
 import { useQuery } from "react-query";
-import { BigNumber, providers, constants, utils } from "ethers";
+import { BigNumber, constants } from "ethers";
 
 import { useBalanceBySymbol, useConnection } from "hooks";
-import {
-  getConfig,
-  Route,
-  max,
-  getProvider,
-  fallbackEstimatedGasCosts,
-} from "utils";
-import { getPaddedGasEstimation } from "utils/transactions";
-
-const config = getConfig();
+import { Route, max, getProvider, gasExpenditureDeposit } from "utils";
 
 export function useMaxBalance(selectedRoute: Route) {
   const { balance } = useBalanceBySymbol(
@@ -35,14 +26,9 @@ export function useMaxBalance(selectedRoute: Route) {
           selectedRoute.fromTokenSymbol !== "ETH"
             ? balance
             : // For ETH, we need to take the gas costs into account before setting the max. bridgable amount
-              await estimateGasCostsForDeposit(selectedRoute, signer)
-                .then((estimatedGasCosts) =>
-                  max(balance.sub(estimatedGasCosts), 0)
-                )
-                .catch((err) => {
-                  console.error(err);
-                  return max(balance.sub(fallbackEstimatedGasCosts), 0);
-                });
+              await estimateGasCostsForDeposit(selectedRoute).then(
+                (estimatedGasCosts) => max(balance.sub(estimatedGasCosts), 0)
+              );
       } else {
         maxBridgeAmount = constants.Zero;
       }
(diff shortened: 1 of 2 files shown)


src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -51,38 +37,13 @@ export function useMaxBalance(selectedRoute: Route) {
...
+      retry: true,
     }
   );
 }
 
-async function estimateGasCostsForDeposit(
-  selectedRoute: Route,
-  signer: providers.JsonRpcSigner
-) {
+async function estimateGasCostsForDeposit(selectedRoute: Route) {
```
pxrl, line 45: Do we want to explicitly document that this assumes an empty message? AFAIK we're not going to support messages via the FE, but if that changes then this computation would under-price deposits.

src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -51,38 +37,13 @@ export function useMaxBalance(selectedRoute: Route) {
...
-  const paddedGasEstimation = await getPaddedGasEstimation(
-    selectedRoute.fromChain,
-    spokePool,
-    "deposit",
-    ...Object.values(argsForEstimation),
-    { value: selectedRoute.isNative ? amount : 0 }
-  );
   const gasPrice = await provider.getGasPrice();
-  return gasPrice.mul(paddedGasEstimation);
+  return gasPrice.mul(gasExpenditureDeposit);
```
pxrl, line 48: `getGasPrice()` can be unreliable because it's mostly ignorant of the gas pricing strategies of each chain. It's probably fine as an approximation, but should we apply a multiplier? I think something really coarse would be fine - i.e. 3x or so. Given that the deposit cost is so low it will then reserve only a small amount of the native token, and that should be fine by users.

src/views/Bridge/hooks/useMaxBalance.ts
```diff
@@ -35,14 +26,9 @@ export function useMaxBalance(selectedRoute: Route) {
...
             : // For ETH, we need to take the gas costs into account before setting the max. bridgable amount
-              await estimateGasCostsForDeposit(selectedRoute, signer)
-                .then((estimatedGasCosts) =>
-                  max(balance.sub(estimatedGasCosts), 0)
-                )
-                .catch((err) => {
-                  console.error(err);
-                  return max(balance.sub(fallbackEstimatedGasCosts), 0);
-                });
+              await estimateGasCostsForDeposit(selectedRoute).then(
```
pxrl, line 29: ooc, how regularly does this run? I wonder if we want some smoothing logic so that the estimated gas cost doesn't bounce around and cause the deposit amount to flip-flop between being valid and invalid.

src/views/Bridge/hooks/useMaxBalance.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/873/commits/f20b1aff4ced9e53c7098798f58c9cee9ae746fe

nicholaspai (approved):
This does seem like it'll make the FE faster but doesn't it make estimations less accurate (in the PR you say it makes it more accurate) because we use a hardcoded gas cost?

src/views/Bridge/hooks/useMaxBalance.ts
pxrl: Wow - I'd expected something like 2 - 10 seconds. 5 minutes is actually really long; we might want to revisit that separately, because it might lead to issues if gas becomes volatile.

pxrl:
> This does seem like it'll make the FE faster but doesn't it make estimations less accurate (in the PR you say it makes it more accurate) because we use a hardcoded gas cost?

Yeah, it would lead to a bit less accuracy vs. a successful gas estimation. The thinking behind this PR is that there have been several reports of falling back to the default gas buffer of 0.01 ETH, mostly on deposits originating on zkSync. Secondarily, incurring the cost of an estimation on every single deposit is not really ideal - it's both faster and cheaper to use a fixed cost. As far as I understand, the user's wallet should anyway simulate this against their own RPCs before they sign it.

pxrl (approved):

src/views/Bridge/hooks/useMaxBalance.ts
dohaki: Ah wait, this actually updates more frequently I noticed. Because it is dependent on the `useBalance` query which queries the balance on every new block.

PR #875, opened 2023-10-16
fix: :bug: remove ability to specify backup capital cost fee

james-a-morris:
We shouldn't be allowed to set a backup default cost config fee. This is indicative of a misconfured capital cost setting. We should have an intentional default for a given token or else we should fail.

A failure condition has been added to https://github.com/across-protocol/sdk-v2/pull/430 and we will need to bump the SDK before we merge this change.


dohaki (approved):
LGTM

nicholaspai (approved):

PR #880, opened 2023-10-20
improve: whitelist tokens from available endpoint

james-a-morris:
This adds token whitelisting to available routes / token routes endpoints.


pxrl (approved):
Works for me :+1:

PR #881, opened 2023-10-25
feat: add redis caching support for getBalance

james-a-morris:
This PR adds the ability for redis to be employed by the vercel edge functions.


api/_imports.ts
james-a-morris, line 1: This is required to polyfill `fetch` at a global level.

api/_utils.ts
```diff
@@ -620,45 +621,31 @@ export const isRouteEnabled = (
...
-      params: {
-        chainId,
-        account,
-        token,
-      },
-    }
-  );
-  // Return the balance
-  return BigNumber.from(response.data.balance);
+  const key = `balance_${chainId}-${account}-${token}`.toLowerCase();
```
james-a-morris, line 629: Our keys are in checksum case, but I don't think it would hurt to make the key case insensitive.

api/account-balance.ts
james-a-morris, line 1: We no longer need this endpoint.

Corollary - we can get rid of the cached coingecko endpoint as well.

api/limits.ts
james-a-morris, line 1: We gain the ability to go back to a single function interface with `getBalance`

api/_utils.ts
```diff
@@ -35,6 +35,7 @@ import {
   BLOCK_TAG_LAG,
 } from "./_constants";
 import { PoolStateResult } from "./_types";
+import { kv } from "@vercel/kv";
```
james-a-morris, line 38: The setup/destroy of this redis is fully handled by vercel. We it is federated in SFO, TX, and Singapore currently.

api/_utils.ts
```diff
@@ -620,45 +621,31 @@ export const isRouteEnabled = (
...
-/**
- * Resolves the cached balance of a given ERC20 token at a provided address. If no token is provided, the balance of the
- * native currency will be returned.
- * @param chainId The blockchain Id to query against
- * @param account A valid Web3 wallet address
- * @param token The valid ERC20 token address on the given `chainId`.
- * @returns A promise that resolves to the BigNumber of the balance
- */
-export const getCachedTokenBalance = async (
+export const getBalance = async (
```
dohaki, line 624: NIT: It feels somewhat unintuitive to me that this function `getBalance` returns cached balances per default. Maybe it makes sense to rename it back to `getCachedTokenBalance`?

api/_utils.ts
```diff
@@ -620,45 +621,31 @@ export const isRouteEnabled = (
...
+    return BigNumber.from(data);
+  } else {
+    const balance = await sdk.utils.getTokenBalance(
+      account,
+      token,
+      getProvider(Number(chainId)),
+      BLOCK_TAG_LAG // We should do this for consistency
+    );
+    await kv.set(key, balance.toString(), {
+      ex: 60 * 5, // 5 minutes
```
dohaki, line 641: Is it safe to set the expiry time to 5 minutes? We might also make this a parameter.

api/suggested-fees.ts
```diff
@@ -94,10 +95,11 @@ const handler = async (
...
       }
-      const isRecipientAContract = await sdk.utils.isContractDeployedToAddress(
-        recipientAddress,
-        provider
-      );
+      const isRecipientAContract =
+        (await sdk.utils.isContractDeployedToAddress(
+          recipientAddress,
+          provider
+        )) || true;
```
dohaki, line 102: Hmm doesn't this always return `true`? Maybe a leftover from manual testing?

dohaki (commented):
Left a few comments and questions

api/_imports.ts
```diff
@@ -0,0 +1,5 @@
+import * as fetch from "node-fetch";
+
+if (!globalThis.fetch) {
+  (globalThis as any).fetch = fetch;
```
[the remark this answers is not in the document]
james-a-morris: Yep - unfortunately `@vercel/kv` relies on `fetch`. There are pros/cons to trying to adapt, but the major pro here is that Vercel handles all the setup/teardown of these connections when we use their official package.

api/_utils.ts
```diff
@@ -1,3 +1,4 @@
+import "./_imports";
```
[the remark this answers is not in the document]
james-a-morris: Done

api/suggested-fees.ts
james-a-morris: This was for testing - it's since resolved.

api/_utils.ts
james-a-morris: Done

nicholaspai (approved):

api/_utils.ts
james-a-morris: This is currently what we have in our balance endpoint

api/suggested-fees.ts
```diff
@@ -94,10 +94,11 @@ const handler = async (
...
```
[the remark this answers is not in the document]
james-a-morris: This is outdated - since changed - it was an artifact from testing

api/_utils.ts
```diff
@@ -620,45 +621,31 @@ export const isRouteEnabled = (
...
+  if (sdk.utils.isDefined(data)) {
+    return BigNumber.from(data);
+  } else {
+    const balance = await sdk.utils.getTokenBalance(
+      account,
+      token,
+      getProvider(Number(chainId)),
+      BLOCK_TAG_LAG // We should do this for consistency
+    );
+    await kv.set(key, balance.toString(), {
```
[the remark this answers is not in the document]
james-a-morris: Good call - we should be able to continue even if the cache mechanism fails.

dohaki (approved):
Nice 👍

dohaki:
Closing as this was part of the performance improvements PR #1194

PR #882, opened 2023-10-26
improve: change recipient address

james-a-morris:
This PR moves the recipient address provided in `/limits` from the zero address to the `DEFAULT_SIMULATED_RELAYER_ADDRESS` - a known EOA.


api/limits.ts
```diff
@@ -150,7 +150,7 @@ const handler = async (
         ethers.BigNumber.from("10").pow(18),
         computedOriginChainId,
         Number(destinationChainId),
-        constants.ZERO_ADDRESS,
+        constants.DEFAULT_SIMULATED_RELAYER_ADDRESS,
```
pxrl, line 153: I think we need to revert to the old 0xb... address that we used. Setting the relayer as the recipient will short-circuit the ERC20 transfer and will under-estimate the gas.

james-a-morris: Fair call - updated: https://github.com/across-protocol/frontend-v2/pull/882/commits/81c4ea7d6e938fdca4043d0b7bd9d791caeddbc8

pxrl (approved):

PR #884, opened 2023-10-26
fix(suggested-fees): Use dest provider for contract determination

pxrl:


api/suggested-fees.ts
```diff
@@ -97,7 +98,7 @@ const handler = async (
       }
       const isRecipientAContract = await sdk.utils.isContractDeployedToAddress(
         recipientAddress,
-        provider
+        getProvider(destinationChainId)
```
james-a-morris, line 101: Good call. This is a subtle bug

james-a-morris (approved):

PR #885, opened 2023-10-26
improve(suggested-fees): Filter underlying simulation reasons

pxrl:


api/suggested-fees.ts
```diff
@@ -199,16 +199,22 @@ const handler = async (
...
+        computedOriginChainId,
+        destinationChainId,
+        recipientAddress,
+        tokenPrice,
+        message,
+        relayerAddress
+      );
+    } catch {
+      throw new InputError("Failed to estimate relayer fees");
+    }
```
james-a-morris, line 217: We should place this in `_utils.ts` where that function is instantiated. 

We'd just need to change on line ~532 of `_utils.ts` 

```
    // Resolve and transform the error
    const e = _e as Error;
    // We want to mask this error as an Input error.
    throw new InputError(e?.message);
```
to
```
    throw new InputError("Failed to estimate relayer fees.");
```

james-a-morris (approved):

api/_utils.ts
```diff
@@ -729,6 +727,14 @@ export function applyMapFilter<InputType, MapType>(
   }, []);
 }
 
+export function resolveEthersError(err: unknown): string {
+  return sdk.typeguards.isEthersError(err)
+    ? `${err.reason}: ${err.code}`
+    : sdk.typeguards.isError(err)
+    ? err.message
+    : "unknown error";
```
pxrl, line 735: I think the code construction here is OK, but this auto-formatting is ++annoying. I'd prefer it to be:
```suggestion
  return sdk.typeguards.isEthersError(err)
    ? `${err.reason}: ${err.code}`
    : sdk.typeguards.isError(err)
      ? err.message
      : "unknown error";
```

pxrl: Have manually overridden it via a prettier directive.

api/suggested-fees.ts
pxrl: Great - that was actually clobbering the ethers error in the catch statement that I had, so it definitely makes sense to put it there.

PR #886, opened 2023-10-26
feat: include new relayer address override

james-a-morris:
We need to be mindful of Synthetix's relayer when dealing with OP/Homestead routes of SNX. Specifically, our quoting should by default come from SNX's relayer unless specified otherwise where the SNX token <-> 10 or 1 is concerned.


dohaki (approved):
Nice LGTM

api/_constants.ts
```diff
@@ -95,6 +95,16 @@ const defaultRelayerFeeCapitalCostConfig: {
   },
 };
 
+export const defaultRelayerAddressOverride: Record<
+  string,
+  { relayer: string; destinationChains: number[] }
+> = {
+  SNX: {
+    relayer: "0x19cdc2b23af0cc791ca64dda5bfc094cddda31cd",
```
nicholaspai, line 103: This should be checksummed right?

james-a-morris: We don't directly compare against these addresses - but I agree it's more sanitized to checksum the address.

nicholaspai (approved):
LGTM

PR #887, opened 2023-10-27
improve: more effectively estimate suggested fees quote

james-a-morris:
We discovered a short circuit in the Vercel API's estimation of fee quotes. 

This was causing a quote of roughly 80K gas units to make a fill. 

The original recipient address being used on the FE is `0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B`, which will incur a storage cost during the initial ERC20 transfer of 145,000 gas units. Properly passing `recipientAddress` in the suggested-fees endpoint on prod will incur 120,000 gas units.

In either case, we now default our recipient address to incur the storage cost of the ERC20 initializing a user's balance in the token.

Thus, the cost calculation will be preserved regardless of `recipientAddress` being provided and using the Across UI will produce a more accurate gas estimation making it the cheaper alternative over aggregators. 

Closes: ACX-1643

### DIFF
--- a/src/views/Transactions/hooks/useSpeedUp.tsx
+++ b/src/views/Transactions/hooks/useSpeedUp.tsx
@@ -20,7 +20,7 @@ export function useSpeedUp(transfer: Deposit, token: Token) {
     transfer.sourceChainId,
     transfer.destinationChainId,
     token.symbol,
-    transfer.depositorAddr
+    transfer.recipientAddr
   );
 
   const [speedUpTxLink, setSpeedUpTxLink] = useState<string>("");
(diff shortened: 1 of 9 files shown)


nicholaspai (commented):
Why?

james-a-morris:
> Why?

We are issuing quotes with a simulation of `relayerFeePct` and `realizedFeePct` at zero. This simulation reduces overall gas costs compared to setting `relayerFeePct` and `realizedFeePct` at a non-zero level. 

I'd like to get this calculation more accurate, and we can do so by setting a non-zero fee pct

src/views/Transactions/hooks/useSpeedUp.tsx
```diff
@@ -19,7 +19,8 @@ export function useSpeedUp(transfer: Deposit, token: Token) {
     BigNumber.from(transfer.amount),
     transfer.sourceChainId,
     transfer.destinationChainId,
-    token.symbol
+    token.symbol,
+    transfer.depositorAddr
```
pxrl, line 23: Is this strict correct? The depositor address isn't necessarily the same as the recipient address.

pxrl (approved):
OK pending one query on the speed-up :+1:

james-a-morris:
> OK pending one query on the speed-up 👍

Whoops did not push the latest commit. Good eye

PR #888, opened 2023-10-30
refactor: align global css vars + remove some legacy code

dohaki:
This PR refactors a few things:
- Removes some legacy code
- Aligns the global CSS vars for colors with the ones in Figma
- Aligns the `COLORS` constant with above CSS vars
- Refactors `Text` component to use above constant

The motivation here is to reduce the usage of hardcoded HEX colors throughout the codebase. I know that we always wanted to refactor this but as I started to work on the new UI updates, I was in need of something like this.

Closes ACX-1645


james-a-morris (approved):
Good call. I was able to load this on my preview link and I don't notice anything broken. Is there a specific segment of this code that I could look at specifically that you want me to highlight in my review?

PR #889, opened 2023-10-30
refactor: use consistent button component 

dohaki:
Based #892 because I used Storybook for implementing the component.

This PR does the following:
- Introduction of new consistent `<Button />` component
  - While working on the new UI update, I again noticed that we don't have a consistent way of implementing buttons. Sometimes, we used the legacy component, and other times the V2 component. The V2 component was also inconsistent with the Figma designs and had a few visual bugs. This PR therefore adds a new component that uses the consistent color design token of PR #888.

Closes ACX-1646


james-a-morris (commented):
This is a huge diff. Can you outline a few locations in the code where I should drill down on?

dohaki:
> This is a huge diff. Can you outline a few locations in the code where I should drill down on? 

It's actually not that huge! Most of the diffs are dependency changes in the lock file.

I can try to split it out if you want.

dohaki:
> This is a huge diff. Can you outline a few locations in the code where I should drill down on?

@james-a-morris I split up this PR into https://github.com/across-protocol/frontend-v2/pull/892 and this one. So you can review #892 first and then this one.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,145 @@
...
+    gap: string;
+    borderRadius: string;
+    mobileSize: ButtonSize;
+  }
+> = {
+  lg: {
+    fontSize: "1.125rem",
+    lineHeight: "1.625rem",
+    height: "4rem",
+    padding: "0rem 2.5rem",
```
james-a-morris, line 39: Have we begun to switch to `rem` instead of `px` for height/padding values?

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,145 @@
...
+    padding: "0rem 1rem 0.0625rem 1rem",
+    gap: "0.25rem",
+    borderRadius: "20px",
+    mobileSize: "sm",
+  },
+};
+
+const BaseButton = styled.button<BaseButtonProps>`
+  // Text styles
+  font-family: "Barlow";
```
james-a-morris, line 66: Is this required to define?

james-a-morris (commented):
Added a few comments. I don't notice any regressions when testing via the preview link.

src/components/Button/Button.tsx
dohaki: I noticed that in Figma we now use `rem` instead of `px` and I am assuming that Tim chose this setting for a reason. Converting these to `px` in your head when implementing the designs also doesn't really make sense to me, so I kept rems here as well.

I know that this leads to a bit of inconsistency but I think moving to `rem` makes sense. Wdyt?

src/components/Button/Button.tsx
dohaki: Yea I guess this is redundant

dohaki: Removed

src/components/Button/Button.tsx
dohaki: Ah wait, just checked again. This is a setting on my side that led to this 😓 I think in that case it makes sense to use px for now. And move over to rem later if we decide to do so

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,144 @@
...
+    gap: ${({ size = "lg" }) => sizeMap[sizeMap[size].mobileSize].gap};
+    border-radius: ${({ size = "lg" }) =>
+      sizeMap[sizeMap[size].mobileSize].borderRadius};
+  }
+`;
+
+export const UnstyledButton = styled(BaseButton)`
+  border: none;
+  padding: 0;
+  height: auto;
```
james-a-morris, line 106: Should we use `fit-content`?

james-a-morris (approved):
This looks good to me. I don't have much else to add & everything works well on the local storybook instance I loaded. 

@dohaki should we look at maybe re-introducing chromatic?

dohaki:
> This looks good to me. I don't have much else to add & everything works well on the local storybook instance I loaded.
> 
> @dohaki should we look at maybe re-introducing chromatic?

Yea we could. I don't recall why we stopped using it 🤔

amateima (approved):

PR #890, opened 2023-10-30
refactor: Align /suggested-fees inputs w/ SpokePool names

pxrl:
To reduce misalignment between the SpokePool deposit() input parameters and the parameter names supplied to /suggested-fees, rename the following:

 recipientAddress -> recipient
 relayerAddress -> relayer

This leaves only `token` as the variable name that differs.


pxrl:
(nb. proposing this because it left me scratching my head for a while - it's intuitive (and simpler in code) to ensure that the input parameter names align as much as possible.

james-a-morris (approved):
Fair enough!

PR #891, opened 2023-10-30
fix: Clarify misleading error message

pxrl:


api/limits.ts
```diff
@@ -108,12 +108,9 @@ const handler = async (
...
     // If any of the above fails or the route is not enabled, we assume that the
-    if (
-      disabledL1Tokens.includes(l1Token.toLowerCase()) ||
-      tokenDetailsResult.status === "rejected" ||
-      routeEnabledResult.status === "rejected" ||
-      !routeEnabledResult.value
-    ) {
+    if (tokenDetailsResult.status === "rejected" || routeEnabledResult.status === "rejected") {
+      throw new InputError(`Unable to query route.`);
+    } else if (!routeEnabledResult.value || disabledL1Tokens.includes(l1Token.toLowerCase())) {
```
pxrl, line 113: fwiw we could actually check `disabledL1Tokens.includes(...)` before making this query. I guess the reason that we don't is just to keep the error messages grouped together.

james-a-morris (approved):
lgtm

PR #893, opened 2023-10-31
feat: new deposit status route

dohaki:
Closes ACX-1628

Requires #889 and #892

This PR implements the redesigned deposit status as a new route to provide better UX. The diff got blown up a bit by that because a lot of boilerplate code had to be added etc. I outlined the main changes with some comments.


src/hooks/useLocalPendingDeposits.ts
dohaki, line 1: This was refactored to `src/utils/local-deposits`. The reasoning behind this:
- There was no need for these helpers to be inside a hook
- Allows us to use these outside of react components

src/views/Bridge/hooks/useBridge.ts
dohaki, line 1: Because the deposit status tracking is now a separate route, the logic in this hook got a lot more cleaner and easier to reason about.

src/views/DepositStatus/components/DepositStatusLowerCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+  fromBridgePagePayload?: FromBridgePagePayload;
+};
+
+export function DepositStatusLowerCard({
+  fromChainId,
+  toChainId,
+  bridgeTokenSymbol,
+  fromBridgePagePayload,
+}: Props) {
+  const { quote } = fromBridgePagePayload || {};
```
dohaki, line 27: With the current implementation, this component will only display the fees for the respective user that did the transfer. In the future, we could add some fallback logic here that tries to fetch data from the Scraper or Contract. So this could also serve as a generic status page that can be visited by anyone.

src/views/DepositStatus/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+    status: "pending" as const,
+    filled: "0",
+    sourceChainId: sendDepositArgs.fromChain,
+    destinationChainId: sendDepositArgs.toChain,
+    assetAddr: sendDepositArgs.tokenAddress,
+    depositorAddr: utils.getAddress(fromBridgePagePayload.account),
+    recipientAddr: sendDepositArgs.toAddress,
+    message: sendDepositArgs.message || "0x",
+    amount: BigNumber.from(sendDepositArgs.amount).toString(),
+    depositTxHash: data.depositTxReceipt.transactionHash || "0x",
```
james-a-morris, line 25: Should we throw if the deposit hash doesn't exist?

src/views/DepositStatus/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function convertForFillQuery(
+  data: Awaited<ReturnType<typeof getFillByDepositTxHash>>,
+  fromBridgePagePayload: FromBridgePagePayload
+) {
+  const { sendDepositArgs, quote } = fromBridgePagePayload;
+
+  return {
+    depositId: Number(
+      data.depositByTxHash.parsedDepositLog.args.depositId || 0
```
james-a-morris, line 58: Same here ^

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
```diff
@@ -0,0 +1,361 @@
...
+`;
+
+const AnimatedDividerToChain = styled(AnimatedDivider)<{
+  status: DepositStatus;
+}>`
+  background: ${({ status }) =>
+    status === "depositing"
+      ? "#9daab3"
+      : status === "filling"
+      ? "#FFFFFF"
```
james-a-morris, line 263:
```suggestion
      ? "#ffffff"
```

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
```diff
@@ -0,0 +1,361 @@
...
+  const { elapsedSeconds: fillTxElapsedSeconds } = useElapsedSeconds(
+    depositTxCompletedTime,
+    fillTxCompletedTime
+  );
+
+  const status = !depositTxCompletedTime
+    ? "depositing"
+    : !fillTxCompletedTime
+    ? "filling"
+    : "filled";
```
james-a-morris, line 78: Thoughts on codifying this as an `enum` or `type BridgeStatus = "depositing" | "filling" | "filled"`

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
```diff
@@ -0,0 +1,361 @@
...
+  [ChainId.MAINNET]: <EthereumGrayscaleLogo />,
+  [ChainId.ZK_SYNC]: <ZkSyncGrayscaleLogo />,
+  [ChainId.BASE]: <BaseGrayscaleLogo />,
+  // testnets
+  [ChainId.GOERLI]: <EthereumGrayscaleLogo />,
+  [ChainId.ARBITRUM_GOERLI]: <EthereumGrayscaleLogo />,
+  [ChainId.MUMBAI]: <EthereumGrayscaleLogo />,
+  [ChainId.ZK_SYNC_GOERLI]: <ZkSyncGrayscaleLogo />,
+  [ChainId.BASE_GOERLI]: <BaseGrayscaleLogo />,
+};
```
james-a-morris, line 39: Should we place this in constants? This feels like a more general structure

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
```diff
@@ -0,0 +1,361 @@
...
+import { ChainId, QUERIESV2 } from "utils";
+
+import { useElapsedSeconds } from "../hooks/useElapsedSeconds";
+import { useDepositTracking } from "../hooks/useDepositTracking";
+import { DepositTimesCard } from "./DepositTimesCard";
+import { ElapsedTime } from "./ElapsedTime";
+import { DepositStatus, FromBridgePagePayload } from "../types";
+
+const grayscaleLogos: {
+  [key: number]: JSX.Element;
```
james-a-morris, line 25: Should we use `Record< >` or keep the index format for consistency? I've always thought `Record<>` is a good tradeoff, and it is more succinct - but I'm not sure the benefits of this type

src/views/DepositStatus/DepositStatus.tsx
james-a-morris, line 1: Do we handle the case where a tx hash is passed that is not an across related transaction?

src/views/Bridge/hooks/useBridgeAction.ts
```diff
@@ -108,76 +97,30 @@ export function useBridgeAction(
...
+      {
+        fromBridgePagePayload: {
+          sendDepositArgs: frozenPayload,
+          quote: frozenQuote,
+          referrer,
+          account: frozenAccount,
+          timeSigned: Date.now(),
+          tokenPrice,
+        },
       }
```
james-a-morris, line 123: Is there a cleaner way to do this? Embedding the class constructor in the string formatter was a tad hard to follow.

james-a-morris (commented):
Left a few comments

src/views/DepositStatus/utils.ts
dohaki: We can actually remove this because this was initially a workaround for some type issues that I fixed already.

src/views/DepositStatus/utils.ts
dohaki: We can actually remove this because this was initially a workaround for some type issues that I fixed already.

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
dohaki: So there is already a type `src/views/DepositStatus/types.ts`:
```
export type DepositStatus = "depositing" | "filling" | "filled";
```

And I think it suffices here in this case that the type gets inferred correctly.

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
dohaki: I think using the COLORS constants makes sense here. WIll change

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
dohaki: Hmm not sure... This would require us to change the file extension of the `constants.ts` file to `constants.tsx`. And currently these logos are only used here

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
dohaki: Yea agree! I copy-pasted a lot from the old components and the index format has been used there. Will change

src/views/DepositStatus/DepositStatus.tsx
dohaki: Not explicitly... It would just show a never-ending loading screen

src/views/Bridge/hooks/useBridgeAction.ts
dohaki: I could pull it out of the templated string maybe?

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/893/commits/95cb1ba1940d0e32f3510ec77e0893ff7ff7ef0f

src/views/DepositStatus/utils.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/893/commits/84e229edea6e1319393d5f4e4bdad800e510da67

src/views/DepositStatus/utils.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/893/commits/84e229edea6e1319393d5f4e4bdad800e510da67

src/views/Bridge/hooks/useBridgeAction.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/893/commits/17901501c0936979b6325e8901891c193bd7c7f9

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
james-a-morris: Ah shoot good point with the need for a file change.

src/views/DepositStatus/DepositStatus.tsx
james-a-morris: Could we maybe check that the txn is related to the spoke pool contract?

james-a-morris (commented):
One more question: https://github.com/across-protocol/frontend-v2/pull/893#discussion_r1378864556

src/views/DepositStatus/DepositStatus.tsx
dohaki: So I added some handling of invalid deposit tx hash here https://github.com/across-protocol/frontend-v2/pull/893/commits/36d753c0f052e05f9e17af62d372589940969798. This covers the cases:
- Invalid tx hash format
- Valid tx hash format and mined but not from a SpokePool

One case that this does not cover is:
- Valid tx hash format but not mined yet

I am not sure how we can check whether a not mined tx is from a SpokePool or not.

james-a-morris (approved):
The design matches the spec + all outstanding questions have been answered. LGTM

PR #898, opened 2023-11-02
fix: quote update after user connected wallet via bridge action button

dohaki:
If a user connects a wallet via the primary CTA button on the bridge page, it leads to the quote not being updated. Ideally, the call to `connect()` should be in a separate handler/mutation and not inside the hook `useBridgeAction`. But we can refactor that in a separate PR.

Fixes ACX-1668


mrice32 (approved):
LGTM!

PR #899, opened 2023-11-02
fix: throw speed up mutation if depositor does not match signer

dohaki:
Fixes ACX-1667


mrice32 (approved):
Looks great! Nice, clean fix!

PR #900, opened 2023-11-02
feat: allow OP rebate on chainId=10 destination

james-a-morris:
This PR is aimed at adding an OP rebate on the fee preview.


src/views/Bridge/hooks/useEstimatedTable.ts
```diff
@@ -8,54 +8,73 @@ import {
...
+  destinationChainId: number,
   gasFee?: BigNumber,
   bridgeFee?: BigNumber
 ) {
   const [isDetailedFeesAvailable, setIsDetailedFeesAvailable] = useState(false);
 
+  const rewardToken = useMemo(
+    () => getToken(destinationChainId === 10 ? "OP" : "ACX"),
+    [destinationChainId]
+  );
```
dohaki, line 26: We need a way to enabled/disable this via a config somehow. Because the OP rebate program is currently not fully ready yet. But with this implementation (AFAICT), we will already show the OP rebates if the destination is Optimism right?

dohaki: Wdyt of parameterizing the `rewardToken` and maybe `availableRewardPercentage` in https://github.com/across-protocol/frontend-v2/pull/900/files#diff-a6b83e50a5e77bc3e7c2c5cd7c8b356b6fdefbafc174174649e04ae2846edb70R61? Or we could add another hook `useRewardToken`:
```tsx
function useRewardToken() {
  const rewardToken = useMemo(
    () => {
      if (OP_REBATE_ENABLED) {
         return getToken(destinationChainId === 10 ? "OP" : "ACX")
      }
      return getToken("ACX")
    },
    [destinationChainId]
  );

  const availableRewardPercentage = useQuery(
    ["rewards-percentage", rewardToken.symbol],
    async () => {
      if (rewardToken.symbol === "OP") {
        // OP rebate percentage
      }
      // fetch specialized endpoint that only returns referral tier
      // ACX rebate percentage
    },
    {
      // use more aggressive caching options
    }
  )
}
```

james-a-morris: Good call - I'm not a huge fan of using `useQuery` in this context as it won't always be used for querying an endpoint, but this would be a good candidate

src/views/Bridge/hooks/useEstimatedTable.ts
```diff
@@ -8,54 +8,73 @@ import {
...
+
   const { referrer } = useReferrer();
   const { summary: referralSummary } = useReferralSummary(referrer);
```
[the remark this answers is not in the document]
james-a-morris: I'm waiting for Alex to push this forward. He mentioned we should be good to start testing when the endpoint is available starting on 2023-11-07 @amateima

src/views/Bridge/hooks/useEstimatedTable.ts
james-a-morris: https://github.com/across-protocol/frontend-v2/pull/900/commits/2442363be4e4c6c9898ad4ea820ece6248d2c491

src/hooks/useRewardToken.ts
```diff
@@ -0,0 +1,52 @@
+import { useConnection } from "hooks";
+import { useMemo } from "react";
+import { useQuery } from "react-query";
+import { getToken, rebateTokensAvailable } from "utils";
+import { useReferralSummary } from "./useReferralSummary";
+import useReferrer from "./useReferrer";
+
+export function useRewardToken(destinationChainId: number) {
+  const rewardToken = useMemo(() => {
+    if (rebateTokensAvailable.includes("OP")) {
```
james-a-morris, line 10: We now have a record of available rebate tokens to make things a bit more robust.

dohaki: Nice 👍

src/views/Bridge/components/EstimatedTable.tsx
```diff
@@ -36,23 +35,33 @@ const PriceFee = ({
...
+        <Text size="md" color="grey-400">
+          {`$${formatUnits(baseCurrencyFee, 18)}`}
+        </Text>
+      </>
     )}
     {tokenFee ? (
-      <Text size="md" color={highlightTokenFee ? "primary" : "white"}>
+      <Text
+        size="md"
+        color={highlightTokenFee ? (isACX ? "primary" : "op-red") : "white"}
```
dohaki, line 63: Not sure if this text is supposed to be red
image

dohaki (commented):
Looks good 💪 Left one small thing

src/views/Bridge/components/EstimatedTable.tsx
james-a-morris: I was looking at a different frame: done: https://github.com/across-protocol/frontend-v2/pull/900/commits/3f5d09be22eb62382c72a7fa642a2fa71eff93e3

dohaki (approved):
Looking gooood

src/views/Bridge/hooks/useEstimatedTable.ts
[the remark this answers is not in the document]
james-a-morris: Completed via a recent commit

PR #903, opened 2023-11-03
fix(suggested-fees): Defer default relayer assignment

pxrl:
We should use getDefaultRelayerAddress() now because it contains conditional logic for which relayer to select based on the token being bridged.


dohaki (approved):
LGTM

PR #906, opened 2023-11-06
fix: add stale-while-revalidate to limits

james-a-morris:
We should add `stale-while-revalidate` to our caching policy. The UX is improved by not requiring an unlucky user to request a fresh, uncached response. 

Additionally, we have boilerplate on all endpoints to return a response. We should centralize our returned headers. This PR adds a small response function that is tested in limits.

---

Requirements Fulfilled:
- [x]  Returned value on `/limits` is not more than 5 minutes
- [x] Recalculation is computed in the background.

---

Partially closes ACX-1692

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1259,14 +1259,17 @@ export function sendResponse(
   response: VercelResponse,
   body: Record<string, unknown>,
   statusCode = 200,
-  cache?: number
+  cacheSeconds = 300,
+  staleWhileRevalidateSeconds?: number
 ) {
-  // We only want to cache if the status code is 200 and the
-  // caching time has been defined.
-  if (statusCode === 200 && sdk.utils.isDefined(cache)) {
+  if (cacheSeconds > 0) {
+    // Per Vercel's documentation, Vercel will only cache a response when
+    // the status code is 200. I.e. we can always set a cache policy
     response.setHeader(
       "Cache-Control",
-      `s-max-age=${cache}, stale-while-revalidate=${cache}`
+      (staleWhileRevalidateSeconds ?? 0) > 0
+        ? `s-maxage=${cacheSeconds}, stale-while-revalidate=${staleWhileRevalidateSeconds}`
+        : `s-maxage=${cacheSeconds}`
     );
   }
   return response.status(statusCode).json(body);
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1261,7 +1261,6 @@ export function sendResponse(
   statusCode = 200,
   cache?: number
 ) {
-  response.setHeader("Content-Type", "application/json");
   // We only want to cache if the status code is 200 and the
   // caching time has been defined.
   if (statusCode === 200 && sdk.utils.isDefined(cache)) {
@@ -1270,5 +1269,5 @@ export function sendResponse(
       `s-max-age=${cache}, stale-while-revalidate=${cache}`
     );
   }
-  return response.status(statusCode).send(body);
+  return response.status(statusCode).json(body);
 }
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1260,16 +1260,16 @@ export function getDefaultRelayerAddress(
 export function sendResponse(
   response: VercelResponse,
   body: Record<string, unknown>,
-  statusCode = 200,
-  cacheSeconds = 300,
-  staleWhileRevalidateSeconds?: number
+  statusCode: number,
+  cacheSeconds: number,
+  staleWhileRevalidateSeconds: number
 ) {
   if (cacheSeconds > 0) {
     // Per Vercel's documentation, Vercel will only cache a response when
     // the status code is 200. I.e. we can always set a cache policy
     response.setHeader(
       "Cache-Control",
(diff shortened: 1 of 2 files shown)


api/_utils.ts
```diff
@@ -1246,3 +1246,28 @@ export function getDefaultRelayerAddress(
...
+  cache?: number
+) {
+  // We only want to cache if the status code is 200 and the
+  // caching time has been defined.
+  if (statusCode === 200 && sdk.utils.isDefined(cache)) {
+    response.setHeader(
+      "Cache-Control",
+      `s-max-age=${cache}, stale-while-revalidate=${cache}`
+    );
+  }
```
james-a-morris, line 1271: It's good practice to set this when the status code is 2xx (success). We only use 200 for success in the API.

api/limits.ts
```diff
@@ -235,8 +236,7 @@ const handler = async (
       message: "Response data",
       responseJson,
     });
-    response.setHeader("Cache-Control", "s-maxage=300");
-    response.status(200).json(responseJson);
+    sendResponse(response, responseJson, 200, 300);
```
james-a-morris, line 239: Send a 200 response with 300 seconds of caching

api/_utils.ts
```diff
@@ -1246,3 +1246,28 @@ export function getDefaultRelayerAddress(
...
+
+/**
+ * Performs the needed function calls to return a Vercel Response
+ * @param response The response client provided by Vercel
+ * @param body A payload in JSON format to send to the client
+ * @param statusCode The status code - defaults to 200
+ * @param cache The cache time in seconds - defaults to 300
+ * @returns The response object
+ */
+export function sendResponse(
```
james-a-morris, line 1258: The logic in this function is essentially copied/pasted throughout the codebase. I think we should look to centralizing this as adding it in more locations just makes it error-prone.

api/_utils.ts
```diff
@@ -1246,3 +1246,28 @@ export function getDefaultRelayerAddress(
...
+  response: VercelResponse,
+  body: Record<string, unknown>,
+  statusCode = 200,
+  cache?: number
```
nicholaspai, line 1262: why would `cache` ever be undefined? Where does it default to 300?

james-a-morris: We've omitted the cache in previous endpoints. However, the 300 second default is now set.

james-a-morris: Optional parameters have now been removed.

api/_utils.ts
```diff
@@ -1246,3 +1246,33 @@ export function getDefaultRelayerAddress(
...
+  statusCode: number,
+  cacheSeconds: number,
+  staleWhileRevalidateSeconds: number
+) {
+  if (cacheSeconds > 0) {
+    // Per Vercel's documentation, Vercel will only cache a response when
+    // the status code is 200. I.e. we can always set a cache policy
+    response.setHeader(
+      "Cache-Control",
+      staleWhileRevalidateSeconds > 0
```
nicholaspai, line 1272: Why do you check if `staleWhileRevalidateSeconds` is > 0 now? since its required, just set it equal to the passed in value?

nicholaspai (approved):
https://github.com/across-protocol/frontend-v2/pull/906/files#r1385076852

api/_utils.ts
james-a-morris: What if it's negative? 

[RFC 7234 ](https://datatracker.ietf.org/doc/html/rfc7234)(which governs HTTP caching behavior) doesn't define behavior for negative values - max-age/stale-while-revalidate are defined as non-negative. 

I don't expect ever to pass a negative value, but given that this ternary is within an endpoint that will be cached, I'd lean toward code correctness.

nicholaspai: why would it ever be negative..? why wouldn't you instead catch the negative at the beginning of the function?

james-a-morris: Follow up: the caching policy header is passed as `Caching-Policy: s-max-age=X, stale-while-revalidate=Y`.

If we encounter a browser that sees a negative `stale-while-revalidate` and decides to ignore the entire header, we're going to encounter a bunch of cache misses.

api/_utils.ts
pxrl: Nice. Is it worth setting some defaults on the input arguments? statusCode might be one where we could default to 200 right?

api/_utils.ts
```diff
@@ -1246,3 +1246,31 @@ export function getDefaultRelayerAddress(
...
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
+ * @see https://datatracker.ietf.org/doc/html/rfc7234
+ * @note Be careful to not set anything negative please. The comment in the fn explains why
+ */
+export function sendResponse(
+  response: VercelResponse,
+  body: Record<string, unknown>,
+  statusCode: number,
+  cacheSeconds: number,
+  staleWhileRevalidateSeconds: number
```
pxrl, line 1267: A/It might be nice if these were part of an object so they could be set by name, rather than inferred by order. Then you could probably nuke the comment as well.

options

PR #909, opened 2023-11-07
feat: updated paginated deposits table

dohaki:
Add a paginated variant of the new deposits table. Also only testable via Storybook.

Closes ACX-1623


src/components/DepositsTable/DataRow.tsx
```diff
@@ -28,12 +28,12 @@ type Props = {
...
   headerCells: HeaderCells;
   disabledColumns?: ColumnKey[];
-  onClickSpeedUp?: () => void;
+  onClickSpeedUp?: (deposit: Deposit) => void;
 };
 
 const config = getConfig();
 
-const MAX_PENDING_STATE_TIME_UNTIL_SLOW = 15 * 60; // 15 mins
+const MAX_PENDING_STATE_TIME_UNTIL_SLOW = 30 * 60; // 30 mins
```
james-a-morris, line 36: Nit: should this go in a constants file?

src/components/DepositsTable/DataRow.tsx
```diff
@@ -59,8 +59,9 @@ export function DataRow({
   );
   const isSlowRelay =
     deposit.status === "pending" &&
-    DateTime.fromSeconds(deposit.depositTime).diffNow("seconds").as("seconds") <
-      MAX_PENDING_STATE_TIME_UNTIL_SLOW;
+    Math.abs(
+      DateTime.fromSeconds(deposit.depositTime).diffNow("seconds").as("seconds")
+    ) > MAX_PENDING_STATE_TIME_UNTIL_SLOW;
```
james-a-morris, line 64: Should we allow any instance of future deposits to show in the FE?

src/components/DepositsTable/DepositsTable.tsx
```diff
@@ -4,9 +4,9 @@ import { HeadRow, headerCells, ColumnKey } from "./HeadRow";
 import { DataRow } from "./DataRow";
 import { Deposit } from "hooks/useDeposits";
 
-type Props = {
+export type Props = {
```
james-a-morris, line 7: Because it's exported, maybe we should name this `DepositsTableProps` or something similar

src/components/Pagination/Pagination.styles.tsx
```diff
@@ -38,6 +38,11 @@ export const PageSizeSelectButton = styled(UnstyledButton)`
   svg {
     margin-left: 6px;
   }
+
+  @media ${QUERIESV2.sm.andDown} {
+    margin-bottom: 1rem;
+    padding: 10px 20px;
```
james-a-morris, line 44: QQ: was this meant to be two different units?

james-a-morris (commented):
Left a few changes

src/components/DepositsTable/DataRow.tsx
dohaki: Can you explain what future deposits are?

src/components/Pagination/Pagination.styles.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/909/commits/2b3381fa75a92f260a474fe57dc85b44fe5f53a5

src/components/DepositsTable/DepositsTable.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/909/commits/2b3381fa75a92f260a474fe57dc85b44fe5f53a5

src/components/DepositsTable/DataRow.tsx
dohaki: https://github.com/across-protocol/frontend-v2/pull/909/commits/2b3381fa75a92f260a474fe57dc85b44fe5f53a5

james-a-morris (commented):
One outstanding thread: https://github.com/across-protocol/frontend-v2/pull/909#discussion_r1386989589

src/components/DepositsTable/DataRow.tsx
dohaki: The Scraper derives the deposit date from the block timestamp of the deposit tx. So there shouldn't be cases of future deposits, no?

dohaki: Something to note here is that `diffNow` uses the formula
```
DATE_ARG - NOW
```
i.e. if the passed-in date is in the past then it is negative. That's why I am using `Math.abs` here.

james-a-morris: Ah I see, if we would never see future deposits, is the `Math.abs(...)` on the diff just for convenience?

james-a-morris (approved):
Approving but still curious about that `Math.abs`

src/components/DepositsTable/DataRow.tsx
dohaki: If the deposit timestamp is in the past `diffNow` always returns a negative number actually. Because it uses
```
depositTime - nowTime
```
That's why I need that Math.abs

PR #914, opened 2023-11-15
feat: rewrite base referrals page

james-a-morris:
This PR aims to create a generic reward program template that we can map to the OP rebates & more.


james-a-morris:
@dohaki I'll need access to the new txns table for this to be considered Non-WIP

PR #919, opened 2023-11-23
hotfix: patch balancer sdk

dohaki:
Related to ACX-1722 but doesn't fully fix because depends on new version of balancer-sdk. Instead locally patch until fixed upstream.


james-a-morris (approved):
I am able to see the wstETH/ACX pool

PR #921, opened 2023-11-24
fix: `formatUSD` helper

dohaki:
The `formatUSD` didn't handle some edge cases correctly which led to a crash for small amounts.

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -291,6 +291,6 @@ export function humanReadableNumber(num: number, decimals = 0): string {
  * @note USD only has 2 decimal places of precision, so this will round to the nearest cent.
  */
 export function formatUSD(value: BigNumberish): string {
-  const formattedString = formatUnits(value, 18);
-  return numeral(formattedString).format("0,0.00");
+  const formattedString = ethers.utils.formatUnits(value, 18);
+  return numeral(Number(formattedString).toFixed(2)).format("0,0.00");
 }
--- a/src/utils/tests/format.test.ts
+++ b/src/utils/tests/format.test.ts
@@ -1,9 +1,12 @@
+import { utils } from "ethers";
+
 import {
   isValidString,
   shortenAddress,
   shortenString,
   shortenTransactionHash,
   smallNumberFormatter,
+  formatUSD,
 } from "../format";
 const VALID_ADDRESS = "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828";
 const TX_HASH =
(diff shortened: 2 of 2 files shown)


src/utils/format.ts
```diff
@@ -291,6 +291,6 @@ export function humanReadableNumber(num: number, decimals = 0): string {
  * @note USD only has 2 decimal places of precision, so this will round to the nearest cent.
  */
 export function formatUSD(value: BigNumberish): string {
-  const formattedString = formatUnits(value, 18);
```
dohaki, line 294: We should rename our own `formatUnits` utility function to something more descriptive like `formatUnitsAndTruncate`. Because it behaves quite different than ethers formatter function and I at least got confused often by that

src/utils/format.ts
```diff
@@ -291,6 +291,6 @@ export function humanReadableNumber(num: number, decimals = 0): string {
...
-  return numeral(formattedString).format("0,0.00");
```
dohaki, line 295: If the variable `formattedString` was a very small number like `0.000000000000000001`, the numeral lib returns `NaN` which cause a crash.

src/utils/tests/format.test.ts
```diff
@@ -55,3 +58,17 @@ describe("#smallNumberFormatter", () => {
...
+  it("should format a wei BigNumber to 2 decimal places", () => {
+    expect(formatUSD(utils.parseEther("1.1"))).toEqual("1.10");
+  });
+
+  it("should format a very small wei number to 2 decimal places", () => {
+    expect(formatUSD(1)).toEqual("0.00");
+  });
+
+  it("should format a large wei BigNumber to 2 decimal places", () => {
+    expect(formatUSD(utils.parseEther("100000.123456"))).toEqual("100,000.12");
```
dohaki, line 72: With the previous implementation this returned `100,000.00` but I would expect the decimals to not be truncated. Hence the usage of `ethers.utils.formatUnits` instead of our own

src/utils/tests/format.test.ts
```diff
@@ -55,3 +58,17 @@ describe("#smallNumberFormatter", () => {
...
   });
 });
+
+describe("#formatUSD", () => {
+  it("should format a wei BigNumber to 2 decimal places", () => {
+    expect(formatUSD(utils.parseEther("1.1"))).toEqual("1.10");
+  });
+
+  it("should format a very small wei number to 2 decimal places", () => {
+    expect(formatUSD(1)).toEqual("0.00");
```
dohaki, line 68: This would return `NaN` because the returned float has too many decimals. Now it returns as expected `0.00`

amateima (approved):

PR #922, opened 2023-11-24
fix: only show invalid amt when there's an amount

james-a-morris:
The UI has a bug where it marks an empty input field with a red "please input positive number" error. The UI should go back to the unchanged state when a user removed their amount.

Steps to reproduce.

1. Visit `https://across.to/bridge`
2. Enter a number into the amount field.
3. Remove the amount entirely

You will be left with this:
image


dohaki (approved):
LGTM

PR #926, opened 2023-11-29
feat: add OP rewards page

james-a-morris:
This PR adds the frame to track OP rebates rewards.


src/views/Rewards/hooks/useRewards.ts
```diff
@@ -10,6 +10,12 @@ export function useRewards() {
   const { isConnected, connect, account } = useConnection();
   const { summary, isLoading } = useRewardSummary("referrals", account);
 
+  console.log(summary);
```
dohaki, line 13: Can be removed

dohaki (approved):
LGTM 👍 Left one nit

PR #927, opened 2023-11-29
feat(referrals): handle empty, error, loading states

dohaki:
Adds some additional handling of empty, error and loading states on the referrals page


src/hooks/useRewards.ts
```diff
@@ -32,75 +30,43 @@ export function useRewards(
 
   const queryKey = enabledQuery
     ? rewardsQueryKey(program, account, limit, offset)
-    : "DISABLED_REFERRALS_KEY";
+    : "DISABLED_REWARDS_KEY";
```
james-a-morris, line 33: We're probably going to want to specify the program name within this disabled key type for consistency.

src/views/Transactions/components/AllTransactions.tsx
```diff
@@ -38,9 +38,10 @@ export function AllTransactions({ statusFilter }: Props) {
         <Text size="lg">Something went wrong... Please try again later.</Text>
         <SecondaryButton
           size="md"
-          onClick={() => {
-            queryClient.cancelQueries({ queryKey: ["deposits"] });
-            queryClient.invalidateQueries({ queryKey: ["deposits"] });
+          onClick={async () => {
+            await queryClient.cancelQueries({ queryKey: ["deposits"] });
```
james-a-morris, line 42: Do we want to encapsulate this in a useCallback?

james-a-morris, line 42: It also seems that these all come from the same hook. Could we export `exportQueries` from this?

And, OOC - why are we cancelling the query?

james-a-morris (approved):
Left a few comments

PR #930, opened 2023-11-30
fix: allow boba withdrawal

james-a-morris:
This PR allows withdrawing from the Boba Liquidity Pool. There was a regression when the pool itself was disabled, but this has now been corrected.


james-a-morris:
> @james-a-morris I tried to test a withdrawal but the `MAX` button on the `Remove` page just sets 0.
> 
> boba

Try this now

nicholaspai (approved):

pxrl (approved):
lgtm

PR #931, opened 2023-11-30
fix: change consistency of UI on txn page

james-a-morris:
This makes the UI consistent in the transactions page. For small monitors, we will see a huge width change between personal & all transactions.

### DIFF
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -1,14 +1,14 @@
 import styled from "@emotion/styled";
 import { useState } from "react";
 
-import { Text } from "components/Text";
-import Footer from "components/Footer";
-import { COLORS } from "utils";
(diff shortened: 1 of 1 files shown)


dohaki (commented):
@james-a-morris Did you talk with Ryan about this? Because I remember that he initially wanted to make this table as wide as possible... So if you have a wide enough screen, you should not need to scroll to see all columns

james-a-morris:
> @james-a-morris Did you talk with Ryan about this? Because I remember that he initially wanted to make this table as wide as possible... So if you have a wide enough screen, you should not need to scroll to see all columns

We can expand the table, if you use a normal desktop resolution then widths wildly change. I'll follow up with screenshots

james-a-morris:
@dohaki Ryan said this was okay - I'm waiting for a final signoff but for now this is open for another review round.

dohaki (approved):

PR #934, opened 2023-12-05
fix: resolve ui dropdown

james-a-morris:
Fixes a bug where the user cannot see the dropdown column when looking to resize a page.

Closes ACX-1735


dohaki (approved):
LGTM

PR #937, opened 2023-12-08
fix: base currency conversions for fee preview

dohaki:
Fixes base currency conversions on the fee preview for values > 1k. 

The root cause was again unexpected behavior of the helper `formatUSD`. I have seen quite a few bugs recently due to this function. From what I can tell, the function's naming suggests something else than it does. We should refactor the name soon.

### DIFF
--- a/src/views/Bridge/hooks/useEstimatedTable.ts
+++ b/src/views/Bridge/hooks/useEstimatedTable.ts
@@ -69,9 +69,9 @@ export function useEstimatedTable(
 
   const baseCurrencyConversions = useMemo(() => {
     const parseUsd = (usd?: number) =>
-      isDefined(usd) ? parseUnits(String(usd.toFixed(2)), 18) : undefined;
+      isDefined(usd) ? parseUnits(String(usd), 18) : undefined;
     const formatNumericUsd = (usd: BigNumber) =>
-      Number(ethersUtils.formatUnits(usd, 18));
+      Number(Number(ethersUtils.formatUnits(usd, 18)).toFixed(2));
     const gasFeeInUSD = convertL1ToBaseCurrency(gasFee);
     const bridgeFeeInUSD = convertL1ToBaseCurrency(bridgeFee);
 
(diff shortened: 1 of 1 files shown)


src/views/Bridge/hooks/useEstimatedTable.ts
```diff
@@ -69,9 +62,6 @@ export function useEstimatedTable(
   const hasDepositReferralReward = depositReward?.rewardAsL1.gt(0) ?? false;
 
   const baseCurrencyConversions = useMemo(() => {
-    const parseUsd = (usd?: number) =>
-      isDefined(usd) ? parseUnits(String(usd), 18) : undefined;
-    const formatNumericUsd = (usd: BigNumber) => Number(formatUSD(usd));
```
dohaki, line 74: It was failing here. If the arg `usd` was a number > 1k, then the `formatUSD` helper would format it to a human readable custom format like `1,000`. This would then return `NaN` if trying to parse to a `number`.

james-a-morris (approved):

src/views/Bridge/hooks/useEstimatedTable.ts
```diff
@@ -84,21 +74,22 @@ export function useEstimatedTable(
...
-    const numericGasFee = formatNumericUsd(gasFeeInUSD);
-    const numericBridgeFee = formatNumericUsd(bridgeFeeInUSD);
-    const numericReward = availableRewardPercentage
-      ? (numericBridgeFee + numericGasFee) *
-        Number(formatUnits(availableRewardPercentage, 18))
+    const rewardInUSD = availableRewardPercentage
+      ? gasFeeInUSD
+          .add(bridgeFeeInUSD)
+          .mul(availableRewardPercentage)
+          .div(fixedPointAdjustment)
```
james-a-morris, line 81: @dohaki does doing this impact the table estimates?

The reason the code was written this way was so that the numbers always add up to the total fee. There's a rounding issue that occasionally happens where the total fee is off by one cent and that's just a result of the conversion rates I think we should keep it as is and we should just change the bug which is in format USD to remove the commas.

james-a-morris (changes requested):
Left a comment

james-a-morris (approved):
LGTM - changes made post sync seem to work well

PR #938, opened 2023-12-11
feat: utility script for generating static routes file

dohaki:
The static `routes.json` files became hard to maintain with the recent additions of new chains/tokens. This PR adds a script to make future additions/removals easier.

### DIFF
--- a/scripts/generate-routes.ts
+++ b/scripts/generate-routes.ts
@@ -1,0 +1,445 @@
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants-v2";
+import { utils as sdkUtils } from "@across-protocol/sdk-v2";
+import { utils } from "ethers";
+import { writeFileSync } from "fs";
+import * as prettier from "prettier";
+
+const { getDeployedAddress } = sdkUtils;
+
+const enabledRoutes = {
+  [CHAIN_IDs.MAINNET]: {
+    hubPoolChain: CHAIN_IDs.MAINNET,
+    hubPoolAddress: getDeployedAddress("HubPool", CHAIN_IDs.MAINNET),
+    hubPoolWethAddress: TOKEN_SYMBOLS_MAP.WETH.addresses[CHAIN_IDs.MAINNET],
+    acrossConfigStoreAddress: getDeployedAddress(
+      "AcrossConfigStore",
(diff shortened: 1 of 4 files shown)


scripts/generate-routes.ts
```diff
@@ -0,0 +1,445 @@
...
+          },
+          {
+            chainId: CHAIN_IDs.BASE,
+            tokens: ["WETH", "ETH", "USDC", "DAI", "BAL"],
+          },
+        ],
+      },
+      {
+        fromChain: CHAIN_IDs.POLYGON,
+        fromSpokeAddress: getDeployedAddress("SpokePool", CHAIN_IDs.POLYGON),
```
james-a-morris, line 160: Did we want to expose this for all chain_IDs?

scripts/generate-routes.ts
```diff
@@ -0,0 +1,445 @@
...
+      address: "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D",
+      enabledChains: [
+        CHAIN_IDs.MAINNET,
+        CHAIN_IDs.OPTIMISM,
+        CHAIN_IDs.POLYGON,
+        CHAIN_IDs.BASE,
+        CHAIN_IDs.ARBITRUM,
+      ],
+    },
+    routes: [
```
james-a-morris, line 39: This routes object feels like it could go into the constants-v2 repo. Thoughts on moving this there?

james-a-morris (commented):
Looks great!

scripts/generate-routes.ts
dohaki: Sorry, I don't fully understand the comment 🤔 Can you elaborate a bit more?

scripts/generate-routes.ts
dohaki: Hmm, not so sure about that tbh. Doesn't this just add more friction when trying to update/remove/add routes ad-hoc? AFAICT the FE is also the only place where this is currently used. So I don't see any good reason to move it to the contracts-v2 repo

scripts/generate-routes.ts
james-a-morris: Sorry later on in the code is I see we handle this. I misunderstood the code file.

james-a-morris (approved):
LGTM

PR #942, opened 2023-12-18
fix: resolve wsteth/acx staking pool

james-a-morris:
Resolve the UI bug when two tokens are displayed next to one another - only visible on the wrapped staked eth/ acx


dohaki (approved):
LGTM

PR #945, opened 2023-12-28
fix: correctly handle slow fills on deposit confirmation screen

dohaki:
Deposits with multiple fills, i.e. slow fills, were not correctly handled on the deposit confirmation screen.

### DIFF
--- a/src/utils/deposits.ts
+++ b/src/utils/deposits.ts
@@ -94,13 +94,22 @@ export async function getFillByDepositTxHash(
     )
   );
 
-  if (!filledRelayEvents.length) {
+  if (filledRelayEvents.length === 0) {
     throw new Error(
       `Could not find FilledRelay events for depositId ${depositId} on chain ${toChainId}`
     );
   }
 
-  const filledRelayEvent = filledRelayEvents[0];
+  const filledRelayEvent = filledRelayEvents.find((event) =>
+    event.args.amount.eq(event.args.totalFilledAmount)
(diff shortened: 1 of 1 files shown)


src/utils/deposits.ts
```diff
@@ -94,13 +94,22 @@ export async function getFillByDepositTxHash(
...
+  if (filledRelayEvents.length === 0) {
     throw new Error(
       `Could not find FilledRelay events for depositId ${depositId} on chain ${toChainId}`
     );
   }
 
-  const filledRelayEvent = filledRelayEvents[0];
+  const filledRelayEvent = filledRelayEvents.find((event) =>
+    event.args.amount.eq(event.args.totalFilledAmount)
+  );
```
james-a-morris, line 105: Are the `filledRelayEvents` in order? We could just get the last element and check instead of introducing this loop

james-a-morris (commented):
Left a comment

src/utils/deposits.ts
dohaki: Not sure if they are ordered tbh 🤔 But I would leave it like this. This makes it more explicit to only consider the fill that fully filled the deposit. Also, because we are using appropriate filters when querying the events, this should be less than < 5 max. items that we are looping through

james-a-morris: Sounds good!

james-a-morris (approved):

PR #947, opened 2024-01-04
fix: correctly label usdc.e for polygon

dohaki:
Fixes ACX-1805

### DIFF
--- a/scripts/generate-routes.ts
+++ b/scripts/generate-routes.ts
@@ -431,7 +431,11 @@ function generateRoutes(hubPoolChainId = 1) {
 }
 
 function getUsdcSymbol(chainId: number) {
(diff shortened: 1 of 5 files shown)


scripts/generate-routes.ts
```diff
@@ -431,7 +431,11 @@ function generateRoutes(hubPoolChainId = 1) {
 }
 
 function getUsdcSymbol(chainId: number) {
-  if ([CHAIN_IDs.ARBITRUM, CHAIN_IDs.OPTIMISM].includes(chainId)) {
+  if (
+    [CHAIN_IDs.ARBITRUM, CHAIN_IDs.OPTIMISM, CHAIN_IDs.POLYGON].includes(
```
james-a-morris, line 435: We have enough chains now that require the UI to map `USDC` -> `USDC.e`. Thoughts on making this a constant?

james-a-morris (approved):
This is a good change. I have nothing actionable. LGTM

PR #949, opened 2024-01-15
feat: enable op rewards claim

amateima:


src/hooks/useUnclaimedProofs.ts
```diff
@@ -0,0 +1,100 @@
...
+import { rewardProgramTypes } from "utils";
+
+export function useUnclaimedProofs(rewardsType: rewardProgramTypes) {
+  const { isConnected, account } = useConnection();
+  const queryIdentifier = useMemo(() => {
+    if (rewardsType === "referrals") {
+      return ["referral-rewards", "unclaimed"];
+    } else {
+      return ["op-rewards", "unclaimed"];
+    }
```
dohaki, line 16: We could move that into util fn
```ts
getUnclaimedProofsQueryKey(rewardsType, account)
```
and reuse it also in the queries below.

src/utils/merkle-distributor.ts
```diff
@@ -5,6 +5,7 @@ import {
   airdropWindowIndex,
   rewardsApiUrl,
   referralsStartWindowIndex,
+  rewardProgramTypes,
```
dohaki, line 8: NIT: Should prob be uppercase because it is a type def. But yae things like that should be caught by our linter. No action needed but we should beacklog that

src/views/RewardsProgram/hooks/useClaimModal.ts
```diff
@@ -1,23 +1,36 @@
-import { useUnclaimedReferralProofs } from "hooks/useUnclaimedReferralProofs";
+import { useUnclaimedProofs } from "hooks/useUnclaimedProofs";
 import { useWalletTokenImport } from "hooks/useWalletTokenImport";
-import { getConfig, getToken, rewardProgramTypes, rewardPrograms } from "utils";
+import {
+  ChainId,
```
dohaki, line 4: can be removed

dohaki (commented):
Nice work 👍 Left some initial comments on code. But would def wait for a review by @james-a-morris as he has been working also on the touched components.

src/utils/merkle-distributor.ts
amateima: We should backlog it, yes. I didn't want to rename it just to keep the PR small and concise

src/hooks/useUnclaimedProofs.ts
```diff
@@ -0,0 +1,93 @@
...
+    unclaimed,
+  };
+}
+
+async function fetchIsClaimedForIndices(
+  indices: { accountIndex: number; windowIndex: number }[],
+  rewardsType: rewardProgramTypes
+) {
+  const isClaimedResults = await Promise.all(
+    indices.map(async ({ accountIndex, windowIndex }) => {
```
james-a-morris, line 78: nit: we can use `mapAsync` from the SDK

src/utils/config.ts
```diff
@@ -186,10 +200,26 @@ export class ConfigClient {
...
+    rewardsType: constants.rewardProgramTypes,
+    signer?: Signer
+  ): AcrossMerkleDistributor {
+    let address = "";
+
+    if (rewardsType === "referrals") {
+      address = this.getMerkleDistributorAddress();
+    } else if (rewardsType === "op-rebates") {
+      address = this.getOpRewardsMerkleDistributorAddress();
+    }
```
james-a-morris, line 213: nit: thought on this:
```suggestion
    let address = rewardsType === "referrals" ?
        this.getMerkleDistributorAddress() : 
         this.getOpRewardsMerkleDistributorAddress();
```

src/utils/config.ts
```diff
@@ -186,10 +200,26 @@ export class ConfigClient {
...
+
+    let provider =
+      signer ??
```
james-a-morris, line 216: Should we use a signer?.provider to resolve the underlying provider if we're looking for a provider?

src/utils/merkle-distributor.ts
james-a-morris: Further to this - we have constants that are also lowercase. Google has a nice styleguide for TS and it includes making global constants/enums in uppercase:

https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

src/utils/query-keys.ts
```diff
@@ -154,3 +154,14 @@ export function isAirdropClaimedQueryKey(
...
+
+export function getUnclaimedProofsQueryKey(
+  rewardsType: rewardProgramTypes,
+  account?: string
+) {
+  if (rewardsType === "referrals") {
+    return ["referral-rewards", "unclaimed", account];
+  } else {
+    return ["op-rewards", "unclaimed", account];
+  }
```
james-a-morris, line 166: Thoughts on collapsing this to:
```suggestion
    return [rewardsType, "unclaimed", account];
```

james-a-morris (commented):
Left some comments/suggestions. I think this is good code

james-a-morris (approved):
I think we should have a sign off from @dohaki as well

PR #952, opened 2024-01-20
improve(error): improve error readout

james-a-morris:
## Issue Summary
The absence of `originChainId` parameter passed in `/limits` or `/suggested-fees` API calls can lead to a route collision due to a wildcard search, causing an "Invalid route" error even when valid routes exist.

## Immediate Action
Implement a custom error response indicating a route collision due to missing `originChainId`. This will provide clearer feedback, helping users distinguish between a true invalid route and a route collision.

## Long-term Resolution
Gradually enforce the inclusion of `originChainId` in API requests through a phased approach, with ample notice to partners for a smooth transition. This will prevent route collisions and improve the accuracy of route determination.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -198,7 +198,7 @@ export const resolveVercelEndpoint = () => {
   }
 };
 
-export const getTokenDetails = async (
+const _getTokenDetails = async (
   provider: providers.Provider,
   l1Token?: string,
   l2Token?: string,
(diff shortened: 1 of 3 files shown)


api/_utils.ts
```diff
@@ -604,7 +616,7 @@ export const isRouteEnabled = (
   fromToken: string
 ): boolean => {
   const enabled = ENABLED_ROUTES.routes.some(
-    ({ fromTokenAddress, fromChain, toChain, fromTokenSymbol }) =>
+    ({ fromTokenAddress, fromChain, toChain }) =>
```
james-a-morris, line 619: This is just a cleanup of an unused variable

api/limits.ts
```diff
@@ -118,11 +119,20 @@ const handler = async (
...
-    ) {
+    } else if (disabledL1Tokens.includes(l1Token.toLowerCase())) {
       throw new InputError(`Route is not enabled.`);
+    } else if (!routeEnabledResult.value) {
+      const routeCollision = await hasPotentialRouteCollision(
+        provider,
+        undefined,
+        token,
+        originChainId
+      );
```
james-a-morris, line 130: We only need to check if there was a route collision IF we have a route disabled warning

api/limits.ts
```diff
@@ -118,11 +119,20 @@ const handler = async (
...
+      throw new InputError(
+        routeCollision
+          ? `Potential route collision. Please specify the originChainId.`
```
nicholaspai, line 133: I think `route collision` is not going to be very clear to the user. Can we expand this message to make it more clear? I do think including "Please specify the originChainId" is great +1

nicholaspai (approved):
LGTM just one comment about the error msg

PR #953, opened 2024-01-25
feat: enable CORS in api

dohaki:
This enables CORS for our Vercel API. Also allow overriding the Vercel API endpoint via env var.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -776,3 +776,6 @@ export const gasExpenditureDeposit = BigNumber.from(90_000);
 
 // Used to determine whether to show the "delayed" warning in the deposits table
 export const pendingStateTimeUntilDelayed = 5 * 60; // 5 mins
+
+export const vercelApiBaseUrl =
+  process.env.REACT_APP_VERCEL_API_BASE_URL_OVERRIDE || "";
--- a/vercel.json
+++ b/vercel.json
@@ -36,7 +36,7 @@
           "key": "NEL",
           "value": "{\"report_to\":\"default\",\"max_age\":2592000,\"include_subdomains\":true,\"failure_fraction\":0.5}"
         },
-        { "key": "Access-Control-Allow-Origin", "value": "http://localhost" },
+        { "key": "Access-Control-Allow-Origin", "value": "*" },
         {
           "key": "Access-Control-Allow-Methods",
           "value": "GET,OPTIONS"
--- a/vercel.json
+++ b/vercel.json
@@ -35,6 +35,15 @@
         {
           "key": "NEL",
           "value": "{\"report_to\":\"default\",\"max_age\":2592000,\"include_subdomains\":true,\"failure_fraction\":0.5}"
+        },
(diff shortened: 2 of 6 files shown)


vercel.json
```diff
@@ -35,6 +35,15 @@
         {
           "key": "NEL",
           "value": "{\"report_to\":\"default\",\"max_age\":2592000,\"include_subdomains\":true,\"failure_fraction\":0.5}"
+        },
+        { "key": "Access-Control-Allow-Origin", "value": "*" },
```
james-a-morris, line 39: We may want to avoid the wildcard and explicitly add domains to this list on request.

james-a-morris (commented):
Left a comment. Looks clean otherwise

vercel.json
dohaki: Yea agree... Do you know how to whitelist localhost also?

dohaki: https://github.com/across-protocol/frontend-v2/pull/953/commits/ea13bfbbfa62fa0fa599c8356a7a565363905104

dohaki: I think we can use the wildcard for now and revisit if our quota really increases a lot by this change.

src/utils/constants.ts
```diff
@@ -776,3 +776,6 @@ export const gasExpenditureDeposit = BigNumber.from(90_000);
 
 // Used to determine whether to show the "delayed" warning in the deposits table
 export const pendingStateTimeUntilDelayed = 5 * 60; // 5 mins
+
+export const vercelApiBaseUrl =
```
james-a-morris, line 780: Have you looked at `resolveVercelEndpoint()` in `_utils.ts`?

james-a-morris (commented):
My only question is related to the URL override. I'm curious if we need it given the existing `resolveVercelEndpoint`. If we do need the override maybe we can use the ENV variable in that function

src/utils/constants.ts
dohaki: AFAICT `resolveVercelEndpoint` is used for the serverless API only, no? The use case of this variable is to be able to override the base URL in FE. The nice thing about that approach is that you can run the bundler without the need to run the Vercel API locally.

dohaki: I added this because I needed to test the CORS enable headers. This is only possible if we can override it in the FE bundle

james-a-morris: Ah right, I meant this function: https://github.com/across-protocol/frontend-v2/blob/master/src/utils/network.ts

dohaki: Ah ok. But `REACT_APP_VERCEL_URL` is set by Vercel right? I think these have different purposes 🤔 The linked function is used for referral links whereas this constant is used for the API

james-a-morris: Sounds good. LGTM - I'm just not a huge fan of adding override urls when Vercel provides them for us.

james-a-morris (approved):
https://github.com/across-protocol/frontend-v2/pull/953#discussion_r1468900336

PR #954, opened 2024-01-30
feat: set lpFeePct to 0

dohaki:
draft until tested

### DIFF
--- a/api/suggested-fees.ts
+++ b/api/suggested-fees.ts
@@ -27,7 +27,6 @@ import {
   positiveIntStr,
   boolStr,
   HUB_POOL_CHAIN_ID,
-  ENABLED_ROUTES,
   getSpokePoolAddress,
   getCachedTokenBalance,
   getDefaultRelayerAddress,
@@ -89,7 +88,7 @@ const handler = async (
       provider.getBlock("latest"),
       getTokenDetails(provider, undefined, token, originChainId),
     ]);
-    const { l1Token, hubPool, chainId: computedOriginChainId } = tokenDetails;
+    const { l1Token, chainId: computedOriginChainId } = tokenDetails;
 
     const tokenInformation = Object.values(TOKEN_SYMBOLS_MAP).find(
       (details) => details.addresses[HUB_POOL_CHAIN_ID] === l1Token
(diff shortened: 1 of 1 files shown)


api/suggested-fees.ts
```diff
@@ -233,18 +234,42 @@ const handler = async (
...
+
     const responseJson = {
       capitalFeePct: relayerFeeDetails.capitalFeePercent,
       capitalFeeTotal: relayerFeeDetails.capitalFeeTotal,
       relayGasFeePct: relayerFeeDetails.gasFeePercent,
       relayGasFeeTotal: relayerFeeDetails.gasFeeTotal,
-      relayFeePct: relayerFeeDetails.relayFeePercent,
-      relayFeeTotal: relayerFeeDetails.relayFeeTotal,
-      lpFeePct: realizedLPFeePct.toString(),
+      relayFeePct: totalRelayerFeePct.toString(), // capitalFeePct + gasFeePct + lpFeePct
```
nicholaspai, line 248: Can we also add a `totalFeePct = relayFeePct + lpFeePct` so we can eventually prompt integrators to use the new `totalFeePct`?

api/suggested-fees.ts
```diff
@@ -233,18 +234,42 @@ const handler = async (
...
+      relayFeeTotal: totalRelayerFee.toString(), // capitalFeeTotal + gasFeeTotal + lpFeeTotal
+      lpFeePct: "0", // Note: lpFeePct is now included in relayFeePct. We set it to 0 here for backwards compatibility.
       timestamp: parsedTimestamp.toString(),
       isAmountTooLow: relayerFeeDetails.isAmountTooLow,
       quoteBlock: blockTag.toString(),
       spokePoolAddress: getSpokePoolAddress(Number(computedOriginChainId)),
+      // V3's new fee structure
+      relayerFee: {
```
nicholaspai, line 256: maybe we should prefix these new fee structs with `v3`?

src/utils/math.ts
```diff
@@ -22,7 +22,7 @@ export function receiveAmount(
   deductionsSansRelayerGas: BigNumber;
   receivable: BigNumber;
 } {
-  const deductions = fees.relayerFee.total.add(fees.lpFee.total);
```
nicholaspai, line 25: can this be `totalFee` ? might have to add a new entry to `BridgeFees`

nicholaspai (commented):
I think we should add a `totalFee` that we use internally to mark the new total fee, which I think is renamed `relayFee`. I think total fee might be less confusing.

We shouldn't change the API's "relayerFeePct` field, which you haven't done so +1 there

nicholaspai (approved):

PR #962, opened 2024-02-21
chore: remove airdrop page

dohaki:
Closes ACX-1863


james-a-morris (approved):
LGTM - we should also nix the Discord Authentication related components as these were only around for the airdrop

PR #964, opened 2024-02-21
feat(v3): funds deposited for V3 events

james-a-morris:
This PR accounts for the V3 events when a user wishes to check their deposit confirmation status. This is backward compatible with V2 events

### DIFF
--- a/src/utils/deposits.ts
+++ b/src/utils/deposits.ts
@@ -29,17 +29,21 @@ export function parseFundsDepositedLog(
       return [];
     }
   });
-  const v3Deposit = parsedLogs.find((log) => log.name === "V3FundsDeposited");
   const v2Deposit = parsedLogs.find((log) => log.name === "FundsDeposited");
-
-  if (!(isDefined(v3Deposit) || isDefined(v2Deposit))) {
+  const v3Deposit = parsedLogs.find((log) => log.name === "V3FundsDeposited");
+  if (isDefined(v3Deposit)) {
+    return {
+      ...v3Deposit,
+      isV2: false,
+    };
+  } else if (isDefined(v2Deposit)) {
+    return {
+      ...v2Deposit,
+      isV2: true,
+    };
+  } else {
     return undefined;
   }
-
-  return {
-    ...(v3Deposit ?? v2Deposit!),
-    isV2: isDefined(v2Deposit),
-  };
 }
 
 export async function getDepositByTxHash(
--- a/src/utils/deposits.ts
+++ b/src/utils/deposits.ts
@@ -20,7 +20,7 @@ export function parseFundsDepositedLog(
     topics: string[];
     data: string;
   }>
-): (LogDescription & { isV2: boolean }) | undefined {
+): LogDescription | undefined {
   const spokePoolIface = SpokePool__factory.createInterface();
   const parsedLogs = logs.flatMap((log) => {
     try {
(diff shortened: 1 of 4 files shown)


src/utils/deposits.ts
```diff
@@ -26,7 +29,17 @@ export function parseFundsDepositedLog(
       return [];
     }
   });
-  return parsedLogs.find((log) => log.name === "FundsDeposited");
+  const v3Deposit = parsedLogs.find((log) => log.name === "V3FundsDeposited");
+  const v2Deposit = parsedLogs.find((log) => log.name === "FundsDeposited");
```
nicholaspai, line 33: Does this handle a V2 and V3 deposit in the same txn..? multicalled tg for example

src/utils/deposits.ts
```diff
@@ -76,45 +89,96 @@ export async function getFillByDepositTxHash(
...
+    // that corresponds to the deposit we are looking for.
+    const filledRelayEvent = filledRelayEvents[0];
+
+    if (!filledRelayEvent) {
+      throw new Error(
+        `Could not find FilledRelay event that fully filed depositId ${depositId} on chain ${toChainId}`
+      );
+    }
+
+    const fillTxBlock = await filledRelayEvent.getBlock();
```
nicholaspai, line 176: Can you refactor this bit at all with the V2 block above?

dohaki (approved):

src/utils/deposits.ts
james-a-morris: No - but would that be better served in a separate PR? That seems out of scope for this status page given that it's main use-case is to track the status of single user deposits.

mrice32: I think the either-or that James has here is totally fine since we should never make a multicall from the FE, right?


I do think the casework and logic is a little more complex than it needs to be. Thoughts on something a little clearer, like:
```ts
if (isDefined(v3Deposit)) {
  return { ...v3Deposit, isV2: false };
}

if (isDefined(v2Deposit)) {
  return { ...v2Deposit, isV2: true };
}

// Not strictly necessary, but just here to be explicit.
return undefined;
```

src/utils/deposits.ts
```diff
@@ -76,8 +89,10 @@ export async function getFillByDepositTxHash(
 
   const depositId = Number(parsedDepositLog.args.depositId);
   const depositor = String(parsedDepositLog.args.depositor);
+  const isV2 = parsedDepositLog.isV2;
   const destinationSpokePool = config.getSpokePool(toChainId);
-  const filledRelayEvents = await destinationSpokePool.queryFilter(
+
+  const v2FilledRelayEvents = await destinationSpokePool.queryFilter(
```
mrice32, line 95: Minor point (and not necessary to change now), but I think it'd be a bit easier to reason about if we just branched with an if(isV2) at the top and then had separate logic branches for all the v2 and v3 logic instead of querying both events and then using a bunch of ternaries to check the version in each expression.

mrice32 (approved):
LGTM!

src/utils/deposits.ts
james-a-morris: @mrice32 done https://github.com/across-protocol/frontend-v2/pull/964/commits/0be58a5ef3aa4ca38438d3cc1c2b1cce36cf4d97

src/utils/deposits.ts
```diff
@@ -17,7 +20,7 @@ export function parseFundsDepositedLog(
     topics: string[];
     data: string;
   }>
-) {
+): (LogDescription & { isV2: boolean }) | undefined {
```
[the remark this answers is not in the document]
james-a-morris: Agreed. Here's a rework https://github.com/across-protocol/frontend-v2/pull/964/commits/f2f37ef3df610378256bbb9cfa84904f0d3c032d

src/utils/deposits.ts
james-a-morris: I had that in a previous commit. Changed here: https://github.com/across-protocol/frontend-v2/pull/964#discussion_r1498016327

james-a-morris:
Merging with offline signoff from @nicholaspai

PR #970, opened 2024-02-22
fix: support v3 speed up

dohaki:
Support for speed up of v3 deposits


james-a-morris (commented):
Will circle back when the tests pass

james-a-morris (approved):
LGTM - it's a good call to keep the V2/V3 flow separate +1

PR #975, opened 2024-02-27
fix: only show speed up for pending

dohaki:


src/components/DepositsTable/cells/ActionsCell.tsx
```diff
@@ -43,7 +43,7 @@ export function ActionsCell({ deposit, onClickSpeedUp }: Props) {
   }, [deposit, onClickSpeedUp]);
 
   const speedUp =
-    !isDelayed && !isProfitable ? (
+    !isDelayed && !isProfitable && deposit.status === "pending" ? (
```
james-a-morris, line 46: Don't we implicitly check this in isProfitable?

pxrl, line 46: LGMTM

pxrl (approved):
LGTM

src/components/DepositsTable/cells/ActionsCell.tsx
dohaki: No, only in `isDelayed` but because it is negated, we need to check again

pxrl: Doesn't seem so. It seems a bit counter-intuitive to evaluate profitability on relay that was already filled though I guess.

speedup

dohaki: We only use the isProfitable here.. Yea I guess it makes sense to add `status=="pending"` when checking for profitability.

dohaki: chenged here https://github.com/across-protocol/frontend-v2/pull/975/commits/5aa7a1d3f3a4514164926cff84d3ca1095d2996b

james-a-morris (approved):

src/components/DepositsTable/cells/ActionsCell.tsx
dohaki: Hmm, I had to re-add this check even though I implicitly check the pending status in `isProfitable`.

PR #978, opened 2024-03-06
feat: update spokepoolverifier address + deprecate v2

dohaki:
UPDATE: Updating to the new SpokePoolVerifer addresses also required bumping the contracts-v2 package to the latest stable version. This on the other hand required some changes to the v2 deprecation. This PR therefore contains some of the changes of #980 and #982

### DIFF
--- a/scripts/generate-routes.ts
+++ b/scripts/generate-routes.ts
@@ -27,7 +27,7 @@ const enabledRoutes = {
       },
     ],
     spokePoolVerifier: {
-      address: "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D",
+      address: "0xbcfEb7d29BB1b40c6eA3A9e3d0822a79819d63F3",
       enabledChains: [
         CHAIN_IDs.MAINNET,
         CHAIN_IDs.OPTIMISM,
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -215,7 +215,7 @@ export async function sendAcrossDeposit(
   const depositor = await signer.getAddress();
   const inputTokenInfo = config.getTokenInfoByAddress(fromChain, tokenAddress);
   const outputTokenInfo = config.getTokenInfoBySymbol(
-    fromChain,
+    destinationChainId,
     inputTokenInfo.symbol
   );
   const inputAmount = amount;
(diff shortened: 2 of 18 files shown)


james-a-morris (approved):

mrice32 (approved):
LGTM!

dohaki:
NOTE: We still need to update the args when calling `SpokePoolVerifier.deposit()` in the FE because it now a passthrough to `depositV3`

src/utils/bridge.ts
```diff
@@ -242,7 +260,7 @@ export async function sendAcrossDeposit(
   const signerChainId = await signer.getChainId();
   if (signerChainId !== fromChain) {
     onNetworkMismatch?.({
-      signerAddress: await signer.getAddress(),
+      signerAddress: depositor,
```
james-a-morris, line 263: QQ: why was this changed? It makes sense, but was there any inconsistency that prompted it?

james-a-morris (approved):
LGTM

src/utils/bridge.ts
dohaki: Because I assign the value already here https://github.com/across-protocol/frontend-v2/pull/978/files#diff-5d22ab6b9a7272e68f42731b880e896d1671416cc379fcf34bf6106cc5b9658eR215.

scripts/generate-routes.ts
```diff
@@ -27,7 +27,7 @@ const enabledRoutes = {
       },
     ],
     spokePoolVerifier: {
-      address: "0x269727F088F16E1Aea52Cf5a97B1CD41DAA3f02D",
+      address: "0xbcfEb7d29BB1b40c6eA3A9e3d0822a79819d63F3",
```
amateima, line 30: Is this contract used in production?

amateima (approved):

scripts/generate-routes.ts
dohaki: Yea so the contracts are deployed to mainnet now. As soon as we merge this PR, they will also be used primarily.

dohaki: But I don't think you need to change/add any logic on the Scraper. The new `SpokePoolVerifier` contracts have fixes for the `depositor/msg.sender` issue we had in the old contracts.

nicholaspai: @amateima can the scraper handle fill deadlines that are < MAX_UINT256?

dohaki:
We will hold off on this PR for now because there are still open questions. E.g. Are all other parts of the system able to handle `fillDeadline` which is not `infinite`

PR #980, opened 2024-03-08
improve(V3): remove V2 event tracking

james-a-morris:
This PR removes V2 event tracking & V2 FillRelay simulation in the UI/API.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -36,6 +36,7 @@ import {
   defaultRelayerAddressOverride,
 } from "./_constants";
 import { PoolStateResult } from "./_types";
+import { bnUint32Max } from "@across-protocol/sdk-v2/dist/types/utils";
 
 type LoggingUtility = sdk.relayFeeCalculator.Logger;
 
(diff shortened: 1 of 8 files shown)


api/_utils.ts
```diff
@@ -520,18 +521,20 @@ export const getRelayerFeeDetails = async (
...
-        relayerFeePct: sdk.utils.bnOne,
-        realizedLpFeePct: sdk.utils.bnOne,
-        recipient: recipientAddress,
-        message: message ?? sdk.constants.EMPTY_MESSAGE,
         quoteTimestamp: sdk.utils.getCurrentTime(),
-        originToken,
-        destinationToken,
+        inputToken: originToken,
+        outputToken: destinationToken,
+        fillDeadline: bnUint32Max.toNumber(), // Defined as `INFINITE_FILL_DEADLINE` in SpokePool.sol
```
pxrl, line 534: Current time + 60 seconds is fine for this, since we only require that the fillDeadline is in the future relative to SpokePool time. That'd allow you to drop the `bnUint32Max` import.

api/_utils.ts
pxrl, line 1: Just fwiw, I think it'd be good to split this the fee details change out separately, since it has no blockers to going in right now. Bumping contracts seems like it needs more testing but it doesn't need to hold this change up.

api/_utils.ts
james-a-morris: I'd prefer to set these as closely as defined in the deposit() function of the spokePool

api/_utils.ts
james-a-morris: Tracking here: https://github.com/across-protocol/frontend-v2/pull/981

api/_utils.ts
```diff
@@ -520,18 +521,20 @@ export const getRelayerFeeDetails = async (
...
         depositId: sdk.utils.bnUint32Max.toNumber(),
         depositor: recipientAddress,
+        recipient: recipientAddress,
         destinationChainId,
         originChainId,
-        relayerFeePct: sdk.utils.bnOne,
-        realizedLpFeePct: sdk.utils.bnOne,
-        recipient: recipientAddress,
-        message: message ?? sdk.constants.EMPTY_MESSAGE,
         quoteTimestamp: sdk.utils.getCurrentTime(),
```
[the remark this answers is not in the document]
james-a-morris: Tracked here: https://github.com/across-protocol/frontend-v2/pull/981

dohaki (approved):
LGTM

PR #983, opened 2024-03-14
fix: remove fallback message in speed up

dohaki:
Related to ACX-1968. The actual fix was this PR though https://github.com/across-protocol/scraper-api/pull/362

The fallback `0x` masked the root cause. We therefore remove it in this PR


james-a-morris (commented):
Would this hurt to keep the fallback in-place? I.e. will keeping the 0x break anything?

james-a-morris:
> > Would this hurt to keep the fallback in-place? I.e. will keeping the 0x break anything?
> 
> 
> 
> @james-a-morris we had a discussion on that offline.

I was not privy to that - LGTM

PR #984, opened 2024-03-14
fix: only set parsed amount if true-ish

dohaki:
I noticed the some sporadic negative amount input on the bridge page. Not 100% sure what the root cause is but I suspect some race condition combined with the effect in `useAmountInput` hook.

### DIFF
--- a/src/views/Bridge/hooks/useAmountInput.ts
+++ b/src/views/Bridge/hooks/useAmountInput.ts
@@ -66,11 +66,13 @@ export function useAmountInput(selectedRoute: Route) {
   }, [prevFromTokenSymbol, selectedRoute.fromTokenSymbol, clearInput]);
 
   useEffect(() => {
-    try {
-      const parsed = utils.parseUnits(userAmountInput, token.decimals);
-      setParsedAmount(parsed);
-    } catch (error) {
-      setParsedAmount(undefined);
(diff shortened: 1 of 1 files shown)


james-a-morris:
The root cause I've been experiencing is the changing numeric input on the mouse/trackpad scroll. If you have an input with type number focused and attempt to scroll downward, the input goes down to a negative value.

I suspect the easiest solution would be to apply something like this to our input:
```
<input ... onWheel={(e) => e.target.blur()} />
```

james-a-morris (commented):
The root cause I've been experiencing is the changing numeric input on the mouse/trackpad scroll. If you have an input with type number focused and attempt to scroll downward, the input goes down to a negative value.

I suspect the easiest solution would be to apply something like this to our input:
```
<input ... onWheel={(e) => e.target.blur()} />
```

dohaki:
> The root cause I've been experiencing is the changing numeric input on the mouse/trackpad scroll. If you have an input with type number focused and attempt to scroll downward, the input goes down to a negative value.
> 
> I suspect the easiest solution would be to apply something like this to our input:
> 
> ```
> <input ... onWheel={(e) => e.target.blur()} />
> ```

Thanks a lot.. That seems to be the issue. I will close this PR and open a new one with your suggested fix

PR #987, opened 2024-03-18
feat: add expired status to deposits table

dohaki:


amateima (approved):

src/hooks/useDeposits.ts
```diff
@@ -38,6 +38,7 @@ export type Deposit = {
   initialRelayerFeePct?: string;
   suggestedRelayerFeePct?: string;
   fillTime?: number;
+  fillDeadline?: string;
```
james-a-morris, line 41: If `fillTime` is an optional number, then should `fillDeadline` also be an optional number?

james-a-morris (commented):
One comment

src/hooks/useDeposits.ts
dohaki: So `fillTime` is the duration in seconds, whereas `fillDeadline` is an ISO date string

PR #992, opened 2024-03-21
fix: enable usdt on linea

dohaki:


james-a-morris (approved):
LGTM from a code perspective

PR #993, opened 2024-03-22
improve(space optimization): optimize yarn.lock

james-a-morris:
In local testing, this brought the `node_modules` size from `4.4Gb` to `2.2Gb` creating a `50%` reduction in space storage.

Performed using `npx yarn-deduplicate yarn.lock`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@reach/dialog": "^0.16.2",
     "@safe-global/safe-apps-provider": "^0.18.0",
     "@safe-global/safe-apps-sdk": "^8.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "@reach/dialog": "^0.16.2",
     "@safe-global/safe-apps-provider": "^0.18.0",
     "@safe-global/safe-apps-sdk": "^8.1.0",
-    "@sentry/react": "^7.37.2",
+    "@sentry/react": "^7.108.0",
```
james-a-morris, line 22: A bump to sentry so that we're getting the most recent changes to detect errors

PR #994, opened 2024-03-26
fix: bump balancer sdk to fix pools 

dohaki:
Fixes ACX-1988

The beta version of the Balancer SDK finally allows setting a custom CG API key when initializing the SDK. It is still beta, but I don't think that should be a problem currently because our usage of the SDK is very limited.


api/_utils.ts
```diff
@@ -995,6 +995,10 @@ async function getBalancerPoolState(poolTokenAddress: string) {
       },
     },
     rpcUrl: getProvider(HUB_POOL_CHAIN_ID).connection.url,
+    coingecko: {
+      coingeckoApiKey:
+        REACT_APP_COINGECKO_PRO_API_KEY || "CG-ViHyrfvtLz2WSCJzm59TfGow",
```
james-a-morris, line 1000: I'm not sure we want to expose an API key in the codebase. cc @nicholaspai

james-a-morris (changes requested):

api/_utils.ts
dohaki: I would argue because the demo key is only used for the balancer sdk, it is fine to hardcode it as a fallback

dohaki: Removed for now until we rotate to our own demo key

PR #996, opened 2024-03-27
feat: more accurate fast fill time estimations

dohaki:
Closes ACX-1981

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -8,6 +8,7 @@ import {
 import { ERC20__factory } from "./typechain";
 import { tagAddress } from "./format";
 import { getProvider } from "./providers";
+import { getFastFillTimeByRoute } from "./fill-times";
 import { getConfig } from "utils";
 import getApiEndpoint from "./serverless-api";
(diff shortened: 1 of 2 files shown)


src/utils/bridge.ts
```diff
@@ -123,21 +124,32 @@ export const getConfirmationDepositTime = (
   };
 
   if (amount.lte(limits.maxDepositInstant)) {
-    return getTimeEstimateString(1, 5);
+    const fastFillTimeInSeconds = getFastFillTimeByRoute(fromChain, toChain);
+    const fastFillTimeInMinutes = Math.floor(fastFillTimeInSeconds / 60);
+    return {
+      formattedString:
+        fastFillTimeInSeconds < 60
+          ? `~${Math.floor(fastFillTimeInSeconds)} seconds`
```
nicholaspai, line 132: What if it's 1 second? Do we need to remove the s?

nicholaspai (approved):

src/utils/bridge.ts
dohaki: yea good point. handled now

james-a-morris (approved):

PR #999, opened 2024-04-04
refactor(api): statically resolve token details

dohaki:
This PR removes querying the HubPools `SetPoolRebalanceRoute` events is order to resolve the token details. Instead, we use a static approach to improve performance.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -201,62 +201,86 @@ export const resolveVercelEndpoint = () => {
   }
 };
 
-const _getTokenDetails = async (
-  provider: providers.Provider,
-  l1Token?: string,
-  l2Token?: string,
-  chainId?: string
+export const getTokenDetails = (
+  originTokenAddress: string,
+  destinationChainId: number,
+  originChainId?: number
 ) => {
-  const hubPool = HubPool__factory.connect(
-    ENABLED_ROUTES.hubPoolAddress,
-    provider
-  );
+  const token = _getTokenByAddress(originTokenAddress, originChainId);
 
-  // 2 queries: treating the token as the l1Token or treating the token as the L2 token.
-  const l2TokenFilter = hubPool.filters.SetPoolRebalanceRoute(
-    undefined,
-    l1Token,
-    l2Token
-  );
+  if (!token) {
+    throw new InputError(
+      originChainId
+        ? "Unsupported token on given origin chain"
+        : "Unsupported token address"
+    );
+  }
 
-  // Filter events by chainId.
-  const events = (await hubPool.queryFilter(l2TokenFilter, 0, "latest")).filter(
-    (event) => !chainId || event.args.destinationChainId.toString() === chainId
-  );
+  const possibleOriginChainIds = originChainId
+    ? [originChainId]
+    : _getChainIdsOfToken(originTokenAddress, token);
 
-  if (events.length === 0) throw new InputError("No whitelisted token found");
+  if (possibleOriginChainIds.length === 0) {
+    throw new InputError("Unsupported token address");
+  }
 
-  // Sorting from most recent to oldest.
-  events.sort((a, b) => {
-    if (b.blockNumber !== a.blockNumber) return b.blockNumber - a.blockNumber;
-    if (b.transactionIndex !== a.transactionIndex)
-      return b.transactionIndex - a.transactionIndex;
-    return b.logIndex - a.logIndex;
-  });
+  if (possibleOriginChainIds.length > 1) {
+    throw new InputError(
+      "More than one route is enabled for the provided inputs causing ambiguity. Please specify the originChainId."
+    );
+  }
 
-  return events.map((event) => ({
-    hubPool,
-    chainId: event.args.destinationChainId.toNumber(),
-    l1Token: event.args.l1Token,
-    l2Token: event.args.destinationToken,
-  }));
+  const resolvedOriginChainId = possibleOriginChainIds[0];
+  const originToken = token.addresses[resolvedOriginChainId];
+
+  if (!originToken) {
+    throw new InputError("Unsupported token address on given origin chain");
+  }
+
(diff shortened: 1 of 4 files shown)


api/_utils.ts
```diff
@@ -201,63 +201,87 @@ export const resolveVercelEndpoint = () => {
...
-  // Sorting from most recent to oldest.
-  events.sort((a, b) => {
-    if (b.blockNumber !== a.blockNumber) return b.blockNumber - a.blockNumber;
-    if (b.transactionIndex !== a.transactionIndex)
-      return b.transactionIndex - a.transactionIndex;
-    return b.logIndex - a.logIndex;
-  });
+  if (possibleOriginChainIds.length > 1) {
+    throw new InputError(
+      "More than one route is enabled for the provided inputs causing ambiguity. Please specify the originChainId."
```
pxrl, line 229: Nice!

api/_utils.ts
```diff
@@ -201,63 +201,87 @@ export const resolveVercelEndpoint = () => {
...
 };
 
-const _getTokenDetails = async (
-  provider: providers.Provider,
-  l1Token?: string,
-  l2Token?: string,
-  chainId?: string
+export const getTokenDetails = (
+  originTokenAddress: string,
+  destinationChainId: number,
```
pxrl, line 206: With v3 we migrated from {origin/destination}Token to {input/output}Token. Thoughts on doing the same here?

api/limits.ts
```diff
@@ -85,61 +79,39 @@ const handler = async (
...
+    const {
+      l1Token,
+      resolvedOriginChainId: computedOriginChainId,
+      destinationToken,
+      decimals,
+      symbol,
+    } = getTokenDetails(
       token,
-      originChainId
-    );
```
pxrl, line 101: ooc, from a high level, can we describe the strategy for resolving routes based on originChainId and origin/inputToken? Is it such that it's OK for the origin chain to be omitted if the origin/inputToken is unique across all known chains, and that otherwise we'll detect collisions and require the caller to be explicit about  originChainId?

dohaki: Ok yea will add a comment. But your description is correct. Some token addresses are not unique across chains, and therefore require a chain id to be set.

api/_utils.ts
dohaki: Addressed here https://github.com/across-protocol/frontend-v2/pull/999/commits/7594e1ac3aa32c45ab137aa36ce4dfb32dcc3bea

api/limits.ts
dohaki: Added here https://github.com/across-protocol/frontend-v2/pull/999/commits/7594e1ac3aa32c45ab137aa36ce4dfb32dcc3bea

pxrl: Perfect - very clear description.

nicholaspai (approved):
this is a great optimization, how much time do we save?

api/suggested-fees.ts
```diff
@@ -83,24 +83,26 @@ const handler = async (
...
+      !isRouteEnabled(computedOriginChainId, destinationChainId, token)
+    ) {
+      throw new InputError(`Route is not enabled.`);
     }
 
-    relayer ??= getDefaultRelayerAddress(
-      tokenInformation.symbol,
-      destinationChainId
-    );
+    const latestBlock = await provider.getBlock("latest");
```
pxrl, line 103: In the case of a message deposit there's potentially a blocking call between querying the block number here and the point where it's used. The latest block might change in the mean time. At the absolute margins it might cause us to resolve a different hubPool utilisation (i.e. if a deposit/withdrawal, finalisation or root bundle is executed). Do you think it's OK to relocate it to the latest point before it gets used?

api/suggested-fees.ts
```diff
@@ -178,27 +180,10 @@ const handler = async (
     const amount = ethers.BigNumber.from(amountInput);
 
     const blockFinder = new BlockFinder(provider.getBlock.bind(provider));
-    const [{ number: blockTag }, routeEnabled] = await Promise.all([
-      blockFinder.getBlockForTimestamp(parsedTimestamp),
-      isRouteEnabled(computedOriginChainId, destinationChainId, token),
-    ]);
+    const { number: blockTag } = await blockFinder.getBlockForTimestamp(
+      parsedTimestamp
+    );
```
pxrl, line 185: For subsequent work, this sticks out as a candidate for optimisation because it implies at least two RPC calls in addition to the existing `getBlock()` that we already do.

pxrl (approved):
I'm pretty optimistic about this (though there are further opportunities here too!). Also, we'll have to be very careful to synchronise updates with USDC/CCTP migration (and should update the CCTP migration plan to consider this change).

api/suggested-fees.ts
dohaki: Ok will open an issue

dohaki:
> this is a great optimization, how much time do we save?

It is hard to tell how much time it will be in the end because I could only test with a limited number of requests. I set up a Datadog dashboard widget so we can track it though.

api/suggested-fees.ts
dohaki: https://github.com/across-protocol/frontend-v2/pull/999/commits/18dfafb015dc3a3da74e5a0eb06bce94f468612f

api/limits.ts
```diff
@@ -164,7 +136,7 @@ const handler = async (
     ] = await Promise.all([
       getRelayerFeeDetails(
         l1Token,
-        ethers.BigNumber.from("10").pow(tokenDetails.decimals),
+        ethers.BigNumber.from("10").pow(decimals),
         computedOriginChainId,
         Number(destinationChainId),
```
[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/frontend-v2/pull/999/commits/18dfafb015dc3a3da74e5a0eb06bce94f468612f

PR #1000, opened 2024-04-04
refactor(api): optimize getblock call in fees handler

dohaki:
This PR implements the suggestion made by @pxrl [here](https://github.com/across-protocol/frontend-v2/pull/1000#pullrequestreview-1988379973) which removes the need to call `BlockFinder.getBlockForTimestamp` for the default case. This saves up ~ 3 sequential `getBlock` RPC calls which results in a 30-90ms speed-up. Also all `HubPool` contract calls are aggregated via mulitcall, which should decrease the opportunities for delayed RPC calls

Previously
vercel

Now
vercel after

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -135,8 +135,10 @@ export const relayerFeeCapitalCostConfig: {
   )
 );
 
+export const MAINNET_BLOCK_TIME_SECONDS = 12;
+
 // If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
-export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = 12 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
+export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = MAINNET_BLOCK_TIME_SECONDS * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
 
 // If `timestamp` is not passed into a suggested-fees query, then return the latest price rounded to
 // the nearest multiple of this value.
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -135,7 +135,8 @@ export const relayerFeeCapitalCostConfig: {
   )
 );
 
-// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet block minus this buffer.
+// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet block minus this buffer
+// rounded down to the nearest `QUOTE_BLOCK_PRECISION`th block interval. Can be overridden by env var `QUOTE_BLOCK_BUFFER`.
 export const DEFAULT_QUOTE_BLOCK_BUFFER = 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
 
 export const BLOCK_TAG_LAG = -1;
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -135,14 +135,8 @@ export const relayerFeeCapitalCostConfig: {
   )
 );
 
-export const MAINNET_BLOCK_TIME_SECONDS = 12;
-
-// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
-export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = MAINNET_BLOCK_TIME_SECONDS * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
-
-// If `timestamp` is not passed into a suggested-fees query, then return the latest price rounded to
-// the nearest multiple of this value.
-export const DEFAULT_QUOTE_TIMESTAMP_PRECISION = 5 * 60; // 5 minutes
+// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet block minus this buffer.
+export const DEFAULT_QUOTE_BLOCK_BUFFER = 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
 
 export const BLOCK_TAG_LAG = -1;
 
--- a/api/suggested-fees.ts
+++ b/api/suggested-fees.ts
@@ -4,9 +4,8 @@ import { ethers } from "ethers";
 import { type, assert, Infer, optional, string } from "superstruct";
 import {
   disabledL1Tokens,
-  DEFAULT_QUOTE_TIMESTAMP_BUFFER,
   DEFAULT_SIMULATED_RECIPIENT_ADDRESS,
-  MAINNET_BLOCK_TIME_SECONDS,
+  DEFAULT_QUOTE_BLOCK_BUFFER,
 } from "./_constants";
 import { TypedVercelRequest } from "./_types";
 import {
@@ -55,10 +54,7 @@ const handler = async (
     query,
   });
   try {
-    const { QUOTE_TIMESTAMP_BUFFER, QUOTE_TIMESTAMP_PRECISION } = process.env;
-    const quoteTimeBuffer = QUOTE_TIMESTAMP_BUFFER
-      ? Number(QUOTE_TIMESTAMP_BUFFER)
-      : DEFAULT_QUOTE_TIMESTAMP_BUFFER;
+    const { QUOTE_BLOCK_BUFFER, QUOTE_BLOCK_PRECISION } = process.env;
 
     const provider = getProvider(HUB_POOL_CHAIN_ID);
     const hubPool = getHubPool(provider);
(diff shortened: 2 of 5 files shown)


api/suggested-fees.ts
```diff
@@ -164,12 +165,19 @@ const handler = async (
...
-      const precision = Number(
-        QUOTE_TIMESTAMP_PRECISION ?? DEFAULT_QUOTE_TIMESTAMP_BUFFER
-      );
+      const precision = Number(QUOTE_TIMESTAMP_PRECISION ?? 1);
       parsedTimestamp =
         Math.floor((latestBlock.timestamp - quoteTimeBuffer) / precision) *
         precision;
+
+      const quoteTimeBufferInBlocks =
+        quoteTimeBuffer / MAINNET_BLOCK_TIME_SECONDS;
```
pxrl, line 174: Do we need any rounding here?

api/suggested-fees.ts
```diff
@@ -164,12 +165,19 @@ const handler = async (
       // We can use rounding here to increase the chance that a deposit's quote timestamp is re-used, thereby
       // allowing relayers hit the cache more often when fetching a block for a timestamp.
       // Divide by intended precision in seconds, round down to nearest integer, multiply by precision in seconds.
-      const precision = Number(
-        QUOTE_TIMESTAMP_PRECISION ?? DEFAULT_QUOTE_TIMESTAMP_BUFFER
-      );
+      const precision = Number(QUOTE_TIMESTAMP_PRECISION ?? 1);
```
pxrl, line 168: ooc, how is `QUOTE_TIMESTAMP_PRECISION` currently set in the env?

pxrl (commented):
Crazy idea...what if instead of rounding to the nearest 5 minutes, we rounded to the last 25th mainnet block? I think then we'd be able to skip the block search entirely.

We'd still need to resolve the timestamp for that block, but that can be done via multicall if we also bundle in the `hubPool.liquidityUtilizationCurrent()` and `hubPool.liquidityUtilizationPostRelay()` into that call (i.e. we can _also_ bundle a timestamp query).

So the procedure would (loosely) be:
- Query the latest block.
- Round the target block number down to the latest block number, modulo 5.
- Format a HubPool multicall query for:
  - hubPool.liquidityUtilizationCurrent()
  - hubPool.liquidityUtilizationPostRelay()
  - hubPool.getCurrentTime()

In that way we'd condense at least 3 RPC queries into a single one, and would reduce e2e latency because we'd no longer need to resolve the block for the desired timestamp.

(nb. this only works for cases where the quoteTimestamp is not supplied by the user - when the quoteTimestamp is supplied then we do need to resolve the right block).

api/suggested-fees.ts
dohaki: it isn't set actually

api/suggested-fees.ts
dohaki: yea would probably be better

dohaki:
> Crazy idea...what if instead of rounding to the nearest 5 minutes, we rounded to the last 25th mainnet block? I think then we'd be able to skip the block search entirely.
> 
> We'd still need to resolve the timestamp for that block, but that can be done via multicall if we also bundle in the `hubPool.liquidityUtilizationCurrent()` and `hubPool.liquidityUtilizationPostRelay()` into that call (i.e. we can _also_ bundle a timestamp query).
> 
> So the procedure would (loosely) be:
> 
> * Query the latest block.
> * Round the target block number down to the nearest  block number that is a multiple of 25.
> * Format a HubPool multicall query for:
>   
>   * hubPool.liquidityUtilizationCurrent()
>   * hubPool.liquidityUtilizationPostRelay()
>   * hubPool.getCurrentTime()
> 
> In that way we'd condense at least 3 RPC queries into a single one, and would reduce e2e latency because we'd no longer need to resolve the block for the desired timestamp.
> 
> (nb. this only works for cases where the quoteTimestamp is not supplied by the user - when the quoteTimestamp is supplied then we do need to resolve the right block).

Love that idea! Implemented here https://github.com/across-protocol/frontend-v2/pull/1000/commits/f3d6501edc35e8ab067697792495cd95de914ebd. This will speed up the `/suggested-fees` by 30-90ms 🚀 

Previously
image

Now
image

api/suggested-fees.ts
```diff
@@ -149,65 +145,78 @@ const handler = async (
...
-      // Divide by intended precision in seconds, round down to nearest integer, multiply by precision in seconds.
-      const precision = Number(
-        QUOTE_TIMESTAMP_PRECISION ?? DEFAULT_QUOTE_TIMESTAMP_BUFFER
-      );
-      parsedTimestamp =
-        Math.floor((latestBlock.timestamp - quoteTimeBuffer) / precision) *
+      // Divide by intended precision in blocks, round down to nearest integer, multiply by precision in blocks.
+      const precision = Number(QUOTE_BLOCK_PRECISION ?? quoteBlockBuffer);
+      quoteBlockNumber =
+        Math.floor((latestBlock.number - quoteBlockBuffer) / precision) *
```
nicholaspai, line 173: This is good call. Does the FE specify a quote timestamp most of the time?

nicholaspai (commented):
Nice optimization. How often does the FE specify a quote timestamp?

api/suggested-fees.ts
pxrl: Intuitively I expect it's usually the relayers who specify a quoteTimestamp when trying to estimate the LP fee for a deposit, and the aggregators leave it unspecified. I'm sure there are exceptions to that rule though - it'd be great to try to quantify it.

dohaki: > Intuitively I expect it's usually the relayers who specify a quoteTimestamp when trying to estimate the LP fee for a deposit, and the aggregators leave it unspecified. I'm sure there are exceptions to that rule though - it'd be great to try to quantify it.

@pxrl we can get some insight on that from Datadog.

api/suggested-fees.ts
```diff
@@ -149,65 +145,78 @@ const handler = async (
...
         precision;
```
pxrl, line 174: I had to test this locally to visualise it:

```
> precision = 20                
20                              
> quoteBlockBuffer = 5          
5
> for (let block = 500; block < 530; block += 1) {
... quoteBlockNumber = Math.floor((block - quoteBlockBuffer) / precision) * precision;
... console.log(`Block ${block}: quoteBlockNumber: ${quoteBlockNumber}`);
... }
Block 500: quoteBlockNumber: 480
Block 501: quoteBlockNumber: 480
Block 502: quoteBlockNumber: 480
Block 503: quoteBlockNumber: 480
Block 504: quoteBlockNumber: 480
Block 505: quoteBlockNumber: 500
Block 506: quoteBlockNumber: 500
Block 507: quoteBlockNumber: 500
Block 508: quoteBlockNumber: 500
Block 509: quoteBlockNumber: 500
Block 510: quoteBlockNumber: 500
Block 511: quoteBlockNumber: 500
Block 512: quoteBlockNumber: 500
Block 513: quoteBlockNumber: 500
Block 514: quoteBlockNumber: 500
Block 515: quoteBlockNumber: 500
Block 516: quoteBlockNumber: 500
Block 517: quoteBlockNumber: 500
Block 518: quoteBlockNumber: 500
Block 519: quoteBlockNumber: 500
Block 520: quoteBlockNumber: 500
Block 521: quoteBlockNumber: 500
Block 522: quoteBlockNumber: 500
Block 523: quoteBlockNumber: 500
Block 524: quoteBlockNumber: 500
Block 525: quoteBlockNumber: 520
Block 526: quoteBlockNumber: 520
Block 527: quoteBlockNumber: 520
Block 528: quoteBlockNumber: 520
Block 529: quoteBlockNumber: 520
```

I like the values of 5 and 20 respectively. The effect is that the computed quoteBlock/quoteTimestamp is at least 5 blocks behind HEAD and at most 24 blocks behind head, and it increments in steps of 20.

We can reduce precision from 20 to provide greater granularity in the quoteBlock/quoteTimestamp selected. This will mean that depositors pay a more "correct" fee, at the expense of the relayers and dataworker who have to do more work. Given that we currently accept a buffer of up to 60 minutes, I think ranging between 5 and 24 blocks is perfectly within tolerance. I also really like the flexibility that this gives.

api/_constants.ts
```diff
@@ -135,12 +135,8 @@ export const relayerFeeCapitalCostConfig: {
   )
 );
 
-// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet timestamp minus this buffer.
-export const DEFAULT_QUOTE_TIMESTAMP_BUFFER = 12 * 25; // ~25 blocks on mainnet (12s/block), ~= 5 minutes.
-
-// If `timestamp` is not passed into a suggested-fees query, then return the latest price rounded to
-// the nearest multiple of this value.
-export const DEFAULT_QUOTE_TIMESTAMP_PRECISION = 5 * 60; // 5 minutes
+// If `timestamp` is not passed into a suggested-fees query, then return the latest mainnet block minus this buffer.
```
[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/frontend-v2/pull/1000/commits/768905c75183efa80369988016f1dc8324b372d1

pxrl (commented):
I ++like this PR! I think the only thing I'd like to do before it gets merged is to spot test it for consistency against the existing prod implementation, just to make sure it's still producing the same results when the quoteTimestamp is supplied. I expect it should - also because there are minimal changes to that codepath. I'll aim to do that later tomorrow (but if someone else wants to do it sooner then feel free!).

@dohaki Are the env vars already configured in vercel?

api/suggested-fees.ts
```diff
@@ -149,65 +146,89 @@ const handler = async (
...
+      await Promise.all([
+        callViaMulticall3(provider, multiCalls, { blockTag: quoteBlockNumber }),
+        configStoreClient.getRateModel(
+          l1Token,
+          {
+            blockTag: quoteBlockNumber,
+          },
+          computedOriginChainId,
+          destinationChainId
+        ),
```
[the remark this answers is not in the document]
dohaki: I also thought we could do that initially. But if you closely this is a call to the `configStoreClient` and not the contract.

dohaki:
> @dohaki Are the env vars already configured in vercel?
@pxrl Yes, everything is set

api/suggested-fees.ts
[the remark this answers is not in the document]
james-a-morris: Ahhh shoot good point

pxrl:
@dohaki I'm late testing this but I think I found something unexpected. The LP fees always seem to match between prod and this test deployment, but the `timestamp` field in the response for this branch seems to be consistently offset by 5 seconds. i.e. the input query includes `&timestamp=x`, and the response returns `timestamp: (x-5)`.

I am testing with the script below.

```
#!/bin/sh                                                                                                                                                                                      
                                                                                                                                                                                               
prod=app.across.to                                                                                                                                                                             
test=app-frontend-v2-git-dong-ha-acx-2031-optimize-getblo-2c5f6b-uma.vercel.app                                                                                                                                                                                                                                                                                                               
dir="suggested-fees"                                                                                                                                                                           mkdir -p "${dir}"                                                                                                                                                                              
                                                                                                                                                                                               
base_timestamp=1713085000                                                                                                                                                                      
token=0x82aF49447D8a07e3bd95BD0d56f35241523fBab1                                                                                                                                               
srcChainId=42161                                                                                                                                                                               
dstChainId=1                                                                                                                                                                                   
amount=1000000000000000000                                                                                                                                                                     
                                                                                                                                                                                               
for offset in `seq 0 3600 36000`; do                                                                                                                                                             timestamp=`echo $(( $base_timestamp - $offset ))`                                                                                                                                              for host in $prod $test; do                                                                                                                                                                      query="https://${host}/api/suggested-fees?token=${token}&destinationChainId=${dstChainId}&originChainId=${srcChainId}&amount=${amount}&timestamp=${timestamp}"                                 raw_fees=`curl ${query}`                                                                                                                                                                       echo ${raw_fees} | python -c 'import json, sys; print(json.load(sys.stdin)["timestamp"])' > "${dir}/${timestamp}-${host}"                                                                    done                                                                                                                                                                                         done
```

pxrl:
> @dohaki I'm late testing this but I think I found something unexpected. The LP fees always seem to match between prod and this test deployment, but the `timestamp` field in the response for this branch seems to be consistently offset by 5 seconds. i.e. the input query includes `&timestamp=x`, and the response returns `timestamp: (x-5)`. Production currently always returns `&timestamp=x`.
> 
> I am testing with the script below.
> 
> ```
> #!/bin/sh                                                                                                                                                                                      
>                                                                                                                                                                                                
> prod=app.across.to                                                                                                                                                                             
> test=app-frontend-v2-git-dong-ha-acx-2031-optimize-getblo-2c5f6b-uma.vercel.app                                                                                                                                                                                                                                                                                                               
> dir="suggested-fees"                                                                                                                                                                           mkdir -p "${dir}"                                                                                                                                                                              
>                                                                                                                                                                                                
> base_timestamp=1713085000                                                                                                                                                                      
> token=0x82aF49447D8a07e3bd95BD0d56f35241523fBab1                                                                                                                                               
> srcChainId=42161                                                                                                                                                                               
> dstChainId=1                                                                                                                                                                                   
> amount=1000000000000000000                                                                                                                                                                     
>                                                                                                                                                                                                
> for offset in `seq 0 3600 36000`; do
>     timestamp=`echo $(( $base_timestamp - $offset ))`
>     for host in $prod $test; do
>         query="https://${host}/api/suggested-fees?token=${token}&destinationChainId=${dstChainId}&originChainId=${srcChainId}&amount=${amount}&timestamp=${timestamp}"
>         raw_fees=`curl ${query}`
>         echo ${raw_fees} | python -c 'import json, sys; print(json.load(sys.stdin)["timestamp"])' > "${dir}/${timestamp}-${host}"
>     done
> done
> ```

I think the issue is that when timestamp is supplied by the caller, we resolve it to a mainnet block number, and then return the timestamp of that block. So when the supplied timestamp is ahead of the block number, there can be a difference between the input `timestamp` and the output `timestamp`. Since they resolve to the same mainnet block number it should still resolve to the same utilisation and therefore the same LP fee.

api/suggested-fees.ts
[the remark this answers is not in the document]
pxrl: FWIW after this PR is in we could eventually add in the ConfigStore call to this multicall. It'd require some small refactoring, but the ConfigStoreClient class here is very simple and it's simply calling ConfigStore.l1TokenConfig(l1TokenAddress) at the relevant block number.

pxrl (approved):
OK from me, subject to a decision about whether to retain the existing quoteTimestamp behaviour.

dohaki:
> I think the root cause of this is that when timestamp is supplied by the caller, we resolve it to a mainnet block number, and then return the timestamp of that block. So when the supplied timestamp is ahead of the timestamp for the corresponding block number, there will be a difference between the input `timestamp` and the output `timestamp`. Since they resolve to the same mainnet block number it should still resolve to the same utilisation and therefore the same LP fee.
> 
> I don't know whether this is an issue or not; it's probably fine I suppose.

Thanks for testing this so thoroughly! Yea, so I could easily return the provided `timestamp` from the caller in the response instead

james-a-morris (approved):

dohaki:
> OK from me, subject to a decision about whether to retain the existing quoteTimestamp behaviour.

@pxrl I added a commit which recovers the previous quoteTimestamp behavior if provided by caller

api/suggested-fees.ts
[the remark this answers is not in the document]
dohaki: Created an issue here to track https://linear.app/uma/issue/ACX-2053/refactor-configstoreclientgetratemodel-so-that-it-can-be-aggregated

PR #1001, opened 2024-04-05
chore: bump sdk and constants to fixed WETH on Sepolia

dohaki:


james-a-morris (approved):
LGTM pending successful deployment on Vercel

PR #1002, opened 2024-04-05
improve(compliance): target disallow list in recipient

james-a-morris:
We should avoid selecting any disallowed wallets as recipients.


dohaki (approved):
Nice 👍

PR #1003, opened 2024-04-10
refactor: remove name collision of internal `formatUnits` helper

dohaki:
This PR renames our custom utility which formats a number AND truncates the fractional digits to `formatUnitsWithMaxFractions`. Previously it was named `formatUnits` which conflicts with a utility from `ethers`. This name collision has historically led to bugs where our internal utility did things we did not expect based on the name.


james-a-morris (approved):
I ran the UI and didn't see any breakages due to any component attempting to consume a numeric string with fractions. +1

PR #1008, opened 2024-04-15
fix(api): multicaller util

james-a-morris:
We should only prepend the overrides field to the multicaller client if it's defined.


james-a-morris:
Closed in favor of #1007

PR #1009, opened 2024-04-15
fix: make /pools more robust

dohaki:
This doesn't fix the issue with the Balancer pool timeouts but will probably reduce it on the UI. We still need to find a proper fix to somehow speed up the subgraph query

### DIFF
--- a/api/pools.ts
+++ b/api/pools.ts
@@ -48,7 +48,10 @@ const handler = async (
       message: "Response data",
       responseJson: responseData,
     });
-    response.setHeader("Cache-Control", "s-maxage=300");
+    response.setHeader(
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):
Would this double the cache look back?

dohaki:
> Would this double the cache look back?

In the worst case yes. But this shouldn't be a problem for this endpoint. From the docs
```
 A response containing:

     Cache-Control: max-age=600, stale-while-revalidate=30

   indicates that it is fresh for 600 seconds, and it may continue to be
   served stale for up to an additional 30 seconds while an asynchronous
   validation is attempted.
```

PR #1010, opened 2024-04-16
fix(fees-api): pass in hub chain when deriving destination token

dohaki:
`GET /suggested-fees` was failing when `&message=` was specified on testnet


james-a-morris (approved):
+1

PR #1012, opened 2024-04-16
feat: bridge UI updates with new design and UX improvements

dohaki:
Closes ACX-2015, ACX-2046 and ACX-2067


src/hooks/useBalance.ts
```diff
@@ -1,55 +1,43 @@
...
   ConfigClient,
+  getChainInfo,
 } from "utils";
 import { BigNumber } from "ethers";
 
 export function useNativeBalance(
   tokenSymbol?: string,
   chainId?: ChainId,
-  account?: string,
-  blockNumber?: number
```
dohaki, line 19: I removed the dependency to `blockNumber` for a bunch of queries. Some reasons:
1. We never use a specific block number for this query and instead use the latest block
2. Because the query key is based on the block number, every time a new block gets mined a new query is executed. This leads to some weird "flickering" behavior on the FE when we consume this query. The new query has an initial value of `undefined` which causes this issue.
3. Removing this dependency gives us more control on when/how to refetch new data

src/hooks/useAllowance.ts
dohaki, line 1: Deleted because not used anywhere

src/hooks/useBalance.ts
james-a-morris: This is a fair call - I think most of the FE is handled in real-time at the latest block, so this makes sense.

src/hooks/useBalance.ts
```diff
@@ -1,55 +1,43 @@
...
         tokenSymbolToQuery,
         accountToQuery,
-        blockNumberToQuery,
       ];
   const { data: balance, ...delegated } = useQuery(
     queryKey,
     async () => {
       if (!chainIdToQuery) return BigNumber.from(0);
-      return getNativeBalance(chainId, accountToQuery!, blockNumberToQuery);
+      return getNativeBalance(chainId, accountToQuery!);
```
james-a-morris, line 36: We should probably be referencing the key from inside this anonymous function. We can pass the context key as a parameter in the async( ) above ^

src/hooks/useBalance.ts
```diff
@@ -64,41 +52,21 @@ const QueryBalanceBySymbol =
...
-      blockNumberToQuery,
-      config,
-    } = params;
-    if (
-      !chainIdToQuery ||
-      !tokenSymbolToQuery ||
-      !blockNumberToQuery ||
-      !accountToQuery
-    )
+    const { accountToQuery, chainIdToQuery, tokenSymbolToQuery, config } =
```
james-a-morris, line 58: Can this function be hoisted into the file scope? It doesn't look like we're using any variables in the parent scope.

src/hooks/useBalance.ts
```diff
@@ -112,30 +80,20 @@ const QueryBalanceBySymbol =
...
   const chainIdToQuery = chainId;
   const tokenSymbolToQuery = tokenSymbol;
   const accountToQuery = account ?? connectedAccount;
-  const { block: latestBlock } = useBlock(chainId);
-  const blockNumberToQuery = blockNumber ?? latestBlock?.number;
-  const enabledQuery =
-    !!chainIdToQuery && !!accountToQuery && !!blockNumberToQuery;
+  const enabledQuery = !!chainIdToQuery && !!accountToQuery;
   const queryKey = enabledQuery
-    ? balanceQueryKey(
```
james-a-morris, line 127: Should we put this in https://github.com/across-protocol/frontend-v3/blob/master/src/utils/query-keys.ts

src/hooks/useBalance.ts
```diff
@@ -144,13 +102,11 @@ export function useBalanceBySymbol(
...
       chainIdToQuery,
       tokenSymbolToQuery,
-      blockNumberToQuery,
       accountToQuery,
     }),
     {
       enabled: enabledQuery,
-      // We already re-fetch when the block number changes, so we don't need to re-fetch.
-      staleTime: Infinity,
+      refetchInterval: getChainInfo(chainIdToQuery || 1).pollingInterval,
```
james-a-morris, line 109: Should the async function above be referencing the query key that is provided when the handler function is called? 

```
async (key) => QueryBalanceBySymbol({...})
```

src/hooks/useBalance.ts
```diff
@@ -164,64 +120,43 @@ export function useBalanceBySymbol(
...
   // we use useQueries instead of useQuery so we can share cache values with the singular balance query
   const queries = tokenSymbols.map((tokenSymbolToQuery) => {
     const queryKey = enabled
-      ? balanceQueryKey(
-          accountToQuery,
-          blockNumberToQuery,
-          chainIdToQuery,
-          tokenSymbolToQuery
-        )
+      ? balanceQueryKey(accountToQuery, chainIdToQuery, tokenSymbolToQuery)
```
james-a-morris, line 142: Would it make sense to place this here https://github.com/across-protocol/frontend-v3/blob/master/src/utils/query-keys.ts?

src/hooks/useBridgeFees.ts
```diff
@@ -18,34 +17,25 @@ export function useBridgeFees(
...
+    !!toChainId && !!fromChainId && !!tokenSymbol && amount.gt(0);
   const queryKey = enabledQuery
-    ? bridgeFeesQueryKey(
-        tokenSymbol,
-        amount,
-        fromChainId,
-        toChainId,
-        block.number
-      )
+    ? bridgeFeesQueryKey(tokenSymbol, amount, fromChainId, toChainId)
```
james-a-morris, line 23: Similar to above

src/hooks/useBridgeFees.ts
```diff
@@ -18,34 +17,25 @@ export function useBridgeFees(
...
         toChainId: toChainId!,
         fromChainId: fromChainId!,
         recipientAddress,
       });
     },
     {
       enabled: enabledQuery,
-      // We already re-fetch when the block number changes, so we don't need to re-fetch.
-      staleTime: Infinity,
+      refetchInterval: 5000,
```
james-a-morris, line 38: Similar to above for the `async (key) => `

src/hooks/useEns.ts
```diff
@@ -0,0 +1,20 @@
+import { useQuery } from "react-query";
+
+import { ChainId, getProvider } from "utils";
+
+export function useEnsQuery(address: string) {
+  return useQuery(
+    ["ens", address],
+    async () => {
```
james-a-morris, line 8: Similar to above. If the `address` changes mid-flight we could be storing incorrect information within `["ens", address]`

james-a-morris (commented):
A few initial comments - mostly on the react query calls. I do think we should be using the [query function context](https://tanstack.com/query/latest/docs/framework/react/guides/query-functions#queryfunctioncontext) to ensure we're always caching what we expect

src/hooks/useBalance.ts
dohaki: It is already placed there

src/hooks/useBridgeFees.ts
dohaki: It is already placed there

src/hooks/useBalance.ts
dohaki: It is placed there already

src/hooks/useBalance.ts
dohaki: Hmm not sure what you mean. This functions is already in the file scope

dohaki: Ah now I understand. Yea no idea why we have it this way 🤔 Will hoist it

src/hooks/useEns.ts
dohaki: https://github.com/across-protocol/frontend-v3/pull/1012/commits/a243b5c73bf0f7cbf1c0ee075d1b434789f851ed

james-a-morris (approved):
Only other comments I may have are QA related but we can hold off for the more formal review

PR #1015, opened 2024-04-23
fix: use full display width for tx table

dohaki:
Closes ACX-1945

### DIFF
--- a/src/views/Transactions/Transactions.tsx
+++ b/src/views/Transactions/Transactions.tsx
@@ -19,7 +19,7 @@ export function Transactions() {
   );
 
   return (
-    <LayoutV2 maxWidth={1140}>
+    <LayoutV2 maxWidth={activeTab === "personal" ? 1476 : 1132}>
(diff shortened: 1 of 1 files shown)


src/views/Transactions/Transactions.tsx
```diff
@@ -19,7 +19,7 @@ export function Transactions() {
   );
 
   return (
-    <LayoutV2 maxWidth={1140}>
+    <LayoutV2 maxWidth={activeTab === "personal" ? 1476 : 1132}>
```
james-a-morris, line 22: Could we do this for both?

dohaki: We could but looks a bit weird IMO
image

james-a-morris: Ah shoot, you're right - the table isn't responsive for larger displays 😞

james-a-morris (approved):

PR #1016, opened 2024-04-23
docs: add openapi yaml file for api docs

dohaki:
This PR adds the OpenAPI yaml file for our API docs. We also updated the docs according to the changes made here https://github.com/across-protocol/frontend-v3/pull/1013. I opted to open a separate PR to not further increase the diff.

### DIFF
--- a/api-docs-openapi.yaml
+++ b/api-docs-openapi.yaml
@@ -5,6 +5,7 @@ info:
   version: 0.1.0
 servers:
   - url: https://app.across.to/api
+  - url: https://testnet.across.to/api
 paths:
   /suggested-fees:
     get:
@@ -83,21 +84,43 @@ paths:
           required: true
           description: >
             Chain ID where the specified `token` or `inputToken` exists.
-
-
-            _Example:_ 10
           schema:
             type: integer
+            enum: [
+                1,
+                10,
+                137,
+                324,
+                8453,
+                42161,
+                59144,
+                # testnets
+                84532,
+                421614,
+                11155420,
+                11155111,
+              ]
         - name: destinationChainId
           in: query
           required: true
           description: >
             The desired destination chain ID of the bridge transfer.
-
-
-            _Example:_ 10
           schema:
             type: integer
+            enum: [
+                1,
+                10,
+                137,
+                324,
+                8453,
+                42161,
+                59144,
+                # testnets
+                84532,
+                421614,
+                11155420,
+                11155111,
+              ]
         - name: amount
           in: query
           required: true
@@ -114,6 +137,7 @@ paths:
             _Example:_ 100000000000
           schema:
             type: integer
+            minimum: 1
         - name: recipient
           in: query
           required: false
@@ -122,7 +146,7 @@ paths:
             EOA and message is defined, then the API will throw a 4xx error.
 
 
-            _Example_: 0xc186fa914353c44b2e33ebe05f21846f1048beda
+            _Example_: 0xc186fA914353c44b2E33eBE05f21846F1048bEda
           schema:
             type: string
         - name: message
@@ -152,7 +176,7 @@ paths:
             balance to make the fill.
 
 
-            _Example:_ 0x428ab2ba90eba0a4be7af34c9ac451ab061ac010
+            _Example:_ 0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010
           schema:
             type: string
         - name: timestamp
@@ -288,25 +312,44 @@ paths:
           in: query
           required: true
           description: >
-            Used to specify the chain where the specified token address exists.
-            Note: this is only needed to disambiguate when there are matching
-            addresses on different chains. In that case, an arbitrary one will
-            be chosen, so it is recommended that this is always provided.
-
-
-            _Example:_ 10
+            Chain ID where the specified `token` or `inputToken` exists.
           schema:
             type: integer
+            enum: [
+                1,
+                10,
+                137,
+                324,
+                8453,
+                42161,
+                59144,
+                # testnets
+                84532,
+                421614,
+                11155420,
+                11155111,
+              ]
         - name: destinationChainId
           in: query
           required: true
           description: >
             The intended destination chain ID of the bridge transfer.
-
-
-            _Example:_ 42161
           schema:
             type: integer
+            enum: [
+                1,
+                10,
+                137,
+                324,
+                8453,
+                42161,
+                59144,
+                # testnets
+                84532,
+                421614,
+                11155420,
+                11155111,
+              ]
       responses:
         "200":
           description: Transfer limits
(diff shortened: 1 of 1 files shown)


api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+  version: 0.1.0
+servers:
+  - url: https://app.across.to/api
+paths:
+  /suggested-fees:
+    get:
+      summary: Retrieve suggested fee quote for a deposit.
+      description: >
+        Returns suggested fees based on the `token` OR
+        `inputToken`+`outputToken`,  `destinationChainId`, and `amount`.
```
nicholaspai, line 14: isn't `originChainId` required too?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+            Note that this amount is in the native decimals of the token. So,
+            for WETH, this would be the amount of human-readable WETH multiplied
+            by 1e18. For USDC, you would multiply the number of human-readable
+            USDC by 1e6.
+
+
+            _Example:_ 100000000000
+          schema:
+            type: integer
+        - name: originChainId
```
nicholaspai, line 112: I thought we wanted to make this required at some point?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+            higher precedence.**
+
+
+            _Example:_ 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2
+        - name: inputToken
+          in: query
+          schema:
+            type: string
+          description: >
+            Address of token to bridge on the _origin_ chain. Must be used
```
nicholaspai, line 261: Is `originChain` implied/should we require it?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+      responses:
+        "200":
+          description: Transfer limits
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TransferLimits"
+              example:
+                {
+                  "minDeposit": "7799819",
```
nicholaspai, line 333: I think it might be useful to define what these limits translate to in terms of ETA of fill

nicholaspai: ah nvm i see you do this

api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+            The maximum deposit size in the tokens' units. Note: The formatting
+            of this number is the same as minDeposit.
+          example: 22287428516241
+        maxDepositInstant:
+          type: string
+          description: >
+            The max deposit size that can be relayed "instantly" on the
+            destination chain. Instantly means that there is relayer capital
+            readily available and that a relayer is expected to relay within 1-4
+            minutes of the deposit.
```
nicholaspai, line 551: this can be less than 1 minute right?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,585 @@
...
+          description: >
+            The max deposit size that can be relayed with a "short delay" on the
+            destination chain. This means that there is relayer capital
+            available on mainnet and that a relayer will immediately begin
+            moving that capital over the canonical bridge to relay the deposit.
+            Depending on the chain, the time for this can vary. Polygon is the
+            worst case where it can take between 20 and 35 minutes for the
+            relayer to receive the funds and relay. Arbitrum is much faster,
+            with a range between 5 and 15 minutes. Note: if the transfer size is
+            greater than this, the estimate should be between 2-4 hours for a
```
nicholaspai, line 564: This will change with CEX rebalancing FYI

api-docs-openapi.yaml
dohaki: We don't enforce it yet but I can make it required in the docs for now 🤔

api-docs-openapi.yaml
dohaki: yes will change

api-docs-openapi.yaml
dohaki: It is not currently. But I can state it as `required` in the docs before we enforce it eventually.

api-docs-openapi.yaml
dohaki: I think it makes sense yea. At least it doesn't hurt to state it as required in the docs.

api-docs-openapi.yaml
```diff
@@ -0,0 +1,571 @@
...
+
+
+            **Also note that this parameter will eventually replace the
+            parameter `token` above. For now, you need to either specify `token`
+            OR `inputToken` +`outputToken` with the parameter `token` having a
+            higher precedence.**
+
+
+            _Example:_ 0x4200000000000000000000000000000000000006
+        - name: originChainId
```
james-a-morris, line 81: Could we list a "currently supported" set of chain ids that would produce valid results?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,571 @@
...
+
+            Note that this amount is in the native decimals of the token. So,
+            for WETH, this would be the amount of human-readable WETH multiplied
+            by 1e18. For USDC, you would multiply the number of human-readable
+            USDC by 1e6.
+
+
+            _Example:_ 100000000000
+          schema:
+            type: integer
```
james-a-morris, line 116: Is there a way to specify that this number should be positive?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,571 @@
...
+        - name: recipient
+          in: query
+          required: false
+          description: >
+            Recipient of the deposit. Can be an EOA or a contract. If this is an
+            EOA and message is defined, then the API will throw a 4xx error.
+
+
+            _Example_: 0xc186fa914353c44b2e33ebe05f21846f1048beda
```
james-a-morris, line 125: For consistency - thoughts on making these addresses checksummed?

api-docs-openapi.yaml
```diff
@@ -0,0 +1,571 @@
...
+          deprecated: true
+          schema:
+            type: string
+          description: >
+            Address of token to bridge. For native ETH, use the wrapped address,
+            i.e. WETH.
+
+
+            Note that the address provided must exist on the specified
+            `originChainId`.
```
james-a-morris, line 28: We may want to include a small comment here about route collisions.

A possible example URL could be: https://app.across.to/api/limits?token=0xba100000625a3754423978a60c9317c58a424e3D&destinationChainId=1

dohaki: We will make the parameter `originChainId` required in the docs at least. And enforce it soon in the API

api-docs-openapi.yaml
dohaki: https://github.com/across-protocol/frontend-v3/pull/1016/commits/d78cfc9109534bfb1e5a77ec0baa686a344fa9e1

api-docs-openapi.yaml
dohaki: https://github.com/across-protocol/frontend-v3/pull/1016/commits/d78cfc9109534bfb1e5a77ec0baa686a344fa9e1

api-docs-openapi.yaml
dohaki: https://github.com/across-protocol/frontend-v3/pull/1016/commits/d78cfc9109534bfb1e5a77ec0baa686a344fa9e1

james-a-morris (approved):

PR #1004, opened 2024-04-11
feat: add `AmountInput` component with new design

dohaki:
This adds a new component `AmountInput` that implements the new design https://www.figma.com/file/q0GREWOG7XTM0mM0brAxGn/%E2%9D%8E-Bridge-Dapp?type=design&node-id=8861-57238&mode=design&t=nDncCMIbuqP4bult-4.

I also refactored the `Pool` and `Staking` views to use this new component to make all amount inputs consistent.

Note that this PR DOES NOT integrate the new component into the `Bridge` view yet. I chose this approach because I will open a subsequent PR with more changes on the bridge page.

PR #1017, opened 2024-04-25
feat(cctp): bridge ui design updates + input/output token support

dohaki:
Attempt to update production with cumulated changes from:
- #1004 
- #1012


api/_utils.ts
```diff
@@ -397,13 +397,13 @@ export const getRouteDetails = (
 
 const _getTokenByAddress = (tokenAddress: string, chainId?: number) => {
   const [, token] =
-    Object.entries(TOKEN_SYMBOLS_MAP).find(([_symbol, { addresses }]) =>
-      chainId
+    Object.entries(TOKEN_SYMBOLS_MAP).find(([_key, { addresses }]) => {
+      return chainId
```
james-a-morris, line 401: Do we need the return?

src/hooks/useBridgeFees.ts
```diff
@@ -18,34 +17,38 @@ export function useBridgeFees(
...
+    ({ queryKey }) => {
+      const [
+        ,
+        tokenSymbolToQuery,
+        amountToQuery,
+        fromChainIdToQuery,
+        toChainIdToQuery,
+      ] = queryKey;
+
+      if (!toChainIdToQuery || !fromChainIdToQuery || !tokenSymbolToQuery) {
```
james-a-morris, line 37: Thoughts on switching this to use `isDefined( )`?

src/views/Bridge/components/RecipientRow.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onClickChangeToAddress: VoidHandler;
+};
+
+export function RecipientRow(props: Props) {
+  const { data } = useEnsQuery(props.recipient.address);
+
+  return (
+    <Wrapper>
+      <Text color="grey-400">Address:</Text>
+      {data?.avatar && <img src={data.avatar} alt="avatar" />}
```
james-a-morris, line 22: Could we use a next Image tag here?

james-a-morris (approved):
a few nits - otherwise LGTM

src/hooks/useBridgeFees.ts
dohaki: Why do you prefer `isDefined` here`?

src/views/Bridge/components/RecipientRow.tsx
dohaki: We never used `next/image` in the dApp so far... If we introduce it, I would prefer to do so in a separate PR

api/_utils.ts
dohaki: Reverted

PR #1019, opened 2024-04-25
feat(api): `GET /swap-quote` endpoint

dohaki:
Closes ACX-2016

We will use `epic/cctp` for testing, that's why the base branch for this PR is set to that.

Example request:
https://app-frontend-v3-git-swap-quote-endpoint-uma.vercel.app/api/swap-quote?depositor=0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D&acrossInputToken=0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174&swapTokenAmount=100000000&originChainId=137&destinationChainId=10&acrossOutputToken=0x7F5c764cBc14f9669B88837ca1490cCa17c31607&swapToken=0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359

Note that I enabled the following swap route for testing purposes:
- `Native USDC (Polygon)` swap `Bridged USDC.e (Polygon)` bridge `Bridged USDC.e (OP)`

I am aware that this is not a route we want to support when we launch, but the closest route I could think of that fits our use case.

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -11,4 +11,11 @@ jobs:
       - name: 'Checkout Repository'
         uses: actions/checkout@v3
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@v3.0.4
+        uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: high
+          # Comma-separated list of GHSA IDs to allow.
+          # We allow @openzeppelin/contracts@3.4.1-solc-0.7-2 critical vulnerabilities
+          # because they are not exploitable in our usage of the package.
+          allow-ghsas: >
+            "GHSA-fg47-3c2x-m2wr,GHSA-88g8-f5mf-f5rj"
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -17,5 +17,4 @@ jobs:
           # Comma-separated list of GHSA IDs to allow.
           # We allow @openzeppelin/contracts@3.4.1-solc-0.7-2 critical vulnerabilities
           # because they are not exploitable in our usage of the package.
-          allow-ghsas: >
-            "GHSA-fg47-3c2x-m2wr,GHSA-88g8-f5mf-f5rj"
+          allow-ghsas: GHSA-fg47-3c2x-m2wr, GHSA-88g8-f5mf-f5rj
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -1,4 +1,4 @@
-name: 'Dependency Review'
+name: "Dependency Review"
 on: [pull_request]
 
 permissions:
@@ -8,9 +8,9 @@ jobs:
   dependency-review:
     runs-on: ubuntu-latest
     steps:
-      - name: 'Checkout Repository'
+      - name: "Checkout Repository"
         uses: actions/checkout@v3
-      - name: 'Dependency Review'
+      - name: "Dependency Review"
         uses: actions/dependency-review-action@v4
         with:
           fail-on-severity: high
--- a/api/_dexes/1inch.ts
+++ b/api/_dexes/1inch.ts
@@ -1,0 +1,52 @@
+import axios from "axios";
+
+import { AcrossSwap, SwapQuoteAndCalldata } from "./types";
+import { getSwapAndBridgeAddress } from "./utils";
+
+const config = {
+  slippage: 1, // Maximum acceptable slippage percentage for the swap (e.g., 1 for 1%)
+};
+
+export async function get1inchQuoteAndCalldata(
+  swap: AcrossSwap
+): Promise<SwapQuoteAndCalldata> {
+  try {
+    const swapAndBridgeAddress = getSwapAndBridgeAddress(
+      "1inch",
+      swap.swapToken.chainId
+    );
+    const apiBaseUrl = `https://api.1inch.dev/swap/v6.0/${swap.swapToken.chainId}`;
+    const apiHeaders = {
+      Authorization: `Bearer ${process.env.ONEINCH_API_KEY}`,
+      accept: "application/json",
+    };
+
+    const swapParams = {
+      src: swap.swapToken.address,
+      dst: swap.acrossInputToken.address,
+      amount: swap.swapTokenAmount,
+      from: swapAndBridgeAddress,
+      slippage: config.slippage,
+      disableEstimate: true,
+      allowPartialFill: false,
+      receiver: swapAndBridgeAddress,
+    };
+
+    const response = await axios.get(`${apiBaseUrl}/swap`, {
+      headers: apiHeaders,
+      params: swapParams,
+    });
+
+    return {
+      minExpectedInputTokenAmount:
+        response.data.toAmount || response.data.dstAmount,
+      routerCalldata: response.data.tx.data,
+      value: response.data.tx.value,
+      swapAndBridgeAddress,
+      dex: "1inch",
+    };
+  } catch (e) {
+    console.error(e);
+    throw e;
+  }
+}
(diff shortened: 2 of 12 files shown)


api/_dexes/uniswap.ts
dohaki, line 1: Even though we decided to go with aggregators for now, I was thinking of keeping this handler implementation as a fallback? 🤔 But we could also just remove it

api/swap-quote.ts
```diff
@@ -0,0 +1,153 @@
...
+import { get1inchQuoteAndCalldata } from "./_dexes/1inch";
+
+const SwapQuoteQueryParamsSchema = type({
+  swapToken: validAddress(),
+  acrossInputToken: validAddress(),
+  acrossOutputToken: validAddress(),
+  swapTokenAmount: string(),
+  originChainId: positiveIntStr(),
+  destinationChainId: positiveIntStr(),
+  swapSlippage: optional(positiveFloatStr(50)), // max. 50% slippage
```
dohaki, line 27: max. 50% slippage is based on limitations of the 1inch API

api/_dexes/uniswap.ts
dohaki, line 1: Even though we decided to go with aggregators for now, I was thinking of keeping this handler implementation as a fallback? 🤔

api/_dexes/1inch.ts
```diff
@@ -0,0 +1,44 @@
...
+    dst: swap.acrossInputToken.address,
+    amount: swap.swapTokenAmount,
+    from: swapAndBridgeAddress,
+    slippage: swap.slippage,
+    disableEstimate: true,
+    allowPartialFill: false,
+    receiver: swapAndBridgeAddress,
+  };
+
+  const response = await axios.get(`${apiBaseUrl}/swap`, {
```
james-a-morris, line 30: Could we type this call?

api/_dexes/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  swapTokenAmount: string;
+  slippage: number;
+};
+
+export type SwapQuoteAndCalldata = {
+  minExpectedInputTokenAmount: string;
+  routerCalldata: string;
+  value: string;
+  swapAndBridgeAddress: string;
+  dex: string;
```
james-a-morris, line 26: Could this be an enum or a restricted string to only the dexes we support currently?

api/_dexes/uniswap.ts
```diff
@@ -0,0 +1,185 @@
...
+import JSBI from "jsbi";
+import IUniswapV3PoolABI from "@uniswap/v3-core/artifacts/contracts/interfaces/IUniswapV3Pool.sol/IUniswapV3Pool.json";
+
+import { callViaMulticall3, getProvider } from "../_utils";
+import { AcrossSwap, SwapQuoteAndCalldata } from "./types";
+import { getSwapAndBridgeAddress } from "./utils";
+
+const POOL_FACTORY_CONTRACT_ADDRESS =
+  "0x1F98431c8aD98523631AE4a59f267346ea31F984";
+const QUOTER_CONTRACT_ADDRESS = "0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6";
```
james-a-morris, line 28: Are these the same addresses on testnets? If not, should we move these to a utility function and switch on the hub chain id?

api/_dexes/uniswap.ts
```diff
@@ -0,0 +1,185 @@
...
+  );
+  const tokenB = new Token(
+    swap.acrossInputToken.chainId,
+    swap.acrossInputToken.address,
+    swap.acrossInputToken.decimals
+  );
+  const pool = new Pool(
+    tokenA,
+    tokenB,
+    FeeAmount.LOW,
```
james-a-morris, line 52: Will this always be `LOW`?

api/_dexes/uniswap.ts
```diff
@@ -0,0 +1,185 @@
...
+  });
+
+  const options: SwapOptions = {
+    slippageTolerance: new Percent(
+      // max. slippage decimals is 2
+      Number(swap.slippage.toFixed(2)) * 100,
+      10_000
+    ),
+    // 20 minutes from the current Unix time
+    deadline: Math.floor(Date.now() / 1000) + 60 * 20,
```
james-a-morris, line 76: We should get an SDK util called `getCurrentTime()` that we can add 20 minutes to

api/_dexes/uniswap.ts
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+  return {
+    token0: token0 as unknown as string,
+    token1: token1 as unknown as string,
+    fee: fee as unknown as number,
+    tickSpacing: tickSpacing as unknown as number,
+    liquidity: liquidity as unknown as ethers.BigNumber,
+    sqrtPriceX96: slot0[0] as unknown as ethers.BigNumber,
+    tick: slot0[1] as unknown as number,
```
james-a-morris, line 183: Should we do some validation here prior to the unknown casts?

api/_utils.ts
```diff
@@ -297,7 +297,7 @@ export const getRouteDetails = (
   originChainId?: number,
   outputTokenAddress?: string
 ) => {
-  const inputToken = _getTokenByAddress(inputTokenAddress, originChainId);
+  const inputToken = getTokenByAddress(inputTokenAddress, originChainId);
```
james-a-morris, line 300: This may fail in the USDC case

api/swap-quote.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (
+      !ENABLED_ROUTES.swapRoutes.find((route) => {
+        return (
+          route.fromChain === originChainId &&
+          route.toChain === destinationChainId &&
+          route.fromTokenSymbol === acrossInputToken.symbol &&
+          route.toTokenSymbol === acrossOutputToken.symbol &&
+          route.swapTokenAddress.toLowerCase() ===
+            swapTokenAddress.toLowerCase()
+        );
```
james-a-morris, line 94: Could we encapsulate this in a utility function?

api/swap-quote.ts
james-a-morris: Is this just to be able to get a quote?

.github/workflows/dependency-review.yaml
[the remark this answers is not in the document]
dohaki: yes we need these because the vulnerabilities got introduced via the uniswap sdks

api/_dexes/uniswap.ts
dohaki: I don't think we need a validation here. The reason why I am casting to `unknowns` is due to the typing of `callViaMulicall3` which returns the generic `Result` type. But I can push a fix to make this a bit cleaner

api/swap-quote.ts
dohaki: Yea, the user will be able to edit this value also

api/_dexes/types.ts
dohaki: https://github.com/across-protocol/frontend-v3/pull/1019/commits/844624a493f9cc74b4cc7ffb2603864133b74b0c

api/_dexes/uniswap.ts
dohaki: https://github.com/across-protocol/frontend-v3/pull/1019/commits/844624a493f9cc74b4cc7ffb2603864133b74b0c

api/_dexes/uniswap.ts
dohaki: https://github.com/across-protocol/frontend-v3/pull/1019/commits/844624a493f9cc74b4cc7ffb2603864133b74b0c

api/_utils.ts
dohaki: Why would it fail in the USDC case? 🤔

api/swap-quote.ts
dohaki: https://github.com/across-protocol/frontend-v3/pull/1019/commits/844624a493f9cc74b4cc7ffb2603864133b74b0c

PR #1027, opened 2024-05-08
feat: use more detailed est. fill time

dohaki:
Use more detailed est. fill time on bridge page


src/data/fill-times-preset.json
```diff
@@ -0,0 +1,773 @@
+{
+  "result": [
+    {
+      "data": [
+        {
+          "origin_chain": "linea",
+          "destination_chain": "base",
+          "origin_symbol": "DAI",
+          "APPROX_QUANTILES(fill_time_secs, 100)[OFFSET(75)]": 214
```
james-a-morris, line 9: Any chance we can modify this to make it more human readable? This looks like an SQL query

james-a-morris (commented):
Makes sense - just the question about the fill-times-preset JSON. I feel like there's a chance a third party could reference this and we should make it convenient/interpretable as to what this field is.

src/data/fill-times-preset.json
dohaki: I wanted to make the update progress as frictionless as possible. Currently you can just export from Preset and directly use the JSON file without any transformation steps.

james-a-morris (approved):
I agree with the want to make things frictionless, but I just feel like the fieldname in this JSON is clunky and doesn't provide a clean interface if a third party was to reference it.

Maybe we could add to the README.md or even the docs.across.to what this file is and how an external project could reference it.

dohaki:
> I agree with the want to make things frictionless, but I just feel like the fieldname in this JSON is clunky and doesn't provide a clean interface if a third party was to reference it.
> 
> 
> 
> Maybe we could add to the README.md or even the docs.across.to what this file is and how an external project could reference it.

But it is not meant to be referenced externally. Even if we would provide an interface for this data then this would be via an API. So I think this would just cause us more work to clean the JSON from preset. And even though the column is clunky, it is descriptive.

PR #1038, opened 2024-05-15
feat: handle deposit tx reverts gracefully

dohaki:
Fixes ACX-2144

Example URL of reverted deposit https://app-frontend-v3-git-deposit-reverted-screen-uma.vercel.app/bridge/0x4e56d36fb17e0f945826d6d80b59638904e3e81fda7ecf677a4601e566e32809?originChainId=137&destinationChainId=1&inputTokenSymbol=USDC.e&outputTokenSymbol=USDC


src/views/Bridge/utils.ts
```diff
@@ -363,21 +363,39 @@ export function getAllChains() {
...
-  const fromChain = Number(params.get("from"));
-  const toChain = Number(params.get("to"));
-  const token = params.get("token");
+  const fromChain = Number(
+    params.get("from") || params.get("fromChain") || params.get("originChainId")
+  );
+  const toChain = Number(
+    params.get("to") ||
+      params.get("toChain") ||
+      params.get("destinationChainId")
```
james-a-morris, line 372: OOC is there a reason we have all of these possible options?

james-a-morris (approved):
One question - but if we need to get it in now we can talk about it in post

src/views/Bridge/utils.ts
dohaki: Yea, we had a lot of confusion by us and the community what the correct params are... Because we are not very consistent. Our API for example uses `originChain/destinationChain`, historically our FE used `from/to`, etc.

PR #1041, opened 2024-05-15
feat: make default swap slippage configurable

dohaki:
I noticed a few swaps that were reverting on Polygon. Probably due to too low slippage? But in case we want to change it dynamically, I prepared this PR to override via env var.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -854,3 +854,8 @@ export const vercelApiBaseUrl =
 
 export const bnUint32Max = utils.bnUint32Max;
 export const bnZero = utils.bnZero;
+
+// Swap slippage in %, 1 = 1%
+export const defaultSwapSlippage = Number(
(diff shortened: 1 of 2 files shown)


src/utils/constants.ts
```diff
@@ -854,3 +854,8 @@ export const vercelApiBaseUrl =
 
 export const bnUint32Max = utils.bnUint32Max;
 export const bnZero = utils.bnZero;
+
+// Swap slippage in %, 0.5 = 0.5%
+export const defaultSwapSlippage = Number(
+  process.env.REACT_APP_DEFAULT_SWAP_SLIPPAGE || 0.5
```
james-a-morris, line 860: Ah gotcha so we intend to scale 0-100 in the ENV

james-a-morris (approved):

mrice32 (approved):

PR #1043, opened 2024-05-16
feat: add `recommendedDepositInstant` field to `GET /limits` response

dohaki:


api/_constants.ts
```diff
@@ -196,3 +196,9 @@ export const DEFI_LLAMA_POOL_LOOKUP: Record<string, string> = {
 
 export const DEFAULT_SIMULATED_RECIPIENT_ADDRESS =
   "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
+
+export const DEFAULT_RECOMMENDED_DEPOSIT_INSTANT_LIMITS = {
+  [TOKEN_SYMBOLS_MAP.ETH.symbol]: "2",
+  [TOKEN_SYMBOLS_MAP.WETH.symbol]: "2",
+  [TOKEN_SYMBOLS_MAP.USDC.symbol]: "5000",
+};
```
james-a-morris, line 204: Should we keep this in wei so that we don't need to parse on the fly?

dohaki: I wanted to stay consistent with the way we handle LP cushions

james-a-morris (approved):
This is a succinct change. LGTM +1

api/limits.ts
```diff
@@ -177,20 +178,31 @@ const handler = async (
...
-        maxBN(...fullRelayerBalances, ...transferRestrictedBalances),
-        liquidReserves
-      ).toString(),
+      maxDepositInstant,
       // Same as above.
       maxDepositShortDelay: minBN(
         maxBN(...transferBalances, ...transferRestrictedBalances),
         liquidReserves
       ).toString(),
+      recommendedDepositInstant: recommendedDepositInstant
```
james-a-morris, line 201: Will this expression be truthy if `recommendedDepositInstant` is 0? If so LGTM

dohaki: Yes, it will return truthy if it is `0` because `getRecommendedDepositInstantLimit` returns either `string` or `undefined`.

PR #1045, opened 2024-05-17
refactor: reduce bundle size

dohaki:
I noticed that our bundle size grew a lot in the last few months. In the last production build, the main bundled index file has a gzipped size of over 2.4 MB (!)
image

With some optimizations of this PR, I was able to reduce it down to ~1.3 MB gzipped. Which is still pretty large but definitely better than before.
image

The optimizations:
- Add `GET /pools-user` handler to remove the need of pulling in non-treeshakeable modules from `@across-protocol/contracts`
- Specify import paths for `@across-protocol/sdk` for better treeshaking


nicholaspai (commented):
This is a great initiative thanks! Will let someone more familiar with code review and approve

james-a-morris (approved):
Amazing work

PR #1046, opened 2024-05-22
feat(api): add docs on `recommendedDepositInstant`

dohaki:


api-docs-openapi.yaml
```diff
@@ -616,6 +617,14 @@ components:
             greater than this, the estimate should be between 2-4 hours for a
             slow relay to be processed from the mainnet pool.
           example: 2045367713809
+        recommendedDepositInstant:
+          type: string
+          description: >
+            The recommended deposit size that can be relayed "instantly" on the
+            destination chain. Instantly means that there is relayer capital
+            readily available and that a relayer is expected to relay within
+            seconds to 5 minutes of the deposit.
```
james-a-morris, line 626: Maybe we can add a word or two that the value is in the smallest unit of that specific token

dohaki: https://github.com/across-protocol/frontend-v3/pull/1046/commits/98041f4ed90eda8622ae11202e1196f04ab5f48a

james-a-morris (approved):

PR #1044, opened 2024-05-16
feat: support mode network

dohaki:
Draft until fully tested

PR #1047, opened 2024-05-23
chore: bump sdk

james-a-morris:
Bumps the SDK and changes how we call instantiate queries


dohaki (commented):
We have these changes in this PR https://github.com/across-protocol/frontend-v3/pull/1044 already. Any reason why you want it separate?

james-a-morris:
> We have these changes in this PR #1044 already. Any reason why you want it separate?

I was hoping to isolate it so that the diff was smaller and any regressions could be pinpointed to this without having to roll back #1044

dohaki (approved):

PR #1048, opened 2024-05-28
fix: prevent redundant rpc calls before sending bridge tx

dohaki:
This PR attempts to fix a bug that would in rare cases occur when a user accidentally changes the network before certain RPC calls within `ethers.providers.JsonRpcSigner.sendTransaction` are resolved.

To prevent this, we manually serialize and send/sign the tx directly and do not call redundant RPC calls. 

Relevant lines of code in ethers
- https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L276
- https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/providers/src.ts/json-rpc-provider.ts#L240-L243

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -1,4 +1,4 @@
-import { ethers, BigNumber } from "ethers";
+import { ethers, BigNumber, providers } from "ethers";
 import { utils } from "@across-protocol/sdk-v2";
 
 import {
@@ -204,7 +204,7 @@ type NetworkMismatchHandler = (
  * @returns The transaction response obtained after sending the transaction.
  */
 export async function sendDepositTx(
-  signer: ethers.Signer,
+  signer: ethers.providers.JsonRpcSigner,
   {
     fromChain,
     tokenAddress,
@@ -253,7 +253,7 @@ export async function sendDepositTx(
 }
 
 export async function sendDepositV3Tx(
-  signer: ethers.Signer,
+  signer: ethers.providers.JsonRpcSigner,
   {
     fromChain,
     amount,
@@ -316,7 +316,7 @@ export async function sendDepositV3Tx(
 }
 
 export async function sendSwapAndBridgeTx(
-  signer: ethers.Signer,
+  signer: ethers.providers.JsonRpcSigner,
   {
     fromChain,
     toAddress: recipient,
@@ -471,7 +471,7 @@ async function _getSpokePoolAndVerifier({
 async function _tagRefAndSignTx(
   tx: ethers.PopulatedTransaction,
   referrer: string,
-  signer: ethers.Signer,
+  signer: ethers.providers.JsonRpcSigner,
   originChainId: ChainId,
   destinationChainId: ChainId,
   onNetworkMismatch?: NetworkMismatchHandler
@@ -499,5 +499,17 @@ async function _tagRefAndSignTx(
     );
   }
 
-  return signer.sendTransaction(tx);
+  // NOTE: We are manually hexlifying the transaction here and signing/sending tx directly
+  // instead of calling `signer.sendTransaction(tx)` to drop redundant RPC calls that
+  // might open a time window where a user could change their network by accident.
+  const hexTx = providers.JsonRpcProvider.hexlifyTransaction(
+    {
+      ...tx,
+      from: await signer.getAddress(),
+      chainId: originChainId,
+    },
+    { from: true }
+  );
+  const txHash = await signer.provider.send("eth_sendTransaction", [hexTx]);
+  return signer.provider.getTransaction(txHash);
 }


src/utils/bridge.ts
```diff
@@ -499,5 +499,17 @@ async function _tagRefAndSignTx(
...
+  const hexTx = providers.JsonRpcProvider.hexlifyTransaction(
+    {
+      ...tx,
+      from: await signer.getAddress(),
+      chainId: originChainId,
+    },
+    { from: true }
+  );
+  const txHash = await signer.provider.send("eth_sendTransaction", [hexTx]);
+  return signer.provider.getTransaction(txHash);
```
james-a-morris, line 514: Could we put this in a utility function? It may be worth including this for the pool + staking pages to be sure.

james-a-morris (commented):
In your testing did you find an easy way to reproduce this bug? I want to test with the preview link to see if this issue is still visible.

src/utils/bridge.ts
dohaki: I am not if we need this in other places... Because we don't have the `populateTransaction` + `sendTransaction` flow for normal `HubPool` targeting transactions.

dohaki: The additional RPC calls of `sendTransaction` are only redundant if we previously call `populateTransaction`

dohaki: > ooc, is there a theoretical risk for a user depositing raw ETH into the HubPool?

Do you mean if the user is adding liquidity for example? In that case, there could be a tiny window where a user changes the network before some RPC requests resolve, yes.

james-a-morris (approved):

src/utils/bridge.ts
```diff
@@ -499,5 +499,17 @@ async function _tagRefAndSignTx(
...
 
-  return signer.sendTransaction(tx);
+  // NOTE: We are manually hexlifying the transaction here and signing/sending tx directly
+  // instead of calling `signer.sendTransaction(tx)` to drop redundant RPC calls that
+  // might open a time window where a user could change their network by accident.
+  const hexTx = providers.JsonRpcProvider.hexlifyTransaction(
+    {
+      ...tx,
+      from: await signer.getAddress(),
+      chainId: originChainId,
```
mrice32, line 509: Trying to better understand the bug.

Is the trick that we're manually specifying the chainId to ensure it is paired correctly with the to address in the txn? That way, the wallet is forced to sign with the correct chain id, reject the txn entirely, or ask the user to switch chains to sign.

And the order of events that(we think) was previously happening was:
1. Across builds txn assuming chain id X.
2. User switches network to chain id Y.
3. Across sends txn to ethers, where the txn built for X is paired with chain id Y, which is not the chain it was intended for.
4. User signs a transaction on the wrong chain.

Am I understanding this correctly?

Really subtle issue if so. Amazing work

ab697d53cb5378b44dbd84e1fb14721d: > I initially thought that specifying the `chainId` here would enforce the correct chain before signing with

I suspect something else might be a play in this issue.  I expect that the an RPC error similar to below to occur if a change happens without being handled in the app:
`underlying network changed (event="changed", network={"name":"optimism","chainId":10,"ensAddress":null}, detectedNetwork={"name":"arbitrum","chainId":42161,"ensAddress":null,"_defaultProvider":null}, code=NETWORK_ERROR, version=providers/5.7.2)`

dohaki: hey @ab697d53cb5378b44dbd84e1fb14721d, thanks for your input.

> I suspect something else might be at play in this issue. I expect that an RPC error similar to below would occur if a network change happens without being handled in the app:
underlying network changed (event="changed", network={"name":"optimism","chainId":10,"ensAddress":null}, detectedNetwork={"name":"arbitrum","chainId":42161,"ensAddress":null,"_defaultProvider":null}, code=NETWORK_ERROR, version=providers/5.7.2)

It doesn't actually throw that error... Possibly because we explicitly allow any chain that the bridge supports to be connected. 

> One way to do this is by calling the function of an instance of the contract on that network, expect the RPC call to fail any time this is not connected to the current network. This should eliminate any possibility of a race condition between events.

If you look at the code [here](https://github.com/across-protocol/frontend-v3/pull/1048/files#diff-5d22ab6b9a7272e68f42731b880e896d1671416cc379fcf34bf6106cc5b9658eR490) you can see that we do a check like that. That is why I think the only place where the race condition can happen is within the `sendTransaction` implementation.

dohaki: I created an ethers patch where I artificially slow down some of the RPC calls which allows to reproduce the race condition here https://app-frontend-v3-git-demo-changed-network-uma.vercel.app/bridge.

I attached a screen recording in the respective PR https://github.com/across-protocol/frontend-v3/pull/1055

ab697d53cb5378b44dbd84e1fb14721d: > It doesn't actually throw that error... Possibly because we explicitly allow any chain that the bridge supports to be connected.

I understood that part the difference here is I'm suggesting that if you instantiated the contract on that origin chain, when you exeucte your transaction the RPC would trigger the error I mentioned. I am able to reproduce the network changed error when replicating your issue.  I think a secondary benefit of this approach is the function can be simulated to verify the outcome before mining the transaction.

dohaki: @ab697d53cb5378b44dbd84e1fb14721d Yes, I understand your suggestion. But we already do that approach. If you look here https://github.com/across-protocol/frontend-v3/blob/287b7b2c2d72f962632a25016b837ddb0ccb278b/src/utils/bridge.ts#L438-L439
we use these instances of the `SpokePool` contracts connected to the correct provider of the `origin` chain. Then we populate the transaction for example here https://github.com/across-protocol/frontend-v3/blob/287b7b2c2d72f962632a25016b837ddb0ccb278b/src/utils/bridge.ts#L292. This runs a simulation to determine gas costs, etc.

The race condition happens after all these steps right before sending the transaction to the `JsonRpcSigner/Web3Provider` like MetaMask.

dohaki:
Closing as we haven't been seeing this issues with our latest changes

PR #1049, opened 2024-05-29
docs: fix api examples

dohaki:
Fixes ACX-2169

### DIFF
--- a/api-docs-openapi.yaml
+++ b/api-docs-openapi.yaml
@@ -39,6 +39,7 @@ paths:
             _Example:_ 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2
         - name: inputToken
           in: query
+          example: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
           schema:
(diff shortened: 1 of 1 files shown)


api-docs-openapi.yaml
```diff
@@ -209,26 +205,13 @@ paths:
...
-                        },
-                      "relayerGasFee":
-                        {
-                          "pct": "276407094864283",
-                          "total": "276407094864283",
-                        },
-                      "lpFee":
-                        { "pct": "4552495218411721", "total": "1708047000" },
-                    },
+                  "totalRelayFee":
```
mrice32, line 208: Don't we have some deprecated fields that aren't listed here? Is it typical to not mention deprecated fields in docs? (not strongly opinionated)

mrice32 (approved):
LGTM!

api-docs-openapi.yaml
dohaki: Yea, I think the idea was to "officially" deprecate from the docs and soon also from the actual response

PR #1055, opened 2024-05-31
feat: add patch with artificially delayed rpc calls in `sendTransaction`

dohaki:
Demo PR to reproduce bug of #1048 

https://github.com/across-protocol/frontend-v3/assets/17645687/87696864-56c2-443b-8ac7-4966bc7c9356


dohaki:
Closing because we fixed error differently

PR #1061, opened 2024-06-05
feat(deposit-track): expose endpoint

james-a-morris:
This PR adds an endpoint to track deposits given a `originChainId` and a `depositHash`. Because a deposit hash can include multiple deposit records, an array is returned.


api/track-deposit.ts
```diff
@@ -0,0 +1,60 @@
...
+  handleErrorCondition,
+  positiveIntStr,
+  resolveFillStatusOfDeposits,
+  sendResponse,
+  validTxnHash,
+} from "./_utils";
+
+const TrackDepositQueryParamsSchema = object({
+  originChainId: positiveIntStr(),
+  depositHash: validTxnHash(),
```
nicholaspai, line 16: what if user wants to just specify `depositId`?

nicholaspai: i don’t think specifying the transaction has is the right approach, it doesnt’ handle the case where there are multiple deposits in the same hash. Moreover, searching for a log seems inefficient.
Instead, why not make caller pass in the keccak256(deposit) i.e. the keccak256DepositHash

nicholaspai:
let's close this in favor of using the scraper API

james-a-morris:
Closing as this was absorbed by the scraper work

PR #1064, opened 2024-06-11
fix(recommendedDepositInstant): floor as needed

james-a-morris:
Technically, `recommendedDepositInstant` is a hardcoded value that can be greater than `maxDepositInstant` during times of volatility. We may want to limit the `recommendedDepositInstant` to the maximum deposit that can be filled without a slow fill.


dohaki (approved):
Good call

PR #1065, opened 2024-06-12
feat: add buffer multipliers to limits

dohaki:
Closes ACX-2214


api/limits.ts
```diff
@@ -179,10 +179,14 @@ const handler = async (
     );
 
     const maxDepositInstant = minBN(
-      maxBN(...fullRelayerBalances, ...transferRestrictedBalances),
+      maxBN(...fullRelayerBalances, ...transferRestrictedBalances), // balances on destination chain
```
nicholaspai, line 182: Unrelated but we should make sure we add the fast relayer EOA to the env list cc @pxrl

api/suggested-fees.ts
```diff
@@ -196,11 +197,26 @@ const handler = async (
...
+        inputToken.address,
+        outputToken.address,
+        computedOriginChainId,
+        destinationChainId
+      ),
+    ]);
+
+    // Prevent quotes for slow-fills
+    if (amount.gt(limits.maxDepositShortDelay)) {
+      throw new InputError("Amount exceeds max deposit limit for short delay");
```
nicholaspai, line 217: should the log include the limit value?

src/views/Bridge/hooks/useBridge.ts
```diff
@@ -67,7 +66,7 @@ export function useBridge() {
     parsedAmount,
     quotedFees?.isAmountTooLow,
     maxBalance,
-    quotedLimits?.maxDeposit,
+    quotedFees?.limits.maxDepositShortDelay,
```
nicholaspai, line 69: What happens if the `quotedFee` is undefined in the case where the /suggested-fees endpoint throws an error because the amount > maxDepositShortDelay?

nicholaspai (approved):
This looks good, i've briefly QA'd the preview API

nicholaspai:
Screenshot 2024-06-12 at 19 46 13


when i type in an amount greater than the limit my quote loads indefinitely and I see a 4xx error logged

api/suggested-fees.ts
dohaki: good point. added here https://github.com/across-protocol/frontend-v3/pull/1065/commits/84e32246922d9346a698ea9e50be0ad102f1146f

dohaki:
> when i type in an amount greater than the limit my quote loads indefinitely and I see a 4xx error logged

Yes, good catch. I fixed the behavior now so that it displays `Insufficient bridge liquidity` and stops displaying the loaders
image

api/limits.ts
dohaki: Added the address 👍

api/_utils.ts
```diff
@@ -1417,27 +1448,76 @@ export function isSwapRouteEnabled({
...
+ * @param symbol The token symbol to retrieve the multiplier for.
+ * @param fromChainId The chain ID of the origin chain.
+ * @param toChainId The chain ID of the destination chain.
+ */
+export const getDepositShortDelayMultiplier = _makeLimitsBufferMultiplierGetter(
+  "DEPOSIT_SHORT_DELAY_BUFFER_MULTIPLIER",
+  DEFAULT_LIMITS_BUFFER_MULTIPLIERS.depositShortDelay
+);
+
+function _makeLimitsBufferMultiplierGetter(
```
james-a-morris, line 1508: Nit: do we need this preceded with `_`?

james-a-morris (approved):

api/_utils.ts
dohaki: removed

PR #1066, opened 2024-06-12
feat(domain-identifier): add domain identifier to txn calls

james-a-morris:
Appends a domain identifier to the txn after the referrer identifier. If a provider references `domain` in the `build-deposit-tx` and we have a mapping then the endpoint will return calldata with the appended 4byte identifier.

Example Call:
```
/api/build-deposit-tx?amount=10000&token=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48&destinationChainId=10&originChainId=1&recipient=0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D&relayerFeePct=5&quoteTimestamp=5&domain=across
```

### DIFF
--- a/api/build-deposit-tx.ts
+++ b/api/build-deposit-tx.ts
@@ -14,6 +14,7 @@ import {
   tagReferrer,
   validAddressOrENS,
 } from "./_utils";
+import { EXTERNAL_DOMAIN_TO_BYTE_MAP } from "./_constants";
 
 const BuildDepositTxQueryParamsSchema = type({
   amount: parsableBigNumberString(),
@@ -27,6 +28,9 @@ const BuildDepositTxQueryParamsSchema = type({
   maxCount: optional(boolStr()),
   referrer: optional(validAddressOrENS()),
   isNative: optional(boolStr()),
+  // Calling domain that issued the request. This will be
+  // used to tag the transaction with the domain's identifier.
+  domain: optional(string()),
 });
 
 type BuildDepositTxQueryParams = Infer<typeof BuildDepositTxQueryParamsSchema>;
@@ -58,6 +62,7 @@ const handler = async (
       maxCount = ethers.constants.MaxUint256.toString(),
       referrer,
       isNative: isNativeBoolStr,
+      domain,
     } = query;
 
     recipient = ethers.utils.getAddress(recipient);
@@ -89,6 +94,14 @@ const handler = async (
 
     // do not tag a referrer if data is not provided as a hex string.
     tx.data = referrer ? await tagReferrer(tx.data!, referrer) : tx.data;
+
+    // Append the domain identifier to the data field of the transaction.
+    const domainByteCode = domain
+      ? EXTERNAL_DOMAIN_TO_BYTE_MAP[domain]
+      : undefined;
+    if (tx.data && domainByteCode) {
+      tx.data = ethers.utils.hexConcat([tx.data, "0x1DC0de", domainByteCode]);
+    }
 
     const responseJson = {
       data: tx.data,
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -6,7 +6,7 @@ import {
   fixedPointAdjustment,
   referrerDelimiterHex,
 } from "./constants";
-import { tagAddress } from "./format";
+import { tagAcrossDomain, tagAddress } from "./format";
 import { getProvider } from "./providers";
 import { getFastFillTimeByRoute } from "./fill-times";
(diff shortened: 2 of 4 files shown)


api/build-deposit-tx.ts
```diff
@@ -90,6 +95,14 @@ const handler = async (
     // do not tag a referrer if data is not provided as a hex string.
     tx.data = referrer ? await tagReferrer(tx.data!, referrer) : tx.data;
 
+    // Append the domain identifier to the data field of the transaction.
+    const domainByteCode = domain
```
nicholaspai, line 99: Can we not just load the string from the domain parameter?

We could enforce its in hex and in 4 bytes here

src/utils/bridge.ts
```diff
@@ -477,10 +477,11 @@ async function _tagRefAndSignTx(
   onNetworkMismatch?: NetworkMismatchHandler
 ) {
   // do not tag a referrer if data is not provided as a hex string.
-  tx.data =
+  tx.data = tagAcrossDomain(
```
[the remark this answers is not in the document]
james-a-morris: Not entirely sure I follow - the Across domain is being appended in both cases

api/build-deposit-tx.ts
james-a-morris: https://github.com/across-protocol/frontend-v3/pull/1066/commits/8355a92536c1cbf8ee9114b10a56ec118d600825

api/build-deposit-tx.ts
```diff
@@ -90,6 +94,10 @@ const handler = async (
     // do not tag a referrer if data is not provided as a hex string.
     tx.data = referrer ? await tagReferrer(tx.data!, referrer) : tx.data;
 
+    if (tx.data && domainIdentifier) {
+      tx.data = ethers.utils.hexConcat([tx.data, "0x1DC0de", domainIdentifier]);
```
nicholaspai, line 98: Can we use the same logic as setting the referrer?

`tx.data = referrer ? await tagDomain(tx.data!, domainIdentifier) : tx.data;`

james-a-morris: Fair - done https://github.com/across-protocol/frontend-v3/pull/1066/commits/431d9b872649263972e3b78cbcd3915f5ece8f2a

api/_utils.ts
```diff
@@ -954,6 +954,16 @@ export async function tagReferrer(
...
 }
 
+export function tagDomain(dataHex: string, domainIdentifier: string): string {
+  if (!ethers.utils.isHexString(dataHex)) {
+    throw new Error("Data must be a valid hex string");
+  }
+  if (!ethers.utils.isHexString(domainIdentifier)) {
+    throw new Error("Domain identifier must be a valid hex string");
+  }
+  return ethers.utils.hexConcat([dataHex, "0x1DC0de", domainIdentifier]);
```
nicholaspai, line 964: one nit, can you declare a constant `0x1DC0de` to make it more clear what this hardcoded value is?

src/utils/format.ts
```diff
@@ -195,6 +195,11 @@ export function tagString(dataHex: string, tagString: string) {
   return tagHex(dataHex, stringToHex(tagString));
 }
 
+// tags the hardcoded across chain reference
+export function tagAcrossDomain(dataHex: string) {
+  return tagHex(dataHex, "0x0000", "0x1DC0de");
```
nicholaspai, line 200: See https://github.com/across-protocol/frontend-v3/pull/1066/files#r1637099015 about replacing the hardcoded `0x1DC0de` string.

On the `0x0000` is this what we're going to use to denote any deposits sent from the https://across.to domain?

nicholaspai (approved):
A couple nit questions, then LGTM! great work

src/utils/bridge.ts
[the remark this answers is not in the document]
dohaki: > I think removing the ref tag should be done when we sunset the referrals program

I think we sunsetted already, no? So any tagged referrer is currently already being ignored by the Scraper.

dohaki (approved):
Looking good 👍 One small nit on removing ref tags: AFAIK we could already remove the ref tags because earning ref rewards was disabled in May

src/utils/format.ts
james-a-morris: https://github.com/across-protocol/frontend-v3/pull/1066/commits/61c32da65eeabf8795fda217b87e5d38511516a9

james-a-morris: This is per Chase's docs

PR #1077, opened 2024-06-20
fix(limits): set `maxDeposit` equal to `maxDepositShortDelay`

dohaki:
Since we no longer provide fees if the amount > maxDepositShortDelay, but still have `maxDeposit` in the response body of `/limits`, integrators got confused. To prevent this, we set now `maxDeposit` to the same value as `maxDepositShortDelay`

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -191,18 +191,21 @@ const handler = async (
       computedOriginChainId,
       destinationChainId
     );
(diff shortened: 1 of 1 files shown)


nicholaspai (commented):
Can we now remove any of the liquid reserves queries or logic?

pxrl (approved):

nicholaspai (approved):

PR #1080, opened 2024-06-20
refactor(limits): refine buffer logic and default values

dohaki:
Refine default limits buffer multipliers and logic

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -177,13 +177,4 @@ export const DEFI_LLAMA_POOL_LOOKUP: Record<string, string> = {
 export const DEFAULT_SIMULATED_RECIPIENT_ADDRESS =
   "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
 
-export const DEFAULT_LIMITS_BUFFER_MULTIPLIERS = {
-  [TOKEN_SYMBOLS_MAP.DAI.symbol]: "0.8",
-  [TOKEN_SYMBOLS_MAP.USDT.symbol]: "0.8",
-  [TOKEN_SYMBOLS_MAP.WBTC.symbol]: "0.8",
-  [TOKEN_SYMBOLS_MAP.USDC.symbol]: "0.2",
-  [TOKEN_SYMBOLS_MAP.ETH.symbol]: "0.2",
-  [TOKEN_SYMBOLS_MAP.WETH.symbol]: "0.2",
-};
-
 export const DOMAIN_CALLDATA_DELIMITER = "0x1dc0de";
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -178,9 +178,12 @@ export const DEFAULT_SIMULATED_RECIPIENT_ADDRESS =
   "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
 
 export const DEFAULT_LIMITS_BUFFER_MULTIPLIERS = {
-  recommendedDepositInstant: "0.1",
-  depositInstant: "0.2",
-  depositShortDelay: "0.2",
+  [TOKEN_SYMBOLS_MAP.DAI.symbol]: "0.8",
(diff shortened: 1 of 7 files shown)


api/_constants.ts
```diff
@@ -178,9 +178,12 @@ export const DEFAULT_SIMULATED_RECIPIENT_ADDRESS =
   "0xBb23Cd0210F878Ea4CcA50e9dC307fb0Ed65Cf6B";
 
 export const DEFAULT_LIMITS_BUFFER_MULTIPLIERS = {
-  recommendedDepositInstant: "0.1",
-  depositInstant: "0.2",
-  depositShortDelay: "0.2",
+  [TOKEN_SYMBOLS_MAP.DAI.symbol]: "0.8",
```
nicholaspai, line 181: Nit: I think in general we shouldn't be hardcoding override values in the code and instead should rely on env vars but this is fine for now

api/_utils.ts
```diff
@@ -1481,76 +1481,14 @@ export function isSwapRouteEnabled({
...
-    DEFAULT_LIMITS_BUFFER_MULTIPLIERS.recommendedDepositInstant
+export function getLimitsBufferMultiplier(symbol: string) {
+  const limitsBufferMultipliers: Record<string, string> = process.env
+    .LIMITS_BUFFER_MULTIPLIERS
+    ? JSON.parse(process.env.LIMITS_BUFFER_MULTIPLIERS)
+    : {};
+  return ethers.utils.parseEther(
+    limitsBufferMultipliers[symbol] ||
+      DEFAULT_LIMITS_BUFFER_MULTIPLIERS[symbol] ||
+      "1"
```
nicholaspai, line 1492: By removing the `DEFAULT_LIMITS_BUFFER_MULTIPLIERS` constant as per my other comment we could shorten this a bit. Basically, either the buffer is specified in an env var or we default to 1, no intermediary hardcoded buffer

api/_utils.ts
```diff
@@ -1481,76 +1481,14 @@ export function isSwapRouteEnabled({
...
- * @param symbol The token symbol to retrieve the multiplier for.
- * @param fromChainId The chain ID of the origin chain.
- * @param toChainId The chain ID of the destination chain.
- */
-export const getRecommendedDepositInstantMultiplier =
-  makeLimitsBufferMultiplierGetter(
-    "RECOMMENDED_DEPOSIT_INSTANT_MULTIPLIER",
-    DEFAULT_LIMITS_BUFFER_MULTIPLIERS.recommendedDepositInstant
+export function getLimitsBufferMultiplier(symbol: string) {
+  const limitsBufferMultipliers: Record<string, string> = process.env
```
nicholaspai, line 1485: nit: can we cap the buffer to 100%? Not sure why we'd want to ever show more than 100%

nicholaspai (approved):
Looks great, left some nits

src/views/Bridge/components/BridgeForm.tsx
```diff
@@ -148,7 +150,7 @@ const BridgeForm = ({
           errorText={validationErrorTextMap[validationError].replace(
             "[MAX_DEPOSIT_SHORT_DELAY]",
```
nicholaspai, line 151: should this be just MAX_DEPOSIT..?

api/_constants.ts
dohaki: ok removed

api/_utils.ts
dohaki: yea makes sense! https://github.com/across-protocol/frontend-v3/pull/1080/commits/e161390500014181b6cea17638522b22a1d01df5

src/views/Bridge/components/BridgeForm.tsx
dohaki: https://github.com/across-protocol/frontend-v3/pull/1080/commits/e161390500014181b6cea17638522b22a1d01df5

PR #1082, opened 2024-06-21
feat(liquidReserves): Add liquid-reserves endpoint that relayers can use

nicholaspai:
Currently relayers query /limits to figure out the `maxDeposit` they should fill, however `maxDeposit` currently applies a buffer which means the relayer can't fill any deposits that are larger than the `liquidReserves * buffer` but smaller than `liquidReserves`

Create a separate endpoint with a simpler interface (i.e. only need to define L1 token and origin chain ID and both are required params) to get limits for relayer. This should simplify the relayer code in the AcrossApiClient

Signed-off-by: nicholaspai <npai.nyc@gmail.com>

### DIFF
--- a/api/liquid-reserves.ts
+++ b/api/liquid-reserves.ts
@@ -74,15 +74,17 @@ const handler = async (
       }),
     ];
 
-    const [multicallOutput] = await Promise.all([
+    const [multicallOutput, ...lpCushions] = await Promise.all([
       callViaMulticall3(provider, multiCalls, { blockTag: BLOCK_TAG_LAG }),
+      ...l1TokenDetails.map((_l1Token) =>
+        Promise.resolve(
+          ethers.utils.parseUnits(
+            getLpCushion(_l1Token!.symbol),
+            _l1Token!.decimals
+          )
+        )
+      ),
     ]);
-    const lpCushions = l1TokenDetails.map((_l1Token) =>
-      ethers.utils.parseUnits(
-        getLpCushion(_l1Token!.symbol),
-        _l1Token!.decimals
-      )
-    );
     const liquidReservesForL1Tokens = multicallOutput.slice(
       parsedL1Tokens.length
     );
--- a/api/liquid-reserves.ts
+++ b/api/liquid-reserves.ts
@@ -14,14 +14,12 @@ import {
   getProvider,
   getTokenByAddress,
   handleErrorCondition,
-  positiveIntStr,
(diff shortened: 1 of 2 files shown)


api/liquid-reserves.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const [multicallOutput] = await Promise.all([
+      callViaMulticall3(provider, multiCalls, { blockTag: BLOCK_TAG_LAG }),
+    ]);
+
+    const { liquidReserves } = multicallOutput[1];
+    const lpCushion = ethers.utils.parseUnits(
+      getLpCushion(l1TokenDetails.symbol, Number(originChainId)),
+      l1TokenDetails.decimals
+    );
```
pxrl, line 74: This has no dependency on the liquidReserves query, so for a very (very) small latency reduction we could define this as an async closure and execute it in parallel w/ the Promise at line 66.

api/liquid-reserves.ts
```diff
@@ -0,0 +1,95 @@
...
+) => {
+  const logger = getLogger();
+  logger.debug({
+    at: "LiquidReserves",
+    message: "Query data",
+    query,
+  });
+  try {
+    assert(query, LiquidReservesQueryParamsSchema);
+    const { l1Token, originChainId } = query;
```
pxrl, line 45: I think we can probably drop the origin chain component instead, and just require the caller to supply a token that's known to the HubPool. This is anyway what the bots will do now. wdyt?

api/liquid-reserves.ts
pxrl: The overhead of setting that up might well exceed the benefit tbf.

api/liquid-reserves.ts
```diff
@@ -0,0 +1,112 @@
...
+) => {
+  const logger = getLogger();
+  logger.debug({
+    at: "LiquidReserves",
+    message: "Query data",
+    query,
+  });
+  try {
+    assert(query, LiquidReservesQueryParamsSchema);
+    const { l1Tokens } = query;
```
pxrl, line 46: I'm not sure we can supply an array here. I suspect it needs to be a string that we then transform into an array via String.split(), and then we can assert that the result is valid according to the schema.

pxrl: Resolved :+1:

pxrl (approved):
LGTM - this tests fine locally.

PR #1085, opened 2024-06-26
hotfix(wsteth-acx): resolve underlying subgraph error

james-a-morris:
Due to an upgrade, we can no longer rely on the balancer-hosted subgraph to fetch the pool data. We need to use a new on-chain subgraph, which requires us to hardcode the URLs for each required subgraph. The caveat is that we now need an API key to access the subgraphs.

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -115,6 +115,8 @@ export const defaultRelayerAddressOverridePerToken: Record<
 export const defaultRelayerAddressOverride =
   process.env.RELAYER_ADDRESS_OVERRIDE;
 
+export const theGraphAPIKey = process.env.THE_GRAPH_API_KEY;
+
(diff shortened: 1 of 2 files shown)


api/_constants.ts
```diff
@@ -115,6 +115,8 @@ export const defaultRelayerAddressOverridePerToken: Record<
 export const defaultRelayerAddressOverride =
   process.env.RELAYER_ADDRESS_OVERRIDE;
 
+export const theGraphAPIKey = process.env.THE_GRAPH_API_KEY;
```
nicholaspai, line 118: Nit prefer removing "THE"

nicholaspai (approved):

dohaki (approved):
oh wow nice catch!

amateima (approved):

PR #1086, opened 2024-06-26
feat: add polygon amoy support

dohaki:
Add support for Polygon Amoy testnet


james-a-morris (approved):
OOC - how long did this take to implement? We could use this to track time to integrate a "new" chain. cc @pxrl

pxrl:
> OOC - how long did this take to implement? We could use this to track time to integrate a "new" chain. cc @pxrl

Yeah it's a good temp. check. Dong-Ha seemed quite fast to do this, but there were prerequisite changes upstream to the contract configurations that were cumbersome and took a long time. I think we have some opportunities to script those.

pxrl (approved):
LGTM!

PR #1088, opened 2024-06-27
fix: correctly infer output token

dohaki:


pxrl (approved):
I tested this OK with USDC.e and WETH as token and originChainId zkSync :+1:

PR #1090, opened 2024-06-27
feat: add lisk chain support

dohaki:
Closes ACX-2265 and ACX-2257


james-a-morris (approved):
LGTM - when ready - but we should sync with master to ensure that everything works

PR #1092, opened 2024-06-27
fix(Amoy): Bump constants & SDK

pxrl:
Constants now includes an ETH address for Polygon. This is an implicit requirement due to the way some address -> token resolution functions work.


pxrl:
Will update w/ a yarn.lock sync once the packages are available.

dohaki (approved):

PR #1093, opened 2024-06-27
refactor: clean up assets/icons/logos

dohaki:
While working on scripts and improvements to make adding newer chains more streamlined, I noticed that our `src/assets` folder accumulated a lot of bloat over the last few months. This PR removes redundant or unused files and puts every icon into `src/assets/icons`.

In a subsequent PR, I will also clean up token and chain logos as part of some helper scripts I wrote for adding new chain assets on the FE.


james-a-morris (approved):
Amazing work

PR #1094, opened 2024-06-27
feat(op-rewards): extend to base and mode

james-a-morris:
Adds op-rewards support to base and mode on the FE.


dohaki (approved):
Pretty straightforward! Good job on building this in a way that allows us to easily enable on other chains 👍

amateima (approved):

PR #1095, opened 2024-06-28
fix: display initially quoted est. time

dohaki:
This fixes an issue where in-flight deposits close to the short-delay threshold, would cause an estimated time update in the UI. See here https://risklabs.slack.com/archives/C03HV8CFL80/p1719565303568349

We fix this by retrieving the estimated fill time based on the quoted limits. Previously the limits value would update which caused the bug.


src/utils/bridge.ts
```diff
@@ -106,8 +106,8 @@ export type ConfirmationDepositTimeType = {
 export const getConfirmationDepositTime = (
   amount: BigNumber,
   limits: BridgeLimitInterface,
-  toChain: ChainId,
   fromChain: ChainId,
+  toChain: ChainId,
```
james-a-morris, line 110: Did we mean to change this direction?

james-a-morris (commented):
Can you add context to this?

src/utils/bridge.ts
dohaki: Yea, I got confused by the order of the params and refactored it.

dohaki:
> Can you add context to this?

Sorry about that. Added a description

james-a-morris (approved):

PR #1097, opened 2024-07-01
feat: swap price impact warning

dohaki:
Closes ACX-2285

### DIFF
--- a/src/hooks/useSwapQuote.ts
+++ b/src/hooks/useSwapQuote.ts
@@ -1,12 +1,7 @@
 import { BigNumber } from "ethers";
 import { useQuery } from "react-query";
 
-import {
-  defaultRefetchInterval,
-  getConfig,
-  swapQuoteQueryKey,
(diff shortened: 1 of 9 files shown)


src/hooks/useSwapQuote.ts
```diff
@@ -69,7 +64,7 @@ export function useSwapQuoteQuery(params: SwapQuoteQueryKeyParams) {
     },
     {
       enabled: !!params.swapTokenSymbol,
-      refetchInterval: defaultRefetchInterval,
+      refetchInterval: 5_000,
```
james-a-morris, line 67: Why did we hardcode this?

james-a-morris (approved):
One change - any chance I can see a screenshot of what this is supposed to look like?

src/hooks/useSwapQuote.ts
dohaki: The default is 15 seconds and I want this to run more frequently.

PR #1101, opened 2024-07-03
feat: add blast and blast-sepolia support

dohaki:
Closes ACX-2091 and ACX-2092


scripts/chain-configs/blast/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default {
+  ...chainInfoBase,
+  logoPath: "./assets/logo.svg",
+  grayscaleLogoPath: "./assets/grayscale-logo.svg",
+  spokePool: "0x2D509190Ed0172ba588407D4c2df918F955Cc6E1",
+  chainId,
+  publicRpcUrl: "https://blast.din.dev/rpc",
+  blockTimeSeconds: 2,
+  tokens: ["WETH", "ETH", "USDB"],
```
bmzig, line 15: We also have WBTC enabled for Blast. (Not sure if that needs to go here).

dohaki: Ah ok yea good catch. Will add

dohaki: https://github.com/across-protocol/frontend/pull/1101/commits/01682fccf1b9602b005b01efe789bed10a81c7dc

scripts/chain-configs/blast/index.ts
```diff
@@ -0,0 +1,18 @@
...
+const chainId = CHAIN_IDs.BLAST;
+const chainInfoBase = PUBLIC_NETWORKS[chainId];
+
+export default {
+  ...chainInfoBase,
+  logoPath: "./assets/logo.svg",
+  grayscaleLogoPath: "./assets/grayscale-logo.svg",
+  spokePool: "0x2D509190Ed0172ba588407D4c2df918F955Cc6E1",
+  chainId,
+  publicRpcUrl: "https://blast.din.dev/rpc",
```
bmzig, line 13: Do we know if this RPC works? Fwiw it doesn't work when using cast. There is another public RPC here: https://rpc.blast.io. Not sure if this is used in the FE.

src/constants/chains/configs.ts
```diff
@@ -147,6 +157,40 @@ export const baseSepolia = {
...
 
+export const blast = {
+  name: "Blast",
+  fullName: "Blast",
+  chainId: 81457,
+  logoURI: blastLogo,
+  grayscaleLogoURI: blastGrayscaleLogo,
+  logoSvg: blastLogoSvg,
+  grayscaleLogoSvg: blastGrayscaleLogoSvg,
+  rpcUrl: "https://blast.din.dev/rpc",
```
bmzig, line 168: Same as other comment.

bmzig (approved):
LGTM subject to the other comments I have.

scripts/chain-configs/blast/index.ts
dohaki: Hmm yea I don't remember why I used the DIN RPC instead of the other one.. Will change

PR #1102, opened 2024-07-04
feat: decrease /limits cache time

dohaki:
Closes ACX-2294

The env vars for the limits multipliers are also adjusted. We still need to set up some schedulers to keep the cache fresh for major routes so that performance is not impacted too much as suggested by @mrice32

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -267,7 +267,7 @@ const handler = async (
       message: "Response data",
       responseJson,
     });
-    // Respond with a 200 status code and 15 seconds of cache cache with
+    // Respond with a 200 status code and 15 seconds of cache with
     // 45 seconds of stale-while-revalidate.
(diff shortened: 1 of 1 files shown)


james-a-morris (approved):

api/limits.ts
```diff
@@ -267,9 +267,9 @@ const handler = async (
       message: "Response data",
       responseJson,
     });
-    // Respond with a 200 status code and 4 minutes of cache cache with
-    // a minute of stale-while-revalidate.
-    sendResponse(response, responseJson, 200, 240, 60);
+    // Respond with a 200 status code and 15 seconds of cache cache with
```
pxrl, line 270: cache cache?

```suggestion
    // Respond with a 200 status code and 15 seconds of cache with
```

pxrl (approved):

PR #1111, opened 2024-07-12
feat: enable ENV to cap `maxDeposit` for depositing on an origin chain

james-a-morris:


mrice32 (approved):
LGTM! Great turnaround!

PR #1112, opened 2024-07-15
feat: extend chain max limits to all chains

melisaguevara:


james-a-morris (commented):
Will this override https://github.com/across-protocol/frontend/pull/1111?

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
+      computedOriginChainId,
+      inputToken.symbol,
+      includeDefaultMaxValues
+    );
+    if (chainUsdMaxBalance) {
+      const parsedChainUsdMaxBalance = ethers.utils.parseUnits(
+        chainUsdMaxBalance,
+        inputToken.decimals
+      );
+      const relayers = includeRelayerBalances
```
james-a-morris, line 237: `includeRelayerBalances` is currently parsing on tslint as `any`. Any chance we can re-evaluate line 175 to be:

```
    const liteChainIds: number[] =
      liteChainIdsEncoded === "" ? [] : JSON.parse(liteChainIdsEncoded);
```

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
+      const chainAvailableAmountForDeposits =
+        await getAvailableAmountForDeposits(
+          computedOriginChainId,
+          inputToken,
+          parsedChainUsdMaxBalance,
+          tokenPriceUsd,
+          relayers
+        );
+
+      chainMaxBoundary = chainMaxBoundary
```
james-a-morris, line 249: Can this be not null/undefined? We don't set it above

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
+        maxDepositInstant,
+        chainAvailableAmountForDeposits
+      );
+      maxDepositShortDelay = minBN(
+        maxDepositShortDelay,
+        chainAvailableAmountForDeposits
+      );
+    }
+
+    const chainUsdMaxDeposit = getChainMaxDepositUsd(
```
james-a-morris, line 265: Similar thoughts on this naming scheme

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
     }
 
-    if (originChainIsLiteChain) {
-      const relayers = [...fullRelayers, ...transferRestrictedRelayers];
-      const liteChainMaxBoundary = await getLiteChainMaxBoundary(
-        computedOriginChainId,
-        inputToken,
-        tokenPriceUsd,
-        relayers
+    // Apply chain max values when defined
```
james-a-morris, line 222: Could we do something like:

```
 // Apply chain max values when defined
    const includeDefaultMaxValues = originChainIsLiteChain;
    const includeRelayerBalances = originChainIsLiteChain;
    let chainAvailableAmountForDeposits: BigNumber | undefined;
    let chainInputTokenMaxDeposit: BigNumber | undefined;

    const chainUsdMaxBalance = getChainMaxBalanceUsd(
      computedOriginChainId,
      inputToken.symbol,
      includeDefaultMaxValues
    );

    const chainUsdMaxDeposit = getChainMaxDepositUsd(
      computedOriginChainId,
      inputToken.symbol,
      includeDefaultMaxValues
    );

    const parseInInputToken = (value: string) =>
      ethers.utils.parseUnits(value, inputToken.decimals);

    if (chainUsdMaxBalance) {
      const parsedChainUsdMaxBalance = parseInInputToken(chainUsdMaxBalance);
      const relayers = includeRelayerBalances
        ? [...fullRelayers, ...transferRestrictedRelayers]
        : [];
      chainAvailableAmountForDeposits = await getAvailableAmountForDeposits(
        computedOriginChainId,
        inputToken,
        parsedChainUsdMaxBalance,
        tokenPriceUsd,
        relayers
      );
    }

    if (chainUsdMaxDeposit) {
      const parsedChainUsdMaxDeposit = parseInInputToken(chainUsdMaxDeposit);
      chainInputTokenMaxDeposit = parsedChainUsdMaxDeposit
        .mul(sdk.utils.fixedPointAdjustment)
        .div(tokenPriceUsd);
    }

    const bnOrMax = (value?: BigNumber) => value ?? ethers.constants.MaxUint256;
    const resolvedChainAvailableAmountForDeposits = bnOrMax(
      chainAvailableAmountForDeposits
    );
    const resolvedChainInputTokenMaxDeposit = bnOrMax(
      chainInputTokenMaxDeposit
    );

    const chainMaxBoundary = minBN(
      resolvedChainAvailableAmountForDeposits,
      resolvedChainInputTokenMaxDeposit
    );

    minDeposit = minDeposit = minBN(
      minDeposit,
      resolvedChainInputTokenMaxDeposit,
      resolvedChainAvailableAmountForDeposits
    );
    minDepositFloor = minBN(
      minDepositFloor,
      resolvedChainAvailableAmountForDeposits,
      resolvedChainInputTokenMaxDeposit
    );
    maxDepositInstant = minBN(
      maxDepositInstant,
      resolvedChainAvailableAmountForDeposits,
      resolvedChainInputTokenMaxDeposit
    );
    maxDepositShortDelay = minBN(
      maxDepositShortDelay,
      resolvedChainAvailableAmountForDeposits,
      resolvedChainInputTokenMaxDeposit
    );
```

where we collapse some of the shared logic?

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
+    const includeDefaultMaxValues = originChainIsLiteChain;
+    const includeRelayerBalances = originChainIsLiteChain;
+    let chainMaxBoundary;
+
+    const chainUsdMaxBalance = getChainMaxBalanceUsd(
```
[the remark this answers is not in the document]
melisaguevara: Just to be sure I understand, your suggestion is that we should use this name ('chainUsdMaxBalance') for the parsed values which do have the 18 decimals?
I used this name here because it is the max balance in USD, as you mentioned.

api/limits.ts
melisaguevara: Yes, it will be undefined when origin chain has no explicit max balance or max deposit. Otherwise, we override it with the min value between those two. It has kind of the same use as 'overrideMaxDepositForOrigin' had in your implementation. Makes sense to you?

melisaguevara:
> Will this override #1111?

Yes

api/limits.ts
james-a-morris: Right but the first call is always undefined so we wouldn't need the ternary

api/limits.ts
```diff
@@ -220,19 +219,73 @@ const handler = async (
...
+      computedOriginChainId,
+      inputToken.symbol,
+      includeDefaultMaxValues
+    );
+    if (chainUsdMaxBalance) {
+      const parsedChainUsdMaxBalance = ethers.utils.parseUnits(
+        chainUsdMaxBalance,
+        inputToken.decimals
```
james-a-morris: RE: https://github.com/across-protocol/frontend/pull/1112/files/d20bd497a63782f4f1d4d3cedd0fb7a9c7f011ab#r1678277764

This should be `18`  if we know that `parsedChainUsdMaxBalance` will be in USD. We would be changing the precision depending on the input token, i.e. 6 for USDC or 8 for WBTC or 18 for ETH

api/limits.ts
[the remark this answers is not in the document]
james-a-morris: Followed up here: https://github.com/across-protocol/frontend/pull/1112/files/d20bd497a63782f4f1d4d3cedd0fb7a9c7f011ab#r1678381912

But, `chainUsdMaxBalance` to me sounds like this is the max balance for this chain in USD. My understanding is that this is the maximum balance for a token on a given chain in USD. I just think the name should include the word Token so it's more representative of what this value will hold.

james-a-morris (commented):
left a few more comments

api/limits.ts
melisaguevara: yes, you're right
marking as resolved as this is no longer used though

api/limits.ts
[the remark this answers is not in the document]
melisaguevara: Got it! Yes, I agree, changed here [3c9d472](https://github.com/across-protocol/frontend/pull/1112/commits/3c9d472bcf42afb7619446e76767810d71031d35)

api/limits.ts
melisaguevara: same as above, changed here [3c9d472](https://github.com/across-protocol/frontend/pull/1112/commits/3c9d472bcf42afb7619446e76767810d71031d35)

api/limits.ts
```diff
@@ -220,21 +219,73 @@ const handler = async (
...
-      minDeposit = minBN(minDeposit, liteChainMaxBoundary);
-      minDepositFloor = minBN(minDepositFloor, liteChainMaxBoundary);
-      maxDepositInstant = minBN(maxDepositInstant, liteChainMaxBoundary);
-      maxDepositShortDelay = minBN(maxDepositShortDelay, liteChainMaxBoundary);
+    if (chainTokenMaxDepositInUsd) {
+      const parsedChainUsdMaxDeposit = parseInInputToken(
+        chainTokenMaxDepositInUsd
+      );
+      chainInputTokenMaxDeposit = parsedChainUsdMaxDeposit
+        .mul(sdk.utils.fixedPointAdjustment)
```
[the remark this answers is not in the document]
james-a-morris: We do use 18 decimals in most cases for USD. cc @nicholaspai @pxrl

api/limits.ts
melisaguevara: Changed here [c2d7030](https://github.com/across-protocol/frontend/pull/1112/commits/c2d7030894c8453fc181afc9d2444efc1717c698)

api/_utils.ts
```diff
@@ -1520,20 +1520,34 @@ export function getLimitsBufferMultiplier(symbol: string) {
...
 
-export function getLiteChainMaxDepositUsd(chainId: number, symbol: string) {
-  const envVarBase = "LITE_CHAIN_USD_MAX_DEPOSIT";
+export function getChainInputTokenMaxDepositInUsd(
+  chainId: number,
+  symbol: string,
+  includeDefault: boolean
+) {
+  const envVarBase = "CHAIN_USD_MAX_DEPOSIT";
+  const defaultValue = includeDefault
```
james-a-morris, line 1545: What is a scenario where we do/do not want a default?

james-a-morris (approved):
lgtm 🫡

PR #1114, opened 2024-07-15
improve(api): remove some unneeded scope

james-a-morris:
We previously allowed a case where we could quote a price with a non-standard base currency. I.e. `acx/wbtc` was possible.

This is out of scope of existing work and no longer needed. Removing will improve readability, clean up legacy code, and make additional changes to this API endpoint more convenient.

### DIFF
--- a/api/coingecko.ts
+++ b/api/coingecko.ts
@@ -11,8 +11,6 @@ import {
 } from "./_utils";
 import {
   SUPPORTED_CG_BASE_CURRENCIES,
-  CHAIN_IDs,
-  TOKEN_SYMBOLS_MAP,
   coinGeckoAssetPlatformLookup,
 } from "./_constants";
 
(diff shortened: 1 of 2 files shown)


amateima (approved):

dohaki (commented):
This looks good to me! Def more readable. I am just curious: Do you remember why we had this additional support for other base currencies? And did you make sure we don't rely on it anymore?

api/coingecko.ts
```diff
@@ -128,14 +46,17 @@ const handler = async (
...
-    else baseCurrency = baseCurrency.toLowerCase();
-
+    // Format the params for consistency
+    baseCurrency = (baseCurrency ?? "eth").toLowerCase();
     l1Token = ethers.utils.getAddress(l1Token);
 
+    // Confirm that the base Currency is supported by Coingecko
+    if (!SUPPORTED_CG_BASE_CURRENCIES.has(baseCurrency)) {
+      throw new InputError(
+        "Base currency supplied is not supported by this endpoint. See documentation."
```
pxrl, line 56: Worth enumerating `SUPPORTED_CG_BASE_CURRENCIES` here? It's nicer from a devx perspective.

api/coingecko.ts
```diff
@@ -157,77 +78,49 @@ const handler = async (
     // We want to compute price and return to caller.
     let price: number;
 
-    const _fixedTokenPrices: {
-      [token: string]: number;
-    } = FIXED_TOKEN_PRICES !== undefined ? JSON.parse(FIXED_TOKEN_PRICES) : {};
-
     // Make sure all keys in `fixedTokenPrices` are in checksum format.
     const fixedTokenPrices = Object.fromEntries(
```
pxrl, line 82: ooc, do we still need this? iirc we only used it when ACX had no price feed, and briefly for ECO/wECO before they deprioritised their integration. It _might_ be another candidate for proactive cleanup.

james-a-morris: Fair call - not sure if we'll ever have this issue again

api/coingecko.ts
james-a-morris: Done https://github.com/across-protocol/frontend/pull/1114/commits/8f77753530aa7485ae3ac10fd117deaeb3e9eec6

api/coingecko.ts
james-a-morris: I will propose this at the next spring planning meeting since we may use this again

yarn.lock
dohaki, line 1: hmm why did this change?

dohaki (approved):
Looks good 👍

yarn.lock
james-a-morris: I'm not sure - will pull this out

api/coingecko.ts
nicholaspai: agreed

nicholaspai (approved):
lgtm good simplification

PR #1115, opened 2024-07-15
feat: weth to eth/weth bridge functionality

bmzig:
We want to be able to support WETH/ETH -> WETH/ETH deposits. Since the contracts will always unwrap to an EOA, we need to use the multicall handler to perform the weth wraps and unwraps. This commit should inject a message into the transaction request which performs these unwraps (or weth transfers) when a deposit condition is set.


src/utils/bridge.ts
```diff
@@ -172,13 +181,30 @@ export async function sendSpokePoolVerifierDepositTx(
...
     isNative,
+    toNative,
     referrer,
     integratorId,
   }: AcrossDepositArgs,
   spokePool: SpokePool,
   spokePoolVerifier: SpokePoolVerifier,
   onNetworkMismatch?: NetworkMismatchHandler
 ): Promise<ethers.providers.TransactionResponse> {
+  /* We cannot send WETH to the recipient on the destination chain with the spoke pool verifier until
```
bmzig, line 192: Marking this to delete in case we want to merge this PR before ETH -> WETH support

src/views/Bridge/utils.ts
```diff
@@ -72,20 +72,12 @@ export function getReceiveTokenSymbol(
...
 
   if (
-    inputTokenSymbol === "ETH" &&
+    (inputTokenSymbol === "ETH" || inputTokenSymbol === "WETH") &&
     (isDestinationChainPolygon || isReceiverContract)
   ) {
     return "WETH";
   }
 
-  if (
```
bmzig, line 81: I'm pretty sure this is OK to remove, since we now want to say that it's possible to bridge from WETH -> ETH/WETH unless the destination is a contract or Polygon, in which case we enforce sending WETH.

scripts/generate-routes.ts
```diff
@@ -272,6 +272,29 @@ function transformChainConfigs(
...
+          return [
+            "WETH",
+            {
+              inputTokenSymbol: "WETH",
+              outputTokenSymbol: "ETH",
+            },
+          ];
+        }
+        /*
+        * TODO: Handle ETH -> ETH/WETH
```
bmzig, line 286: Marking to remove if we merge this before ETH -> WETH/ETH support.

bmzig:
I've so far checked WETH -> WETH transfers from Arbitrum -> All other chains and then Optimism -> Arbitrum.

src/utils/address.ts
```diff
@@ -1,7 +1,8 @@
 import { utils } from "ethers";
 
-import { ChainId } from "./constants";
+import { ChainId, TOKEN_SYMBOLS_MAP } from "./constants";
 import { getProvider } from "./providers";
+import { getDeployedAddress } from "@across-protocol/contracts";
```
dohaki, line 5: This line unfortunately increases the FE bundle size by ~30%... This is due to the `@across-protocol/contracts` package being non-tree shakeable which is a known issue for some time. Can we for now just hardcode the `MulticallHandler` addresses in https://github.com/across-protocol/frontend/pull/1115/files#diff-ca62a13994aae3c545ad2d796d272f3ee8c0ab33b83bb3d5c023dc553b55f984R29-R36?

I think we'd need to populate the addresses into the `routes*.json` file during the build time ideally. But we can tackle that in a separate PR

src/utils/bridge.ts
```diff
@@ -233,6 +260,14 @@ export async function sendDepositV3Tx(
   fillDeadline ??=
     getCurrentTime() - 60 + (await spokePool.fillDeadlineBuffer());
 
+  // If a user is interacting with the front end, assume that they are an EOA and not
+  // a contract. Since they are an EOA, by default they will receive ETH on the destination
+  // (except for Polygon). Therefore, we only need to use the multicall handler when the EOA
+  // wants to receive WETH after an ETH or WETH deposit.
```
dohaki, line 266: I think we need to actually check whether the `recipient` is an EOA using `isContractDeployedToAddress`. And only if it is not a contracts, use the multicall handler?

src/views/Bridge/utils.ts
dohaki: Yes, that should be fine

dohaki (commented):
This looks very good! I left some smaller remarks though. And another issue we should fix is captured in the screenshot below:

image

The tooltip is technically not wrong but redundant. We should remove it by adding the condition 
```
inputToken.symbol === "WETH"
```
to https://github.com/across-protocol/frontend/blob/11e87ea37640216bae64103ecd707800b53342ab/src/views/Bridge/components/EstimatedTable.tsx#L428

src/utils/address.ts
bmzig: Sounds good. I removed this here https://github.com/across-protocol/frontend/pull/1115/commits/4b70c46c3e8e9ae835ef10656373bfb5690e4477 and just hardcoded in the addresses.

bmzig:
Closing to revisit with an updated PR

PR #1117, opened 2024-07-16
improve(api): remove fixed pricing needs

james-a-morris:
This was prompted in https://github.com/across-protocol/frontend/pull/1114#discussion_r1679488408 - good callout by @pxrl and @nicholaspai


dohaki (approved):
Nice 👍

PR #1119, opened 2024-07-16
hotfix(coingecko): account for derived currencies with strict scope

james-a-morris:
Missed reference to matic in suggested-fees. This enabled a _select_ number of derived tokens.

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -160,7 +160,7 @@ export const BLOCK_TAG_LAG = -1;
 // we've decided to keep this list small for now.
 export const SUPPORTED_CG_BASE_CURRENCIES = new Set(["eth", "usd"]);
 // Note: this is a small set of currencies that the API will derive from the base currencies by using USD as an intermediary.
-export const SUPPORTED_CG_DERRIVED_CURRENCIES = new Set(["matic"]);
+export const SUPPORTED_CG_DERIVED_CURRENCIES = new Set(["matic"]);
 
 // 1:1 because we don't need to handle underlying tokens on FE
 export const EXTERNAL_POOL_TOKEN_EXCHANGE_RATE = utils.fixedPointAdjustment;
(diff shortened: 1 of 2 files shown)


api/_constants.ts
```diff
@@ -159,6 +159,8 @@ export const BLOCK_TAG_LAG = -1;
 // Note: this is a small subset of all the supported base currencies, but since we don't expect to use the others,
 // we've decided to keep this list small for now.
 export const SUPPORTED_CG_BASE_CURRENCIES = new Set(["eth", "usd"]);
+// Note: this is a small set of currencies that the API will derive from the base currencies by using USD as an intermediary.
+export const SUPPORTED_CG_DERRIVED_CURRENCIES = new Set(["matic"]);
```
nicholaspai, line 163: typo, 
```suggestion
export const SUPPORTED_CG_DERIVED_CURRENCIES = new Set(["matic"]);
```

api/coingecko.ts
```diff
@@ -114,17 +123,32 @@ const handler = async (
...
+        TOKEN_SYMBOLS_MAP[
+          baseCurrency.toUpperCase() as keyof typeof TOKEN_SYMBOLS_MAP
+        ];
+      quotePrice = await getCachedTokenPrice(
+        token.addresses[CHAIN_IDs.MAINNET],
+        "usd"
+      );
+      quotePrecision = token.decimals;
+    }
+    price = Number((price / quotePrice).toFixed(quotePrecision));
```
pxrl, line 150: Is the final `Number()` conversion necessary? The division and subsequent use of `.toFixed()` should already imply that it's a number.

james-a-morris: The only thing about this is that the API on prod returns a number but removing the `Number` would would mean the return type is a string.

pxrl: ty - I missed that `Number.toFixed()` returns a string, but of course that makes sense.

pxrl (approved):
This tests OK for me :+1:

PR #1121, opened 2024-07-18
chore: update fill times

dohaki:
Closes ACX-2350

### DIFF
--- a/src/utils/fill-times.ts
+++ b/src/utils/fill-times.ts
@@ -1,17 +1,7 @@
-import { ChainId } from "./constants";
+import { getChainInfo } from "./constants";
 
 // Raw json from https://3251bbf8.us2a.app.preset.io/superset/explore/p/ZxKyKB6om5P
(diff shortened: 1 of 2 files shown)


src/utils/fill-times.ts
```diff
@@ -39,8 +29,8 @@ export function getFastFillTimeByRoute(
   toChainId: number,
   tokenSymbol: string
 ) {
-  const fromChainName = chainIdToPresetChainName[fromChainId];
-  const toChainName = chainIdToPresetChainName[toChainId];
+  const fromChainName = getChainInfo(fromChainId).name?.toLowerCase();
+  const toChainName = getChainInfo(toChainId).name?.toLowerCase();
```
pxrl, line 33: Can `getChainInfo()` ever return `undefined`? It looks like if the chain is unrecognised then it just returns some dummy values instead.

pxrl (approved):

PR #1122, opened 2024-07-18
improve(rewards): cap estimation at 25bips of input

james-a-morris:
Rewards are now capped to `min(25bps, bridge_fee) * 95%`. UI has had updates to reference this information.


dohaki (approved):

melisaguevara (approved):
lgtm!

PR #1124, opened 2024-07-24
feat(api): get timing estimates in suggested-fees

james-a-morris:


api/_timings.ts
```diff
@@ -0,0 +1,62 @@
...
+  const sourceData = timingsLookup[sourceChainId] ?? timingsLookup["OTHER"];
+  const destinationData = sourceData[destinationChainId] ?? sourceData["OTHER"];
+  const symbolData = destinationData[symbol] ?? destinationData["OTHER"];
+  const sortedCutoffs = symbolData.sort(
+    ({ amountUsd: amountUsdA }, { amountUsd: amountUsdB }) =>
+      amountUsdA.lt(amountUsdB) ? -1 : amountUsdA.gt(amountUsdB) ? 1 : 0
+  );
+  return (
+    sortedCutoffs.find((cutoff) => usdAmount.lt(cutoff.amountUsd))
+      ?.timingInSecs ?? 10
```
james-a-morris, line 60: This 10 seconds is a magic number - it's not intended to be hit and exists to appease `tsc`. If we wanted to ensure that there was no fallback, I can add a `!`

api/suggested-fees.ts
```diff
@@ -214,10 +216,12 @@ const handler = async (
         [liteChainIdsEncoded],
       ],
       tokenPrice,
+      tokenPriceUsd,
       limits,
     ] = await Promise.all([
       callViaMulticall3(provider, multiCalls, { blockTag: quoteBlockNumber }),
       getCachedTokenPrice(l1Token.address, baseCurrency),
+      getCachedTokenPrice(l1Token.address, "usd"),
```
james-a-morris, line 224: We need to add an addtl call to the hotpath, unfortunately

api/_timings.ts
james-a-morris, line 1: This lookup is big - but we're going to rely on the warm starts of Vercel/AWS to cache this data in the global scope of the handler.

api/_timings.ts
```diff
@@ -0,0 +1,62 @@
...
+export function resolveTiming(
+  sourceChainId: string,
+  destinationChainId: string,
+  symbol: string,
+  usdAmount: BigNumber
+): number {
+  const sourceData = timingsLookup[sourceChainId] ?? timingsLookup["0"];
+  const destinationData = sourceData[destinationChainId] ?? sourceData["0"];
+  const symbolData = destinationData[symbol] ?? destinationData["OTHER"];
+  const sortedCutoffs = symbolData.sort(
```
mrice32, line 54: Instead of sorting in the hot path, could we just pre-sort, instead?

I don't think this would be any harder than just sorting `timings` first thing before processing, since that will guarantee that the elements get inserted into the data structure in that order (asc or desc), right?

Side note: sort (weirdly) modifies the array in place, so we're technically modifying the array each time we hit this line. Not a bug or anything, but feels a bit odd since this data is intended to be static.

mrice32 (commented):
Only minor comments!

api/_timings.ts
james-a-morris: Oh shoot this was meant to be sorted 😅

api/_timings.ts
```diff
@@ -0,0 +1,62 @@
...
+          [symbol: string]: {
+            amountUsd: BigNumber;
+            timingInSecs: number;
+          }[];
+        };
+      };
+    }
+  );
+
+export function resolveTiming(
```
dohaki, line 45: We previously had a way to override the fill time for specific token. We used that for example for SNX a few times when the SNX relayer had enough balance for fast fills but didn't update their relayer so they couldn't run the bot. So we temp overrode the times that would be displayed.

Do we have this functionality somewhere in this PR?

src/utils/bridge.ts
```diff
@@ -104,73 +106,19 @@ export type ConfirmationDepositTimeType = {
...
-        highEstimate + depositDelay
-      } minutes`,
-      lowEstimate: lowEstimate + depositDelay,
-      highEstimate: highEstimate + depositDelay,
-    };
+  const timeToFill =
+    (estimatedFillTimeSec ?? 900) + // 15 minutes if not provided
+    depositDelay;
+  return {
+    formattedString: `~${timeToFill} seconds`,
```
dohaki, line 118: I think we should parse this to respective minutes if `timeToFill >= 60`

dohaki: And also use `Math.floor` and use the correct word `second` or `minute` in singular for values of `1`

dohaki (commented):
Nice work! Left a few remarks

api/_timings.ts
james-a-morris: We've reasoned about this but ultimately decided against it because AWS caches the imports

src/utils/bridge.ts
james-a-morris: https://github.com/across-protocol/frontend/pull/1124/commits/5ffe3a9aa8f192b75ee72a72db528d13e190b08a

src/data/fill-times-preset.json
```diff
@@ -1,1097 +1,702 @@
...
-        "origin_symbol",
-        "fill_time_seconds"
-      ],
-      "coltypes": [1, 1, 1, 0],
-      "rowcount": 180,
-      "sql_rowcount": 180
-    }
-  ]
-}
+[
```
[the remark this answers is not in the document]
james-a-morris: We won't need to push this after - the `yarn build` will implicitly bring in the updated value in Vercel & if we add this file to the `.gitignore` it can be used as a placeholder/set of fallback values

api/_timings.ts
james-a-morris: https://github.com/across-protocol/frontend/pull/1124/commits/72fba8728fa619105040ba35370a186d16c9e12b

api/_timings.ts
dohaki: We can also add this in a subsequent PR if you want to get this in ASAP

dohaki (approved):
Approving for now. The override option as mentioned [here](https://github.com/across-protocol/frontend/pull/1124#discussion_r1690918613), we can add later

api/_timings.ts
nicholaspai: Does that functionality exist currently and can it modify the dictionary after the above steps?

nicholaspai (approved):

api/_timings.ts
james-a-morris: @nicholaspai This is a hard-coded lookup—it's possible for it to exist in the new system. We may want to encode it in JSON, though, or even in an ENV. However, for now, the change I made will preserve this override functionality.

src/data/fill-times-preset.json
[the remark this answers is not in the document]
mrice32: How will we update the fallback in the future if it's in gitignore? Note: I know there are ways around gitignore, but it makes it a bit trickier, no?

mrice32 (approved):
Only one comment: https://github.com/across-protocol/frontend/pull/1124#discussion_r1691993613

PR #1126, opened 2024-07-25
fix: remove special route rate model key for lite chain

dohaki:


amateima (commented):
Why is this removed?

dohaki:
> Why is this removed?

@pxrl can you provide a bit more context?

pxrl:
> Why is this removed?

The rate model is setup such that deposits should be quoted for 1135 -> 1, not 1135 -> 1135. ATM the FE is quoting 1135 -> 1135, so the depositor is under-paying and the relayer is being shortchanged.

We can't update the rate model such that 1135 -> 1135 has a higher fee because then the relayer actually has to pay that fee, so it ends up earning 0. Instead, the FE should quote assuming that the relayer pays an LP fee for taking repayment on the destination chain, and then the relayer earns the LP fee for taking repayment on origin. This is actually what is happening, since the relayer is effectively doubling as an LP when it fills deposits out of lite chains.

pxrl (approved):

nicholaspai (approved):

melisaguevara (commented):
LGTM
We can delete the call to get lite chain ids then and that would fix the linting issue too

pxrl:
> LGTM We can delete the call to get lite chain ids then and that would fix the linting issue too

Good call! Removed: https://github.com/across-protocol/frontend/pull/1126/commits/17aec009e650a391dffc578538a00af32db14874 https://github.com/across-protocol/frontend/pull/1126/commits/bdf482cd9c3fed2c676440d90a7b6b185280b3ee

PR #1127, opened 2024-07-26
docs: fix try-it section default values

dohaki:
There seems to have been an update in [Scalar](https://github.com/scalar/scalar) (the API client used when clicking `try it` in GitBook) which broke the default values for query params. This PR tries to fix it.


gsteenkamp89 (approved):
Looks like it's just formatting changes?

dohaki:
> Looks like it's just formatting changes?

Yep, exactly

PR #1132, opened 2024-07-26
Fixing timing estimations for Lisk

melisaguevara:
/suggested-fees was breaking as there is no timing estimation for routes involving Lisk.
This is a temporarily fix, will check with @james-a-morris how this should be solved next week.

### DIFF
--- a/api/_timings.ts
+++ b/api/_timings.ts
@@ -69,11 +69,15 @@ export function resolveTiming(
     return override;
   }
 
+  const DEFAULT_FILL_TIME_SECS = 10;
   const sourceData = timingsLookup[sourceChainId] ?? timingsLookup["0"];
-  const destinationData = sourceData[destinationChainId] ?? sourceData["0"];
-  const symbolData = destinationData[symbol] ?? destinationData["OTHER"]; // implicitly sorted
-  return (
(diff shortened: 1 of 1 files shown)


api/_timings.ts
```diff
@@ -69,11 +69,15 @@ export function resolveTiming(
...
 
+  const DEFAULT_FILL_TIME_SECS = 10;
   const sourceData = timingsLookup[sourceChainId] ?? timingsLookup["0"];
-  const destinationData = sourceData[destinationChainId] ?? sourceData["0"];
-  const symbolData = destinationData[symbol] ?? destinationData["OTHER"]; // implicitly sorted
-  return (
-    symbolData.find((cutoff) => usdAmount.lt(cutoff.amountUsd))?.timingInSecs ??
-    10
-  );
+  if (sourceData) {
```
james-a-morris, line 74: This could probably be resolve if we had above:

```
  const sourceData = timingsLookup[sourceChainId] ?? timingsLookup["0"] ?? {
      "0": {
           "OTHER": [{amountUsd: {{MAX_INT}}, timingInSeconds: {{default}}}]
      }
  }

```

james-a-morris (approved):
Good patch for now

amateima (commented):
Is this solution preferred instead of editing the `fill-times-preset.json`?

PR #1135, opened 2024-07-26
fix: update UI message when maxDeposit is 0

melisaguevara:
We were displaying the message "Input amount exceeds limits set to maintain optimal service for all users. Decrease amount to 0 LSK or lower." when maxDeposit was 0.

### DIFF
--- a/src/views/Bridge/components/BridgeForm.tsx
+++ b/src/views/Bridge/components/BridgeForm.tsx
@@ -72,6 +72,8 @@ export type BridgeFormProps = {
 const validationErrorTextMap = {
   [AmountInputError.INSUFFICIENT_BALANCE]:
     "Insufficient balance to process this transfer.",
+  [AmountInputError.PAUSED_DEPOSITS]:
+    "[INPUT_TOKEN] deposits are temporarily paused.",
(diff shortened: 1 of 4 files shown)


dohaki (approved):
lgtm

src/views/Bridge/components/BridgeForm.tsx
```diff
@@ -174,13 +176,15 @@ const BridgeForm = ({
...
-          )}
+          errorText={validationErrorTextMap[validationError]
+            .replace("[INPUT_TOKEN]", selectedRoute.fromTokenSymbol)
+            .replace(
+              "[MAX_DEPOSIT]",
+              `${formatUnitsWithMaxFractions(
+                limits?.maxDeposit || 0,
+                getToken(selectedRoute.fromTokenSymbol).decimals
+              )} ${selectedRoute.fromTokenSymbol}`
+            )}
```
james-a-morris, line 187: This may be worthy of adding to a formatting function given that it's getting more complex

melisaguevara: Yes, that makes sense, I will include this in a separate PR just to avoid showing the wrong message in case we hit the deposit limit again.

PR #1136, opened 2024-07-29
fix: don't claim rewards for windows not live

amateima:
Currently the `Claim` button is enabled because of proofs that corresponds to windows that were not created yet. 

This PR will allow to claim rewards only for the proofs that are associated with windows that were created in the MerkleDistributor contract


dohaki (approved):
Nice 👍

PR #1137, opened 2024-07-29
refactor: use constant retry delay for deposit status tracking

dohaki:
Fixes ACX-2431

Improves UX for fills that take a bit longer on the deposit status screen.


gsteenkamp89 (approved):
Looks good 👍

PR #1146, opened 2024-08-06
feat: add zora

gsteenkamp89:


api/_constants.ts
```diff
@@ -103,6 +103,9 @@ const defaultRelayerFeeCapitalCostConfig: {
 defaultRelayerFeeCapitalCostConfig["USDC.e"] = {
   ...defaultRelayerFeeCapitalCostConfig["USDC"],
 };
+defaultRelayerFeeCapitalCostConfig["USDzC"] = {
+  ...defaultRelayerFeeCapitalCostConfig["USDC"],
+};
```
pxrl, line 108: This change worked to permit USDzC to work...but I don't understand why USDbC works without this. Any clues?

scripts/generate-routes.ts
```diff
@@ -226,6 +231,27 @@ function transformChainConfigs(
...
+                outputTokenSymbol: "USDbC",
+              },
+            ];
+          } else if (toChainConfig.tokens.includes("USDzC")) {
+            return [
+              {
+                inputTokenSymbol: tokenSymbol,
+                outputTokenSymbol: "USDzC",
+              },
+            ];
```
pxrl, line 254: This was added to support bridged USDC -> USDzC on Zora, but the code is very fragile and should be refactored to handle this without a lot of manual rules being added.

pxrl (approved):
To be merged with some residual todos RE bridged USDC -> USDzC routes, Zora logo on fill event, USDzC capital config.

api/_constants.ts
pxrl: nb. USDbC isn't required because it's not a supported inputToken - it is swapped to native USDC on Base first. USDzC is however a supported input token so it requires an explicit config.

PR #1149, opened 2024-08-09
fetch batch balances

gsteenkamp89:
closes ACX-2513

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,7 @@ Adjust values in the created `.env` accordingly.
 Start the frontend with a dev server by running:
 
 ```bash
-PORT=3000 yarn dev
-```
-
-Serve serverless functions with a dev server by running:
-
-```bash
-PORT=3001 yarn dev:api
+PORT=3000 vercel dev
 ```
 
 ### Build production bundle
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@
 
 Web UI for Across.
 
-- App: https://app.across.to
-- Docs: https://docs.across.to
-- Medium: https://medium.com/across-protocol
+- App: <https://app.across.to>
+- Docs: <https://docs.across.to>
+- Medium: <https://medium.com/across-protocol>
 
 ## Development
 
@@ -16,7 +16,7 @@ Web UI for Across.
 
 - [yarn v1.22.22](https://classic.yarnpkg.com/en/docs/install)
 - NodeJS >=v18
-- [Vercel CLI](https://vercel.com/docs/cli)
+- [Vercel CLI (35 or higher)](https://vercel.com/docs/cli)
 
 ### Local setup
 
@@ -35,6 +35,12 @@ Start the frontend with a dev server by running:
 
 ```bash
 PORT=3000 yarn dev
+```
+
+Serve serverless functions with a dev server by running:
+
+```bash
+PORT=3001 yarn dev:api
 ```
 
 ### Build production bundle
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -778,7 +778,7 @@ export const getBatchBalanceViaMulticall3 = async (
   chainId: string | number,
   addresses: string[],
   tokenAddresses: string[],
-  blockTag?: providers.BlockTag
+  blockTag: providers.BlockTag = "latest"
 ): Promise<{
   blockNumber: providers.BlockTag;
   balances: Record<string, Record<string, string>>;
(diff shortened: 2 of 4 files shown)


api/batch-account-balance.ts
```diff
@@ -0,0 +1,100 @@
...
+import {
+  getBatchBalanceViaMulticall3,
+  getLogger,
+  getProvider,
+  handleErrorCondition,
+  validAddress,
+} from "./_utils";
+
+const BatchAccountBalanceQueryParamsSchema = type({
+  tokenAddress: union([literal("native"), validAddress()]),
```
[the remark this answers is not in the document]
dohaki: hmm yea maybe we could use the zero address here? 🤔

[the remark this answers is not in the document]
gsteenkamp89: i'm not 100% sure native tokens are ALWAYS represented by a zero address, though I could be wrong

README.md
```diff
@@ -37,6 +37,12 @@ Start the frontend with a dev server by running:
 PORT=3000 yarn dev
 ```
 
+Serve serverless functions with a dev server by running:
+
+```bash
+PORT=3001 yarn dev:api
```
james-a-morris, line 43: This will also serve the standard website too - not just the API

api/batch-account-balance.ts
```diff
@@ -0,0 +1,105 @@
...
+    // Log the response
+    logger.debug({
+      at: "BatchAccountBalance",
+      message: "Response data",
+      responseJson: balances,
+    });
+    // Set the caching headers that will be used by the CDN.
+    response.setHeader(
+      "Cache-Control",
+      "s-maxage=150, stale-while-revalidate=150"
```
james-a-morris, line 91: Are we caching for 5 min?

api/batch-account-balance.ts
```diff
@@ -0,0 +1,105 @@
...
+    const provider = getProvider(chainIdAsInt);
+
+    if (tokenAddress === sdk.constants.ZERO_ADDRESS) {
+      // fetch native balances
+      const res = await Promise.all(
+        addresses.map((a) => provider.getBalance(a, "latest"))
+      );
+      addresses.forEach(
+        (address, i) => (balances[address] = res[i].toString())
+      );
```
pxrl, line 69: wdyt about using `Multicall3.getEthBalance()` for this? We could then wrap that in `Multicall3.aggregate()` and reduce this down to a single RPC call.

(I am also wondering if we could accept an array of `tokenAddress` on input, in which case we could query (for example) WETH + native balance in a single balance query)

api/batch-account-balance.ts
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      // fetch erc20 balances
+      const res = await getBatchBalanceViaMulticall3(
```
pxrl, line 72: I think we have at least one chain where there's no Multicall3 deployment (Redstone). We can either deploy it ourselves (and have a way of specifying the bespoke address), implement fallback behaviour to query the RPC in parallel, or just throw on an unsupported chain (we will do this implicitly when the RPC call fails anyway).

The chains without Multicall3 are very mostly long-tail and I can't imagine that we'd actually want to be applying exclusivity on those chains, so it _might_ be OK to just ignore the problem for the time being. It also seems like Multicall3 is fairly quick to be deployed to new chains - we previously lacked Lisk but it was deployed there 2 weeks ago: https://blockscout.lisk.com/address/0xcA11bde05977b3631167028862bE2a173976CA11?tab=contract). This would change if we start building dynamic balance checking into the standard suggested-fees workflow though - in that case we'd need something more reliable.

pxrl (commented):
Looks good! I left a couple of high level questions - lmk.

README.md
gsteenkamp89: Ah i see so `vercel dev` serves the FE **and** the api from the same port?
Would it be better then to just alter the existing dev script to use `vercel dev` instead of `vite` (which I assume is only serving the FE)? 

`"dev":  "export REACT_APP_GIT_COMMIT_HASH=$(git rev-parse HEAD) && vercel dev --debug --listen $PORT"`

api/batch-account-balance.ts
gsteenkamp89: This is just boilerplate at this point, how long should we cache it for do you think?

api/_utils.ts
```diff
@@ -765,6 +765,32 @@ export const getBalance = (
...
+ * @param tokenAddress The valid ERC20 token address on the given `chainId`.
+ * @param blockTag Block to query from, defaults to latest block
+ * @returns a Promise that resolves to an array of BigNumbers
+ */
+export const getBatchBalanceViaMulticall3 = async (
+  chainId: string | number,
+  addresses: string[],
+  tokenAddress: string,
+  blockTag: providers.BlockTag = "latest"
+): Promise<ethers.utils.Result[]> => {
```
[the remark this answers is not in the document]
gsteenkamp89: great suggestion!

api/batch-account-balance.ts
gsteenkamp89: both very doable 👍

README.md
gsteenkamp89: Oh I see now how this works. I think we can remove this line and instead alter the original instruction "Start the frontend with a dev server by running:" to `PORT=3000 vercel dev` ?

I don't see why we would need to run the FE without the API, but I could be wrong.

api/batch-account-balance.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  try {
+    // Validate the query parameters
+    assert(query, BatchAccountBalanceQueryParamsSchema);
+    // Deconstruct the query parameters
+    let { tokenAddresses, addresses, chainId } = query;
+
+    console.log("Fetching balances for config on chainId: ", chainId);
+    console.log("TokenAddresses: ", tokenAddresses);
+    console.log("Accounts ", addresses);
```
pxrl, line 77: Is this just for debug?

pxrl (approved):
LGTM! Very clean.

PR #1150, opened 2024-08-09
feat(analytics): track custom domain identifier

james-a-morris:
This PR allows custom tracking of domain ids when a user includes the `?integrator_id` query param.

### DIFF
--- a/src/hooks/useReferrer.ts
+++ b/src/hooks/useReferrer.ts
@@ -1,12 +1,12 @@
-import { useState, useEffect } from "react";
-import { ethers } from "ethers";
+import { useState, useEffect, useMemo } from "react";
+import { ethers, utils } from "ethers";
 import { useQueryParams } from "./useQueryParams";
 import { useConnection } from "hooks";
 import { isDefined } from "utils";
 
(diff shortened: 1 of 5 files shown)


src/hooks/useReferrer.ts
```diff
@@ -46,8 +46,24 @@ export default function useReferrer() {
       setIsResolved(true);
     }
   }, [provider, r]);
+
+  // Return the integratorId as a hex string with a 0x prefix if not already present.
+  // Default to Across if no integrator_id is provided or if the integrator_id is invalid.
+  const integratorId = useMemo(() => {
```
gsteenkamp89, line 52: Does this need to be memoized?

gsteenkamp89 (approved):

src/hooks/useReferrer.ts
dohaki: Yea, I don't think we need to

james-a-morris: resolved. Will merge

james-a-morris:
Merging without the memoization

PR #1152, opened 2024-08-13
improve: capital cost config

james-a-morris:


api/_constants.ts
```diff
@@ -34,19 +34,19 @@ const defaultRelayerFeeCapitalCostConfig: {
...
+    upperBound: ethers.utils.parseUnits("0.0001").toString(),
     cutoff: ethers.utils.parseUnits("1500000").toString(),
     decimals: 18,
   },
   USDC: {
     lowerBound: ethers.utils.parseUnits("0.0001").toString(),
-    upperBound: ethers.utils.parseUnits("0.0004").toString(),
-    cutoff: ethers.utils.parseUnits("1500000").toString(),
+    upperBound: ethers.utils.parseUnits("0").toString(),
+    cutoff: ethers.utils.parseUnits("100000").toString(),
```
melisaguevara, line 44: Should we also set the lowerBound to 0?
I'm asking because of [this assertion](https://github.com/across-protocol/sdk/blob/master/src/relayFeeCalculator/relayFeeCalculator.ts#L199) I see in the SDK

melisaguevara (commented):
One question

api/_constants.ts
```diff
@@ -16,14 +16,14 @@ const defaultRelayerFeeCapitalCostConfig: {
 } = {
   ETH: {
     lowerBound: ethers.utils.parseUnits("0.0001").toString(),
-    upperBound: ethers.utils.parseUnits("0.0004").toString(),
-    cutoff: ethers.utils.parseUnits("750").toString(),
+    upperBound: ethers.utils.parseUnits("0.000075").toString(),
```
melisaguevara, line 19: Same here, should we set lowerBound so it is lower than upperBound?

api/_constants.ts
```diff
@@ -16,14 +16,14 @@ const defaultRelayerFeeCapitalCostConfig: {
...
+    cutoff: ethers.utils.parseUnits("0.3").toString(),
     decimals: 18,
   },
   WETH: {
     lowerBound: ethers.utils.parseUnits("0.0001").toString(),
-    upperBound: ethers.utils.parseUnits("0.0004").toString(),
-    cutoff: ethers.utils.parseUnits("750").toString(),
+    upperBound: ethers.utils.parseUnits("0.000075").toString(),
+    cutoff: ethers.utils.parseUnits("0.3").toString(),
```
melisaguevara, line 26: And also in here.

api/_constants.ts
james-a-morris: It's meant to be inverted

api/_constants.ts
james-a-morris: It's meant to be inverted

api/_constants.ts
james-a-morris: We're inverting - this is a valid case - thanks for calling this out

pxrl (approved):

PR #1154, opened 2024-08-15
Initial framework for relayer exclusivity

pxrl:
TODO:
- Read global config vars.
- Testing & hardening.


api/exclusivity/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  const relayers = getEligibleRelayers(
+    destinationChainId,
+    outputToken,
+    outputAmount,
+    relayerFeePct
+  );
+  const exclusiveRelayer = strategy(relayers);
+  return { exclusiveRelayer, exclusivityPeriod };
```
pxrl, line 38: This should be re-ordered, such that we first resolve the exclusive relayer address, and then set the exclusivityPeriod subject to whether exclusiveRelayer is ZERO_ADDRESS.

pxrl: https://github.com/across-protocol/frontend/pull/1154/commits/89a658ace86bd7d42142c3e279fd9db200291b1e

api/exclusivity/config.ts
```diff
@@ -0,0 +1,80 @@
...
+  [CHAIN_IDs.MAINNET]: 8,
+  [CHAIN_IDs.LINEA]: 1.5,
+  [CHAIN_IDs.SCROLL]: 1.5,
+};
+
+/**
+ * Object mapping named exclusivity strategies to their implementation.
+ */
+const EXCLUSIVITY_STRATEGIES = {
+  none: (_: string[]) => ZERO_ADDRESS,
```
pxrl, line 30: The strategy function takes an array of addresses at the moment, but it will ultimately need to be an array of `RelayerConfig`, where each element includes things like:
- address
- min profitability threshold
- min exclusivity threshold
- max fill amount (usd)

api/exclusivity/config.ts
```diff
@@ -0,0 +1,80 @@
+import * as sdk from "@across-protocol/sdk";
+import { RelayerSelector } from "./types";
+
+const { CHAIN_IDs, ZERO_ADDRESS } = sdk.constants;
+
+// Absolute minimum exclusivity to set for any given transfer.
+const DEFAULT_MIN_EXCLUSIVITY = 3;
```
nicholaspai, line 7: nit: make it clear this is in seconds, right?

api/exclusivity/config.ts
pxrl: Done: https://github.com/across-protocol/frontend/pull/1154/commits/199baacf360aacf7f7cee6f00310c8fffebfd1fe

api/exclusivity/config.ts
dohaki, line 1: We should rename the folder to `_exclusivity` because Vercel Serverless Functions will consider everything that is not prefixed with `_` a valid route. E.g. https://app-frontend-v3-git-pxrl-exclusive-relayer-uma.vercel.app/api/exclusivity/strategies/random-weighted

pxrl: Ah - good point. Relocated: https://github.com/across-protocol/frontend/pull/1154/commits/f631c842be24652acc61c4ab50dafe9ba8f987c4

api/_utils.ts
```diff
@@ -881,6 +881,33 @@ export const getCachedTokenBalance = async (
...
+  const response = await axios.get<BatchAccountBalanceResponse>(
+    `${resolveVercelEndpoint()}/api/batch-account-balance`,
+    {
+      params: {
+        chainId,
+        addresses,
+        tokenAddresses,
+      },
+    }
+  );
```
pxrl, line 906: The API doesn't like receiving this query:
```
api/batch-account-balance?chainId=10&addresses[]=0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D&tokenAddresses[]=0x0000000000000000000000000000000000000000&tokenAddresses[]=0x94b008aA00579c1307B0EF2c499aD98a8ce58e58
```

api/_abis.ts
pxrl, line 1: @gsteenkamp89 If we get some of your improvements in then we should be OK to back this change out.

api/suggested-fees.ts
```diff
@@ -260,13 +264,27 @@ const handler = async (
...
+        destinationChainId,
+        outputToken.address,
+        amount.sub(totalRelayFee),
+        BigNumber.from(relayerFeeDetails.relayFeePercent) // @todo: Subtract destination gas cost.
+      );
+
+    // @todo: This assumes an instant deposit, with 0 user delay on approval & submission.
+    // This is unrealistic and must be padded for consumers _other than_ the Across FE.
+    const exclusivityDeadline =
+      exclusivityPeriod > 0 ? quoteTimestamp + exclusivityPeriod : 0;
```
pxrl, line 286: This doesn't work when the quoteTimestamp is set in the past.

api/_exclusivity/index.ts
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+      if (relayerFeePct.lt(parseUnits(String(config.minProfitThreshold)))) {
+        return false;
+      }
+
+      if (ethers.BigNumber.from(balance[ZERO_ADDRESS]).lt(minNativeBalance)) {
+        return false;
+      }
+
```
pxrl, line 100: Missing: max fill amounts are not considered here. This requires that we fetch a price for the output token to figure out the fill amount. An input token price is technically also needed but we can probably just assume 1:1 value for now.

api/_utils.ts
pxrl: Fixed.

api/_exclusivity/config.ts
```diff
@@ -0,0 +1,108 @@
...
+  randomWeighted,
+} as const;
+
+/**
+ * Select a specific relayer exclusivity strategy to apply.
+ * This currently hardcodes the "none" strategy, but will be updated to support additional strategies
+ * and selection from on env-based configuration.
+ * @returns A handler function that will narrow an array of relayers to a single address.
+ */
+export function getStrategy(): RelayerSelector {
```
pxrl, line 46: This function probably needs to accept route + token information to determine the correct strategy to apply based on config.

```suggestion
export function getStrategy(): RelayerSelector {
```

api/_exclusivity/config.ts
```diff
@@ -0,0 +1,108 @@
...
+export function getExclusivityPeriod(
+  originChainId: number,
+  destinationChainId: number
+): number {
+  return (
+    getOriginChainDelay(originChainId) +
+    getDestinationChainDelay(destinationChainId) +
+    DEFAULT_MIN_EXCLUSIVITY
+  );
+}
```
pxrl, line 87: This should be updated to prefer reading the fill-times JSON and should select the correct route as per the FE estimates currently do.

api/_exclusivity/config.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Read the configured confirmation delay for a given destination chain.
+ * @param destinationChainId Chain ID for the destination chain.
+ * @returns The number of seconds to wait for destination chain confirmation.
+ */
+function getDestinationChainDelay(originChainId: number): number {
+  return (
+    DESTINATION_CONFIRMATION_DELAY[originChainId] ??
+    DEFAULT_DESTINATION_CONFIRMATION_DELAY
+  );
```
pxrl, line 107: To be deleted; may be restored if we find a use case for applying overrides.

api/_exclusivity/index.ts
pxrl: Token information (decimals) is probably also needed here in order to resolve the price correctly.

src/utils/bridge.ts
```diff
@@ -233,10 +239,15 @@ export async function sendDepositV3Tx(
...
-  );
+  const useExclusiveRelayer =
+    !(
+      exclusiveRelayer === ethers.constants.AddressZero &&
+      exclusivityDeadline === 0
+    ) && [690, 1135, 81457, 534352].includes(fromChain); // @todo: Upgrade SpokePools.
+
+  exclusivityDeadline = useExclusiveRelayer
+    ? Math.max(5, exclusivityDeadline - getCurrentTime())
+    : Math.max(getCurrentTime() + 30, exclusivityDeadline);
```
pxrl, line 250: This logic is faulty because the value of exclusivityDeadline might be stale (a response served out of cache). The suggested-fees response probably needs to include an `exclusivityPeriod` or `exclusivityWindow` field that specifies the offset. The name should be ultimately aligned with https://github.com/across-protocol/contracts/pull/584#discussion_r1725763199.

src/utils/bridge.ts
```diff
@@ -233,10 +239,15 @@ export async function sendDepositV3Tx(
...
 
-  const useExclusiveRelayer = !(
-    exclusiveRelayer === ethers.constants.AddressZero &&
-    exclusivityDeadline === 0
-  );
+  const useExclusiveRelayer =
+    !(
+      exclusiveRelayer === ethers.constants.AddressZero &&
+      exclusivityDeadline === 0
+    ) && [690, 1135, 81457, 534352].includes(fromChain); // @todo: Upgrade SpokePools.
```
pxrl, line 246: Temporarily needed because only newer SpokePool deployments support `depositExclusive()`.

src/data/examples/exclusive-relayers.json
pxrl, line 1: This file can possibly be deleted.

api/_exclusivity/config.ts
pxrl: @gsteenkamp89

src/utils/bridge.ts
pxrl: @dohaki

PR #1155, opened 2024-08-16
improve: consolidate ENV vars

james-a-morris:
This consolidates the ENV vars related to relayer quoting override to a single ENV with the structure: 

```
{
  defaultAddr?: string;
  symbols?: {
    [symbol: string]: {
      defaultAddr?: string;
      chains?: { [chainId: string]: string };
    };
  };
}
```

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1450,7 +1450,9 @@ export function getDefaultRelayerAddress(
   destinationChainId: number,
   symbol?: string
 ) {
-  const symbolOverride = defaultRelayerAddressOverride?.symbols?.[symbol ?? ""];
+  const symbolOverride = symbol
+    ? defaultRelayerAddressOverride?.symbols?.[symbol]
+    : undefined;
   return (
     symbolOverride?.chains?.[destinationChainId] ?? // Specific Symbol/Chain override
     symbolOverride?.defaultAddr ?? // Specific Symbol override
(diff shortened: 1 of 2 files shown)


api/_utils.ts
```diff
@@ -1452,19 +1450,13 @@ export function getDefaultRelayerAddress(
...
-  if (overrideForToken?.destinationChains.includes(destinationChainId)) {
-    return overrideForToken.relayer;
-  } else {
-    return (
-      defaultRelayerAddressOverridePerChain[destinationChainId] ||
-      defaultRelayerAddressOverride ||
-      sdk.constants.DEFAULT_SIMULATED_RELAYER_ADDRESS
-    );
-  }
+  const symbolOverride = defaultRelayerAddressOverride?.symbols?.[symbol ?? ""];
```
pxrl, line 1453: What's the intendednbehaviour if symbol is undefined and `symbolOverride` resolves to `"""` ? If I'm reading this correctly then it looks like we might just resolve `""` as the relayer address.

pxrl (approved):
Do we have the SNX mapping in the env already?

api/_utils.ts
james-a-morris: No - this is a convenience on resolving `symbolOverride` to undefined. `record[""]` should undefined.

james-a-morris:
> Do we have the SNX mapping in the env already?

yep!

james-a-morris:
> Do we have the SNX mapping in the env already?

synced offline

PR #1159, opened 2024-08-20
enable depositExclusive

gsteenkamp89:
closes ACX-2529

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -233,7 +233,7 @@ export async function sendDepositV3Tx(
   fillDeadline ??=
     getCurrentTime() - 60 + (await spokePool.fillDeadlineBuffer());
 
-  const useExclusiveRelayer: boolean = !(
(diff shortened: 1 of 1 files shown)


src/utils/bridge.ts
```diff
@@ -233,7 +233,12 @@ export async function sendDepositV3Tx(
   fillDeadline ??=
     getCurrentTime() - 60 + (await spokePool.fillDeadlineBuffer());
 
-  const tx = await spokePool.populateTransaction.depositV3(
+  const useExclusiveRelayer: boolean = !(
```
dohaki, line 236: NIT: I don't think we need to explicitly set the type here and can just rely on inferring the type. wdyt?

dohaki (approved):
Just left one nit

src/utils/bridge.ts
gsteenkamp89: yeah true, it's unnecessary 👍

PR #1161, opened 2024-08-21
fix: routes generation script  + add usdzc swap routes

dohaki:
- Adds missing bridged USDC -> USDzC swap routes
- Refactors routes generation script for handling swap routes + bridged USDC routes

### DIFF
--- a/scripts/chain-configs/arbitrum/index.ts
+++ b/scripts/chain-configs/arbitrum/index.ts
@@ -30,21 +30,4 @@ export default {
   ],
   enableCCTP: true,
   blockTimeSeconds: 1,
-  swapTokens: [
-    {
-      swapInputTokenSymbol: "USDC.e",
-      acrossInputTokenSymbol: "USDC",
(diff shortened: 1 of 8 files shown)


gsteenkamp89:
Is it worthwhile writing tests for this script?

pxrl (approved):
The updated routes look good. Curious to see what you think RE @gsteenkamp89's question. Intuitively I think it seems tricky to test this automatically unless we end up  describing the routes separately, which is the purpose of this script anyway. wdyt - are there any options?

dohaki:
> Is it worthwhile writing tests for this script?

@gsteenkamp89 I also thought about that. But as @pxrl already mentioned, I wouldn't know what values to "expect" in those tests 🤔 The only thing we could write tests for are regarding formatting of the output etc.

Maybe we could add something to compare routes to the ones configured on-chain? But this wouldn't cover swap routes for example

gsteenkamp89 (approved):
looks good!

PR #1162, opened 2024-08-21
fix: infer root color, set stop opacity

gsteenkamp89:
Screenshot 2024-08-21 at 13 19 55


dohaki (approved):
Nice! That's a good fix

PR #1164, opened 2024-08-22
refactor: unify remote configs scripts + prep exclusive relayer weights pulling

dohaki:
While working on ACX-2576, I noticed that we should unify the scripts for pulling in remote configs. This PR therefore does the following:
- modified `scripts/fetch-relayer-config.ts` into `scripts/pre-build.ts` and refactored so that it can be used for different types of remote configs
- merged logic of `preBuild.sh` into above so that it can be used for different types of remote configs in the private bq-reader repo as well
- improved handling of dynamically created/updated config data so that diffing doesn't become annoying

Note that I commented the exclusive relayer weights related functionality because the counterpart work in [this repo](https://github.com/UMAprotocol/across-bq-reader) still needs to merged (I finished everything locally but need push access to open a PR)


src/data/examples/dynamic-weights.json
pxrl, line 1: We never discussed how this should look, but I'd assumed it'd be something like a single file with an object containing both fixed and dynamic weights together. Do you think there are advantages to one or the other? I don't really have an opinion on it.

```
{                                                                                                                                                                                              
  "0x07ae8551be970cb1cca11dd7a11f47ae82e70e67": {                                                                                                                                              
    "fixedWeight": 1.0,                                                                                                                                                                        
    "dynamicWeight": 1.0                                                                                                                                                                       
  }                                                                                                                                                                                            
}
```

scripts/remote-configs/utils.ts
```diff
@@ -0,0 +1,124 @@
...
+  number,
+  record,
+  string,
+  type,
+} from "superstruct";
+
+// Fallback data for remote configs in private repos or local development
+import fillTimesFallbackData from "../../src/data/examples/fill-times.json";
+import dynamicWeightsFallbackData from "../../src/data/examples/dynamic-weights.json";
+import fixedWeightsFallbackData from "../../src/data/examples/fixed-weights.json";
```
pxrl, line 15: ooc, could it possibly be useful to just have a default file in the expected path, and just overwrite it with the imported version at build time? In that way the code would never need to consider any kind of fallback to example config and could just assume that one file contains the necessary configuration.

src/data/examples/dynamic-weights.json
dohaki: Yea, so in the remote config repo, I think these should be separated actually. Because dynamic weights are updated periodically and automatically whereas fixed weights are done manually right?

dohaki: But after pulling them into this repo, it probably makes sense to merge them. Maybe even with the exclusive relayer configs?

scripts/remote-configs/utils.ts
dohaki: IIRC we opted for a different approach with the fill times because every time we build locally, this would override the default and create a diff 🤔

scripts/pre-build.ts
```diff
@@ -0,0 +1,83 @@
...
+  },
+  [remoteConfigTypes.EXCLUSIVE_RELAYERS]: {
+    fetchFn: () =>
+      fetchExclusiveRelayerConfigs(
+        "https://raw.githubusercontent.com/across-protocol/exclusive-relayer-configs",
+        "build/exclusive-relayer-configs.json",
+        getRemoteConfigCommitHash(remoteConfigTypes.EXCLUSIVE_RELAYERS)
+      ),
+    localFilePath: "src/data/exclusive-relayer-configs.json",
+  },
```
gsteenkamp89, line 37: very clean, i like that you're separating this definition from the script itself 👍

gsteenkamp89 (approved):
very neat 👏

src/data/examples/dynamic-weights.json
pxrl: Good point on splitting them according to how they are updated :+1: It's convenient to merge the fixed + dynamic weights as part of the pre-build step; at present there's a separation between the relayer's preferences and their associated weightings; we might ultimately merge them but for now I'm OK if they are separate, because it helps to give more structure to the implementation.

scripts/remote-configs/utils.ts
pxrl: `src/data/fill-times-preset.json` is ignored in .gitignore, so I don't think that'd ever show up in `git status`. Or are you referring to something else?

pxrl (approved):
Working OK for me!

PR #1167, opened 2024-08-23
feat: Relayer exclusivity plumbing

pxrl:
This commit attempts to add the code plumbing necessary for the exclusiveRelayer and exclusivityDeadline fields to be set based on the suggested-fees API response.

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -30,6 +30,8 @@ export type BridgeFees = {
   quoteBlock: ethers.BigNumber;
   limits: BridgeLimitInterface;
   estimatedFillTimeSec: number;
+  exclusiveRelayer: string;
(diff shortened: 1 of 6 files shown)


pxrl:
...disclaimer: I have no idea what I'm doing.

src/utils/bridge.ts
pxrl, line 1: This file needs an additional tweak to ensure that a relative exclusivityDeadline is set for depositExclusive().

dohaki: I can add that in a subsequent PR

dohaki (approved):
LGTM

gsteenkamp89 (approved):

pxrl:
Closed because the epic branch is using a relative timestamp instead, so that will just get merged as part of the larger feature.

PR #1168, opened 2024-08-26
feat: use exclusivity params in FE

dohaki:
Use exclusivity params in bridge action


dohaki:
Ah sorry this is already done in https://github.com/across-protocol/frontend/pull/1167

PR #1173, opened 2024-08-26
fix: strategy weights selection + token price + exclusivityPeriod 

dohaki:
Fixes some TODOs and remarks from [here](https://github.com/across-protocol/frontend/pull/1154)

Closes ACX-2590

### DIFF
--- a/api-docs-openapi.yaml
+++ b/api-docs-openapi.yaml
@@ -510,7 +510,7 @@ components:
             will be disabled. This value is returned in cases where using an
             exclusive relayer is not recommended.
           example: "0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010"
-        exclusivityDeadline:
+        exclusivityPeriod:
           type: string
           description: >
             The suggested exclusivity period (in seconds) the exclusive relayer should be given
--- a/api-docs-openapi.yaml
+++ b/api-docs-openapi.yaml
@@ -510,7 +510,7 @@ components:
             will be disabled. This value is returned in cases where using an
             exclusive relayer is not recommended.
           example: "0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010"
-        exclusivityPeriod:
+        exclusivityDeadline:
           type: string
           description: >
             The suggested exclusivity period (in seconds) the exclusive relayer should be given
--- a/api/_exclusivity/config.ts
+++ b/api/_exclusivity/config.ts
@@ -36,6 +36,7 @@ const EXCLUSIVITY_STRATEGIES = {
   none,
   randomWeighted,
 } as const;
+type ExclusivityStrategy = keyof typeof EXCLUSIVITY_STRATEGIES;
 
 /**
  * Select a specific relayer exclusivity strategy to apply.
@@ -45,11 +46,11 @@ const EXCLUSIVITY_STRATEGIES = {
  */
 export function getStrategy(): { name: string; selectorFn: RelayerSelector } {
   // @todo: use weights from global config
-  const weights = {
-    none: 0.8,
-    randomWeighted: 0.2,
-  } as const;
-  const keys = Object.keys(weights);
+  const weights: Record<ExclusivityStrategy, number> = {
+    none: 0,
+    randomWeighted: 1,
+  };
+  const keys = Object.keys(weights) as ExclusivityStrategy[];
   const cumulativeWeights = Object.values(weights).map((_, index) =>
     Object.values(weights)
       .slice(0, index + 1)
@@ -61,10 +62,7 @@ export function getStrategy(): { name: string; selectorFn: RelayerSelector } {
 
   return {
     name: strategyKey,
-    selectorFn:
-      EXCLUSIVITY_STRATEGIES[
-        strategyKey as keyof typeof EXCLUSIVITY_STRATEGIES
-      ],
+    selectorFn: EXCLUSIVITY_STRATEGIES[strategyKey],
   };
 }
 
--- a/api/_exclusivity/config.ts
+++ b/api/_exclusivity/config.ts
@@ -43,10 +43,29 @@ const EXCLUSIVITY_STRATEGIES = {
  * and selection from on env-based configuration.
  * @returns A handler function that will narrow an array of relayers to a single address.
  */
-export function getStrategy(): RelayerSelector {
-  // @todo: Determine strategy based on configuration.
-  const defaultStrategy = "randomWeighted";
-  return EXCLUSIVITY_STRATEGIES[defaultStrategy];
+export function getStrategy(): { name: string; selectorFn: RelayerSelector } {
+  // @todo: use weights from global config
+  const weights = {
+    none: 0.8,
+    randomWeighted: 0.2,
+  } as const;
+  const keys = Object.keys(weights);
+  const cumulativeWeights = Object.values(weights).map((_, index) =>
+    Object.values(weights)
+      .slice(0, index + 1)
+      .reduce((sum, w) => sum + w, 0)
+  );
+  const randomNumber = Math.random();
+  const strategyIndex = cumulativeWeights.findIndex((w) => randomNumber < w);
+  const strategyKey = keys[strategyIndex];
+
+  return {
+    name: strategyKey,
+    selectorFn:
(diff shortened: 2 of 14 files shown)


api/_exclusivity/index.ts
```diff
@@ -16,30 +16,42 @@ const { fixedPointAdjustment: fixedPoint } = sdk.utils;
...
+  tokenPriceUsd: BigNumber
 ): Promise<ExclusiveRelayer> {
+  let exclusiveRelayer = ZERO_ADDRESS;
+  let exclusivityPeriod = 0;
+
+  const { name, selectorFn } = getStrategy();
+
+  if (name === "none") {
+    return { exclusiveRelayer, exclusivityPeriod };
+  }
```
dohaki, line 41: Short circuit if none strategy to prevent blocking get balance calls

api/suggested-fees.ts
```diff
@@ -275,18 +275,12 @@ const handler = async (
       await selectExclusiveRelayer(
         computedOriginChainId,
         destinationChainId,
-        outputToken.address,
+        outputToken,
         amount.sub(totalRelayFee),
-        BigNumber.from(relayerFeeDetails.relayFeePercent) // @todo: Subtract destination gas cost.
+        BigNumber.from(relayerFeeDetails.relayFeePercent), // @todo: Subtract destination gas cost.
```
dohaki, line 280: @pxrl I was not sure about that `@todo` here. should I subtract the gas fee percent here?

api/_exclusivity/config.ts
```diff
@@ -36,17 +36,34 @@ const EXCLUSIVITY_STRATEGIES = {
...
-  // @todo: Determine strategy based on configuration.
-  const defaultStrategy = "randomWeighted";
-  return EXCLUSIVITY_STRATEGIES[defaultStrategy];
+export function getStrategy(): { name: string; selectorFn: RelayerSelector } {
+  // @todo: use weights from global config
+  const weights: Record<ExclusivityStrategy, number> = {
+    none: 0,
+    randomWeighted: 1,
+  };
+  const keys = Object.keys(weights) as ExclusivityStrategy[];
```
gsteenkamp89, line 53: nit: maybe you can use a const assertion for `weights` to avoid explicitly casting the type of its keys

dohaki: yea I tried that but getting this 🤔 
image

gsteenkamp89: ah you're right this only works for values.

api-docs-openapi.yaml
```diff
@@ -510,7 +510,7 @@ components:
             will be disabled. This value is returned in cases where using an
             exclusive relayer is not recommended.
           example: "0x428AB2BA90Eba0a4Be7aF34C9Ac451ab061AC010"
-        exclusivityDeadline:
```
pxrl, line 513: We've advertised `exclusivityDeadline` externally, so we may have integrators who are depending on it. We might need to retain it to avoid breaking anyone - what do you think?

gsteenkamp89 (approved):

api/_exclusivity/index.ts
```diff
@@ -16,30 +16,41 @@ const { fixedPointAdjustment: fixedPoint } = sdk.utils;
...
   destinationChainId: number,
-  outputToken: string,
+  outputToken: {
+    address: string;
+    decimals: number;
+  },
   outputAmount: BigNumber,
-  relayerFeePct: BigNumber
+  relayerFeePct: BigNumber,
+  tokenPriceUsd: BigNumber
```
pxrl, line 32: suggested-fees precomputes `amountInUsd`, which corresponds to the inputAmount * tokenPrice. I wonder if it could be "good enough" to use that value, since in most cases the fees paid by the user won't be enough to make a difference to whether the relayer wants to opt out of the fill based on its value. wdyt?

api/suggested-fees.ts
pxrl: Yeah I think we need to since, otherwise we're computing on gross relayer profit and are ignoring one of the key costs. I'd expect that we can derive it from values precomputed within suggested-fees - does that sound right?

pxrl (commented):
@dohaki This looks really good.

api-docs-openapi.yaml
dohaki: Ah ok, we advertised it already... I can change it back but the value will be the period correct?

api/_exclusivity/index.ts
dohaki: ok yea that makes sense to me. will change

api/suggested-fees.ts
dohaki: ok will add that change. yep we can easily access it from `relayerFeeDetails` directly

api-docs-openapi.yaml
pxrl: My thinking is that we probably need both fields concurrently. The problem with making `exclusivityDeadline` to a relative timestamp (i.e. an offset) is that it'll revert if used against `depositV3()`, and this is what all external integrators would be doing.

pxrl: Per discussion, it seems like we probably want to accept a new input field that identifies the method. This should deposit to depositV3, but can be overridden to depositExclusive, and in that case would return a relative offset. That would later become the default.

dohaki: https://github.com/across-protocol/frontend/pull/1173/commits/b6bd8f1c84e4c7315f347bf9528828fb6b0c3690

api/suggested-fees.ts
```diff
@@ -286,16 +288,17 @@ const handler = async (
...
-    // This is unrealistic and must be padded for consumers _other than_ the Across FE.
-    const exclusivityDeadline =
-      exclusivityPeriod > 0
-        ? sdk.utils.getCurrentTime() + exclusivityPeriod
-        : 0;
+    const exclusivityDeadline = getExclusivityDeadline(
+      exclusivityPeriod,
+      exclusiveRelayer,
+      depositMethod
+    );
```
pxrl, line 301: Could we just set `exclusivityPeriod` conditional upon the method? Then we could set it to 0 unless depositMethod is "depositExclusive".

api/suggested-fees.ts
```diff
@@ -286,16 +288,17 @@ const handler = async (
         destinationChainId,
         outputToken.address,
         amount.sub(totalRelayFee),
-        BigNumber.from(relayerFeeDetails.relayFeePercent), // @todo: Subtract destination gas cost.
+        amountInUsd,
+        BigNumber.from(relayerFeeDetails.relayFeePercent).sub(
+          relayerFeeDetails.gasFeePercent
+        ),
```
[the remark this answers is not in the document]
dohaki: Ah ok so we could just use the capitalFeePct here then?

api/suggested-fees.ts
dohaki: Sorry, I misunderstood our decision then. So you want to have `exclusivityPeriod` AND `exclusivityDeadline` in the response then? And depending on the provided method one is 0 and the other one >0 (if an exclusive relayer is found)

src/utils/bridge.ts
```diff
@@ -243,11 +243,17 @@ export async function sendDepositV3Tx(
...
+    // @todo: remove this when all other SpokePools are redeployed to support `depositExclusive`.
+    [690, 1135, 81457, 534352].includes(fromChain);
+
+  // @todo: remove this when all other SpokePools are redeployed to support `depositExclusive`.
+  // If we request a quote where exclusivity is enabled, but the SpokePool does not support
+  // `depositExclusive`, we should not use exclusivity.
+  if (!useExclusiveRelayer) {
+    exclusiveRelayer = ethers.constants.AddressZero;
+    exclusivityDeadline = 0;
+  }
```
pxrl, line 256: We could upstream this origin chain ID check to `api/_exclusivity` such that it's handled as part of `selectExclusiveRelayer()`. This would mean that `bridge.ts` could be mostly ignorant of the intricacies of exclusivity, and would just select `depositV3()` or `depositExclusive()` based on
```
const useExclusiveRelayer =
  !(
    exclusiveRelayer === ethers.constants.AddressZero &&
    exclusivityDeadline === 0
  );
```

pxrl (approved):

PR #1174, opened 2024-08-26
feat: add filter for minimum exclusivity period

gsteenkamp89:
Uses the existing logic that calculates the expected fill time for route and filters out exclusive relayer candidates based on their defined `minExclusivityPeriod`.

### DIFF
--- a/api/_exclusivity/index.ts
+++ b/api/_exclusivity/index.ts
@@ -24,15 +24,18 @@ export async function selectExclusiveRelayer(
   destinationChainId: number,
   outputToken: string,
   outputAmount: BigNumber,
-  relayerFeePct: BigNumber
+  relayerFeePct: BigNumber,
+  estimatedFillTimeSec: number
 ): Promise<ExclusiveRelayer> {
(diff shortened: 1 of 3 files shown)


api/_exclusivity/index.ts
```diff
@@ -24,15 +24,18 @@ export async function selectExclusiveRelayer(
   destinationChainId: number,
   outputToken: string,
   outputAmount: BigNumber,
-  relayerFeePct: BigNumber
+  relayerFeePct: BigNumber,
+  estimatedFillTimeSec: number
 ): Promise<ExclusiveRelayer> {
+  const exclusivityPeriodSec = getExclusivityPeriod(estimatedFillTimeSec);
```
pxrl, line 30: For follow-up where we add in the global configurable, I think we'll want to define a global maximum as well. That would fit the case where the estimate is (for example) greater than 30 seconds, in which case I imagine we wouldn't want to apply any exclusivity.

pxrl (approved):
LGTM!

PR #1176, opened 2024-08-27
feat: use rpc provider configs from remote

dohaki:
~Requires #1175~

Closes ACX-2586

Uses configs set in here https://github.com/UMAprotocol/across-configs/blob/master/packages/quotes-api-config/config.json. This allows for a workflow similar to the `fill-times.json` and bot configs.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -18,6 +18,7 @@ import { StructError, define } from "superstruct";
 
 import enabledMainnetRoutesAsJson from "../src/data/routes_1_0xc186fA914353c44b2E33eBE05f21846F1048bEda.json";
 import enabledSepoliaRoutesAsJson from "../src/data/routes_11155111_0x14224e63716afAcE30C9a417E0542281869f7d9e.json";
+import rpcProvidersJson from "../src/data/rpc-providers.json";
 
 import {
   MINIMAL_BALANCER_V2_POOL_ABI,
@@ -47,6 +48,7 @@ import {
 import { PoolStateOfUser, PoolStateResult } from "./_types";
 
 type LoggingUtility = sdk.relayFeeCalculator.Logger;
+type RpcProviderName = keyof typeof rpcProvidersJson.providers.urls;
 
 const {
   REACT_APP_HUBPOOL_CHAINID,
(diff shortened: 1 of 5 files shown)


src/data/examples/rpc-providers.json
```diff
@@ -0,0 +1,23 @@
+{
+  "providers": {
+    "enabled": {
+      "default": ["public"]
+    },
+    "urls": {
+      "public": {
+        "1": "https://mainnet.gateway.tenderly.co",
+        "10": "https://mainnet.optimism.io",
+        "137": "https://rpc.ankr.com/polygon",
```
pxrl, line 10: Alchemy might be an option here instead of Ankr: https://polygon-mainnet.g.alchemy.com/v2/demo

scripts/remote-configs/utils.ts
```diff
@@ -123,3 +135,7 @@ export function getRemoteConfigCommitHash(config: RemoteConfig) {
 export function getBqReaderRemoteBaseUrl() {
   return `https://raw.githubusercontent.com/UMAprotocol/across-bq-reader`;
 }
+
+export function getAcrossConfigsRemoteBaseUrl() {
+  return `https://raw.githubusercontent.com/UMAprotocol/across-configs`;
+}
```
pxrl, line 141: Should we make these configurable via env and have the hardcoded repo set as a fallback?

api/_utils.ts
```diff
@@ -701,6 +710,36 @@ export const getProvider = (
...
+  }
+
+  if (!providerUrl) {
+    console.error(
+      `No provider URL found for chainId ${chainId} in rpc-providers.json`
+    );
+    return undefined;
+  }
+
+  return new ethers.providers.StaticJsonRpcProvider(providerUrl);
```
pxrl, line 740: It might make sense to add a handler for rate-limiting here. That would help to identify when it happens, which could be useful with the public RPCs. We have an example in the relayer repo here: https://github.com/across-protocol/relayer/blob/1a6b1433b81d51ece2edb95a74ebaa5463d73fe4/src/utils/ProviderUtils.ts#L670

pxrl (approved):
LGTM!

api/_utils.ts
dohaki: Yes, good point. I wanted to tackle that a subsequent PR and in the context of one of these issues:
- https://linear.app/uma/issue/ACX-2582/implement-reasonable-fallback-instead-of-infura-on-fe
- https://linear.app/uma/issue/ACX-2429/support-multiple-rpc-providers-with-promiseany

PR #1178, opened 2024-08-27
feat: make exclusivity strategy selection configurable

dohaki:
Closes ACX-2593

### DIFF
--- a/src/data/examples/exclusivity-strategy.json
+++ b/src/data/examples/exclusivity-strategy.json
@@ -1,0 +1,11 @@
+{
+  "default": "none",
+  "tokens": {
+    "USDT": {
+      "default": "none",
+      "destinationChains": {
+        "690": "none"
+      }
+    }
+  }
+}
--- a/src/data/examples/exclusivity-strategy.json
+++ b/src/data/examples/exclusivity-strategy.json
@@ -4,7 +4,7 @@
     "USDT": {
       "default": "none",
       "destinationChains": {
-        "690": "none"
+        "137": "weightedRandom"
       }
     }
   }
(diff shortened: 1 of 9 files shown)


src/data/examples/exclusivity-strategy.json
```diff
@@ -0,0 +1,11 @@
+{
+  "default": "none",
```
pxrl, line 2: Should we have a top-level key called `availableStrategies` (or something) that enumerates the strategies that can be selected from? This could be used to validate that we don't accidentally specify an invalid string later in the config. (Or can we do it better somewhere else?)

src/data/examples/exclusivity-strategy.json
```diff
@@ -0,0 +1,11 @@
...
+  "tokens": {
+    "USDT": {
+      "default": "none",
+      "destinationChains": {
+        "690": "none"
```
pxrl, line 7: I'm keen to test this on Polygon in advance of having the e2e config flow up and running.

```suggestion
        "137": "weightedRandom"
```

src/data/examples/exclusivity-strategy.json
dohaki: Yea, I was thinking to add validation to the repo where the config lives in the end. But this would mean that there is a risk that the formats of the fallback and the real data get out of sync 🤔

src/data/examples/exclusivity-strategy.json
```diff
@@ -0,0 +1,11 @@
+{
+  "default": "none",
+  "tokens": {
+    "USDT": {
```
pxrl, line 4: I realised that we don't have good widespread support for USDT on the chains that currently support depositExclusive, so it limits our ability to test. WETH -> Polygon should be fine to test with.

```suggestion
    "WETH": {
```

pxrl (approved):

PR #1179, opened 2024-08-27
Drop getExclusivityDeadline

pxrl:
Not needed after the previous commit.


pxrl:
This is already handled here: https://github.com/across-protocol/frontend/pull/1178/files

PR #1180, opened 2024-08-28
feat: use retry provider in api

dohaki:
Closes ACX-2582

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -713,31 +713,41 @@ export const getProvider = (
 /**
  * Resolves a provider from the `rpc-providers.json` configuration file.
  */
-function getProviderFromConfigJson(chainId: string) {
+function getProviderFromConfigJson(_chainId: string) {
+  const chainId = Number(_chainId);
+  const urls: string[] = [];
+
   const { providers } = rpcProvidersJson;
   const enabledProviders: RpcProviderName[] =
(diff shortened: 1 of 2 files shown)


api/_utils.ts
```diff
@@ -713,31 +713,41 @@ export const getProvider = (
...
-  return new ethers.providers.StaticJsonRpcProvider(providerUrl);
+  return new sdk.providers.RetryProvider(
+    urls.map((url) => [url, chainId]),
+    chainId,
+    1, // quorum can be 1 in the context of the API
+    3, // retries
+    500, // delay
+    5, // max. concurrency
+    "QUOTES_API", // cache namespace
+    0 // disable RPC calls logging
```
dohaki, line 749: @pxrl @james-a-morris you think these make sense ?

gsteenkamp89: Why disable logging?

gsteenkamp89 (approved):

api/_utils.ts
james-a-morris: I think we should disable the logging for the end user. Maybe we can have this be conditioned on whether or not we're running in development/preview mode?

james-a-morris: These are sensible defaults - we can definitely adjust them as needed

james-a-morris (approved):

PR #1182, opened 2024-08-29
refactor: add rpc logs flag

dohaki:
Forgot to push before merging in the context of https://github.com/across-protocol/frontend/pull/1180#discussion_r1735225611

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -58,6 +58,7 @@ const {
   VERCEL_ENV,
   GAS_MARKUP,
   DISABLE_DEBUG_LOGS,
+  ENABLE_RPC_LOGS,
 } = process.env;
 
 const GOOGLE_SERVICE_ACCOUNT = _GOOGLE_SERVICE_ACCOUNT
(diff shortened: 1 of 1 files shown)


api/_utils.ts
```diff
@@ -746,7 +747,7 @@ function getProviderFromConfigJson(_chainId: string) {
     500, // delay
     5, // max. concurrency
     "QUOTES_API", // cache namespace
-    0 // disable RPC calls logging
+    ENABLE_RPC_LOGS === "true" ? 100 : 0
```
pxrl, line 750: Is it worth making this env var a number of range [0, 100]? Toggling to 100 might be like trying to drink from the firehose in a situation where other things might already be going wrong, and it might have an undesirable performance impact. In the bots we run w/ 1% or something like that, but it's nice to be able to tune it.

pxrl (approved):

api/_utils.ts
james-a-morris: +1 - this makes to include a way to configure the logging in a more granular way

james-a-morris (commented):
https://github.com/across-protocol/frontend/pull/1182#discussion_r1745926282

PR #1184, opened 2024-08-30
feat(api): add redis cache  + cache latest mainnet block

dohaki:
Closes ACX-2606 and ACX-2608


api/_cache.ts
```diff
@@ -0,0 +1,61 @@
...
+  ttl: number,
+  fetcher: () => Promise<T>,
+  parser?: (value: T) => T
+) {
+  const cachedValue = await redisCache.get<T>(key);
+  if (cachedValue) {
+    return parser ? parser(cachedValue) : cachedValue;
+  }
+
+  const value = await fetcher();
```
bmzig, line 58: Just so I'm understanding this correctly, we update the cache every time we access (it looks like we only access it via `getCachedValue`)? If that's true, does this necessarily mean it can fetch a value which hasn't been fetched recently and therefore returns an older, stale value?

dohaki: We update the cache every time there is no cached value (this could be the case if there was no value set ever for a given key or the value expired given a TTL). The `fetcher` function will return the most recent value and put it in the cache with a TTL. But yea, during the lifespan of that cached value, this function will return an older value (but not necessarily stale, because we intend it to be cached).

I think the caching approach of this function is slightly different from those of the bots. We normally don't need to cache values based on deterministic keys like `blockNumber`, etc. in the API. But rather the most recent values, like `latestGasPrice`, `latestGasEstimate`, `latestBlock`. So the approach here is to use a very small TTL (a few seconds) and return/use that older value for this short period.

bmzig: Ah that makes sense. Thanks.

bmzig (approved):

api/_utils.ts
```diff
@@ -731,8 +734,8 @@ function getProviderFromConfigJson(_chainId: string) {
     1, // quorum can be 1 in the context of the API
     3, // retries
     0.1, // delay
-    5, // max. concurrency
-    "QUOTES_API", // cache namespace
+    100, // max. concurrency
```
pxrl, line 737: Did you want to make this configurable per chain? 100 seems fine for Infura but not for smaller providers or public endpoints.

Additionally, do we have an idea of how many concurrent queries we'd expect here? I'd be surprised if we went more than single digits for any given instance.

api/suggested-fees.ts
```diff
@@ -165,7 +167,9 @@ const handler = async (
         throw new InputError("Invalid quote timestamp");
       }
 
-      const blockFinder = new sdk.utils.BlockFinder(provider, [latestBlock]);
+      const blockFinder = new sdk.utils.BlockFinder(provider, [
+        latestBlock as ethers.providers.Block,
```
pxrl, line 171: wdyt about asserting type further upstream - i.e. in `getCachedLatestBlock()`? That would save the consumer of the data from having to do it.

pxrl (commented):
@dohaki Do you have any impression of how this helps performance? I guess best case is probably only a marginal improvement but it should at least reduce the instances of worrst-case performance right?

api/_utils.ts
dohaki: Yea, this is me just tweaking things.. I don't know tbh how many concurrent queries we'd have per instance. Probably 100 is too high but my thinking was that previously (pre-RetryProvider) we didn't have a limit so it wouldn't hurt to set a high limit?

api/suggested-fees.ts
dohaki: yep good point

dohaki:
> @dohaki Do you have any impression of how this helps performance? I guess best case is probably only a marginal improvement but it should at least reduce the instances of worrst-case performance right?

Yea, a marginal improvement only I think. But I will also add other smaller improvements based on the same approach introduced here

api/_utils.ts
pxrl: Yeah I think that's OK. Are we logging/tracking the instance of 429s? That's probably the only additional thing I think we'd want for a limit like this.

pxrl (approved):
OK pending https://github.com/across-protocol/frontend/pull/1184#discussion_r1745140265 :+1:

api/_utils.ts
dohaki: Yep, these should get tracked on Datadog

PR #1185, opened 2024-08-31
fix: use rpc url from config json in balancer handler

dohaki:
Since we started using our custom `RetryProvider` in the API, we can not use `provider.connection.url` anymore because this resolves to `localhost:8545`. This broke the Balancer pool handler. We fix this by using the provider URL from the config JSON directly.


pxrl (approved):
LGTM

PR #1187, opened 2024-09-03
fix: use retry delay as seconds

dohaki:
Retry delay in `RetryProvider` was set to high

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -729,7 +729,7 @@ function getProviderFromConfigJson(_chainId: string) {
     chainId,
     1, // quorum can be 1 in the context of the API
     3, // retries
-    500, // delay
+    0.5, // delay
     5, // max. concurrency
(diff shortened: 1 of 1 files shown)


api/_utils.ts
```diff
@@ -729,7 +729,7 @@ function getProviderFromConfigJson(_chainId: string) {
     chainId,
     1, // quorum can be 1 in the context of the API
     3, // retries
-    500, // delay
+    0.1, // delay
```
pxrl, line 732: We use 1 in the bots, fwiw. I don't know whether we've ever tuned it much, though.

pxrl (approved):

pxrl (approved):

api/_utils.ts
dohaki: Ok, yea I don't want to impact the perf too much. Let's see how this works

PR #1194, opened 2024-09-06
perf: api speed improvements

dohaki:
Closes ACX-2429 and ACX-2610

This PR introduces more potential speed improvements:
1. Usage of [SpeedProvider](https://github.com/across-protocol/frontend/compare/more-speed-improvements?expand=1#diff-428e8c9eb80e1971d1eb1d20244ebe4791f3f5d6c6c664cb694759f503874251R793-R800) 
2. Caching of fill gas units
3. ~Re-introduction of using cached gas price with tweaked params~ EDIT: Will be postponed to separate PR
4. Usage of cached token balances from redis instead of self call

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -613,10 +613,10 @@ export const getRelayerFeeDetails = async (
     originChainId: number;
     destinationChainId: number;
     recipientAddress: string;
+    message?: string;
+    relayerAddress?: string;
   },
   tokenPrice?: number,
-  message?: string,
-  relayerAddress?: string,
   gasPrice?: sdk.utils.BigNumberish,
   gasUnits?: sdk.utils.BigNumberish
 ): Promise<sdk.relayFeeCalculator.RelayerFeeDetails> => {
@@ -627,6 +627,8 @@ export const getRelayerFeeDetails = async (
     originChainId,
     destinationChainId,
     recipientAddress,
+    message,
+    relayerAddress,
   } = deposit;
   const relayFeeCalculator = getRelayerFeeCalculator(destinationChainId, {
     relayerAddress,
@@ -646,7 +648,7 @@ export const getRelayerFeeDetails = async (
       sdk.utils.isMessageEmpty(message),
       relayerAddress,
       tokenPrice,
-      gasPrice, // FIXME: We need properly cache the gas price before this can be used reliably
+      gasPrice,
       gasUnits
     );
   } catch (err: unknown) {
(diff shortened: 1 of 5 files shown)


api/_utils.ts
```diff
@@ -728,15 +790,13 @@ function getProviderFromConfigJson(_chainId: string) {
     return undefined;
   }
 
-  return new sdk.providers.RetryProvider(
+  return new sdk.providers.SpeedProvider(
```
dohaki, line 793: This SpeedProvider makes use of `Promise.any`, i.e. fires of multiple RPC requests at once and returns the fastest successful one. Errors if all error

api/_utils.ts
```diff
@@ -1764,14 +1824,17 @@ export function getCachedLatestBlock(chainId: number) {
 
 export function getCachedGasPrice(chainId: number) {
   const ttlPerChain = {
-    default: 10,
-    [CHAIN_IDs.MAINNET]: 12,
+    default: 5,
+    [CHAIN_IDs.ARBITRUM]: 2,
```
dohaki, line 1828: Not sure if these make more sense now and should prevent errors described [here](https://github.com/across-protocol/frontend/pull/1193#issue-2507955022)

api/_utils.ts
```diff
@@ -1764,14 +1824,17 @@ export function getCachedLatestBlock(chainId: number) {
...
   };
 
   return getCachedValue(
     buildInternalCacheKey("gasPrice", chainId),
     ttlPerChain[chainId] || ttlPerChain.default,
-    () => getProvider(chainId).getGasPrice(),
+    async () => {
+      const gasPrice = await getProvider(chainId).getGasPrice();
+      return gasPrice.mul(2);
+    },
```
dohaki, line 1837: Same as above

api/_utils.ts
```diff
@@ -1794,6 +1857,42 @@ export function getCachedLatestBalance(
...
+  deposit: Parameters<typeof buildDepositForSimulation>[0],
+  overrides?: Partial<{
+    spokePoolAddress: string;
+    relayerAddress: string;
+  }>
+) {
+  const ttlPerChain = {
+    default: 10,
+    [CHAIN_IDs.ARBITRUM]: 10,
+  };
```
dohaki, line 1870: We might need to tweak those as well

nicholaspai (approved):
Let's try this out and test if responses are actually faster with caching

PR #1196, opened 2024-09-06
fix: suggested-fees min amounts

james-a-morris:
Test WIP in the preview links to confirm logging

### DIFF
--- a/api/suggested-fees.ts
+++ b/api/suggested-fees.ts
@@ -265,9 +265,12 @@ const handler = async (
       gasPrice
     );
 
-    const skipAmountLimitEnabled = skipAmountLimit === "true";
+    const isAmountTooLow =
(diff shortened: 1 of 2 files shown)


api/suggested-fees.ts
```diff
@@ -287,6 +288,10 @@ const handler = async (
           amountInUsd
         );
 
+    const isAmountTooLow =
```
nicholaspai, line 291: I think you should move this above line 270 so we're always referencing the same isAmountTooLow

james-a-morris: Done: https://github.com/across-protocol/frontend/pull/1196/commits/eba53ca87163522741b99f9a065b7593af983c93

nicholaspai (approved):

PR #1197, opened 2024-09-06
improve(limits): Allow developer to set minDepositUSD

nicholaspai:
It's easier to maintain this value if we set a global USD min default and allow it to be overridden per chain

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -85,15 +85,12 @@ const handler = async (
       outputToken,
     } = validateChainAndTokenParams(query);
 
-    const minDepositUsdForDestinationChainId = Number(
-      process.env[`MIN_DEPOSIT_USD_${destinationChainId}`] ??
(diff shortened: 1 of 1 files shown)


api/limits.ts
```diff
@@ -88,6 +84,12 @@ const handler = async (
       outputToken,
     } = validateChainAndTokenParams(query);
 
+    const minDepositUsdForDestinationChainId = Number(
+      process.env[`REACT_APP_MIN_DEPOSIT_USD_${destinationChainId}`] ??
+        REACT_APP_MIN_DEPOSIT_USD ??
```
pxrl, line 89: ooc, does this actually need the REACT_APP prefix? I thought that was what gets pushed client side, but this is in the API.

nicholaspai: fair point

james-a-morris (approved):

PR #1198, opened 2024-09-06
feat(API): Align limits and suggested-fees relayerFeeDetails computations

nicholaspai:
Add the following optional parameters to `/limits` to make it possible to compute the exact same limits in suggested fees:

- message
- recipient
- relayer
- amount

This way the `/suggested-fees` endpoint can query `/limits` using those parameters so it doesn't need to recompute `relayerFeeDetails()` which might return very different values.

I believe this also removes the need for `relayerFeeDetails` to compute the `isAmountTooLow` boolean anymore.

### DIFF
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -96,7 +96,7 @@ const handler = async (
     // to compute limits.
     let { amount: amountInput, recipient, relayer, message } = query;
     recipient ??= DEFAULT_SIMULATED_RECIPIENT_ADDRESS;
-    relayer ??= getDefaultRelayerAddress(destinationChainId, inputToken.symbol);
+    relayer ??= getDefaultRelayerAddress(destinationChainId, l1Token.symbol);
     if (sdk.utils.isDefined(message)) {
       if (!sdk.utils.isDefined(amountInput)) {
         throw new InputError("amount must be defined when message is defined");
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -200,6 +200,11 @@ const handler = async (
         )
       ),
     ]);
+    logger.debug({
+      at: "Limits",
+      message: "Relayer fee details from SDK",
+      relayerFeeDetails,
+    });
 
     let { liquidReserves } = multicallOutput[1];
     const [liteChainIdsEncoded] = multicallOutput[2];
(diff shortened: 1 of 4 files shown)


nicholaspai:
This might slow down suggested-fees too much since limits won't be cached nearly as much (if at all)

api/limits.ts
```diff
@@ -137,13 +165,13 @@ const handler = async (
...
         computedOriginChainId,
         destinationChainId,
-        DEFAULT_SIMULATED_RECIPIENT_ADDRESS,
+        recipient,
         tokenPriceNative,
-        undefined,
-        getDefaultRelayerAddress(destinationChainId, l1Token.symbol),
+        message,
+        relayer,
         gasPrice
```
dohaki, line 175: I think we need to return the relayerFeeDetails as part of the `/limits` response as well?
```
https://app-frontend-v3-git-npai-align-uma.vercel.app/api/limits?inputToken=0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2&outputToken=0x82aF49447D8a07e3bd95BD0d56f35241523fBab1&originChainId=1&destinationChainId=42161
```
still returns the "old" limits response and therefore suggested-fees is failing

api/suggested-fees.ts
```diff
@@ -207,32 +178,36 @@ const handler = async (
...
+        // Only pass in the following parameters if message is defined, otherwise leave them undefined so we are more
+        // likely to hit the /limits cache using the above parameters that are not specific to this deposit.
+        depositWithMessage ? amountInput : undefined,
+        depositWithMessage ? recipient : undefined,
+        depositWithMessage ? relayer : undefined,
+        depositWithMessage ? message : undefined
       ),
-      getCachedGasPrice(destinationChainId),
     ]);
+    const { maxDeposit, maxDepositInstant, minDeposit, relayerFeeDetails } =
```
dohaki, line 199: `/limits` seems to not return `relayerFeeDetails` currently. see comment above

dohaki (commented):
Nice stuff 👍 Left one remark regarding missing field in the limits response

api/suggested-fees.ts
dohaki: https://github.com/across-protocol/frontend/pull/1200

dohaki (approved):

PR #1204, opened 2024-09-10
perf: improve api speed by caching gas units + speed provider

dohaki:
Re-opened fixed #1194

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -14,7 +14,14 @@ import {
 } from "@balancer-labs/sdk";
 import { Log, Logging } from "@google-cloud/logging";
 import axios from "axios";
-import { BigNumber, ethers, providers, Signer, utils } from "ethers";
+import {
+  BigNumber,
+  BigNumberish,
+  ethers,
+  providers,
+  utils,
+  Signer,
+} from "ethers";
 import { StructError, define } from "superstruct";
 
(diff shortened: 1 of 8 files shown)


api/_utils.ts
```diff
@@ -630,55 +647,101 @@ export const getTokenSymbol = (tokenAddress: string): string => {
...
+    amount,
+    inputToken,
+    outputToken,
+    recipientAddress,
+    originChainId,
+    destinationChainId,
+    message,
+  } = depositArgs;
+  // Small amount to simulate filling with. Should be low enough to guarantee a successful fill.
+  const safeOutputAmount = sdk.utils.toBN(100);
```
gsteenkamp89, line 722: maybe a stupid question, but won't we now have successful simulations even for amounts that would actually make the fill unprofitable for relayers? or is that what we want?

dohaki: not a stupid question!

it is indeed unprofitable but it won't revert because we are trying to fill a "virtual" deposit that doesn't exist. so theoretically anyone could just call fill with some made-up deposit data. but they wouldn't get refunds of course because there is no matching deposit

dohaki: and because we just want to get the gas expenditure of the transaction this is fine

gsteenkamp89: Oh i see, the purpose of this simulation is just to estimate gas, got it. 👍

gsteenkamp89 (approved):

PR #1209, opened 2024-09-12
perf: cron job for caching relayer balances

dohaki:
This PR introduces a cron-scheduled handler that aims to keep relayer balances for all tokens on all chains fresh. Theoretically, this should further even out worst-case latencies for /limits and /suggested-fees.

Closes ACX-2661

We could use a similar approach for fairly constant and static values as well, such as `gasPrice`, `fillGasExpenditure`, `latestBlock`, etc. One issue is that Vercel only supports a min. interval of minutes. So for TTL values that are in seconds, we would need to create another external trigger

### DIFF
--- a/api/cron-cache-balances.ts
+++ b/api/cron-cache-balances.ts
@@ -1,9 +1,10 @@
 import { VercelResponse } from "@vercel/node";
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
 import { TypedVercelRequest } from "./_types";
 
 import {
   HUB_POOL_CHAIN_ID,
+  getCachedTokenBalances,
   getLogger,
   handleErrorCondition,
   latestBalanceCache,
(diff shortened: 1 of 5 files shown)


api/cron-cache-balances.ts
```diff
@@ -0,0 +1,95 @@
...
+            chain.outputTokens.map((token) => {
+              return latestBalanceCache({
+                chainId: chain.chainId,
+                address: relayer,
+                tokenAddress: token.address,
+              }).set(
+                BigNumber.from(batchResult.balances[relayer][token.address])
+              );
+            })
+          );
```
pxrl, line 79: Should we consider using the batched balance query endpoint? It permits us to query multiple balances across multiple tokens in a single call.

api/cron-cache-balances.ts
```diff
@@ -0,0 +1,95 @@
...
+    const allRelayers = [...fullRelayers, ...transferRestrictedRelayers];
+    for (const chain of mainnetChains) {
+      const batchResult = await getCachedTokenBalances(
+        chain.chainId,
+        allRelayers,
+        [
+          ethers.constants.AddressZero,
+          ...chain.outputTokens.map((token) => token.address),
+        ]
+      );
```
pxrl, line 65: So I think this calls through to the batched account balances endpoint that was recently added, and I _think_ it could theoretically also be served out of vercel's cache. Is it worth bypassing the endpoint and doing the RPC query directly? This would be basically doing exactly what the endpoint does, so maybe the implementation could be factored out.

dohaki: Yes, good point. Luckily the implementation is already factored out. So changing this was just a one-liner https://github.com/across-protocol/frontend/pull/1209/commits/ec7b2cabff5e751ee3a48d6e7f45f6cc6debd8d8

pxrl: Nice one @gsteenkamp89 !

pxrl (approved):

PR #1212, opened 2024-09-13
fix: only use cached gas units if message empty

dohaki:
Cached gas units should only be used for standard bridges with constant gas expenditures


pxrl (approved):
Tested OK by Clip :+1:

PR #1213, opened 2024-09-13
perf: cache `isDeployedContract` value

dohaki:


api/_utils.ts
```diff
@@ -1923,6 +1923,20 @@ export function latestBalanceCache(params: {
   );
 }
 
+export function getCachedIsContract(chainId: number, address: string) {
+  return getCachedValue(
+    buildInternalCacheKey("isContract", chainId, address),
+    24 * 60 * 60, // 1 day - we can cache this for a long time
```
dohaki, line 1929: Does one day make sense? Can this be even longer? I could also set up some CRON handler that pushes fresh values into the cache for a set of known addresses, like the generic MultiCallHandler contract

api/limits.ts
```diff
@@ -101,9 +101,11 @@ const handler = async (
     const isMessageDefined = sdk.utils.isDefined(message);
     if (isMessageDefined) {
       if (!sdk.utils.isDefined(amountInput)) {
-        throw new InputError("amount must be defined when message is defined");
+        throw new InputError(
+          "Parameter 'amount' must be defined when 'message' is defined"
+        );
       }
-      validateDepositMessage(
+      await validateDepositMessage(
```
dohaki, line 108: This needs to be awaited and therefore impact performance a little bit

api/suggested-fees.ts
```diff
@@ -92,16 +92,6 @@ const handler = async (
...
-    if (depositWithMessage) {
-      validateDepositMessage(
-        recipient,
-        destinationChainId,
-        relayer,
-        outputToken.address,
-        amountInput,
-        message!
-      );
-    }
```
dohaki, line 104: We can remove this potentially blocking validation here because it is done in limits

api/_utils.ts
pxrl: I don't see why it couldn't be longer, though I guess the cost of querying this once per day is already a massive reduction on what we do now.

Also, for contract addresses that we control, could we even just check for those first before we fall back to redis? That would avoid the blocking call in a lot of cases (assuming most messages go via the multicallhandler, which is the hope).

pxrl (approved):

api/_utils.ts
dohaki: > Also, for contract addresses that we control, could we even just check for those first before we fall back to redis? That would avoid the blocking call in a lot of cases (assuming most messages go via the multicallhandler, which is the hope).

Oh yea that totally makes sense!

bmzig (approved):

PR #1214, opened 2024-09-18
fix: handle empty addresses query param

dohaki:
Fixes ACX-2700

I noticed a few 500 response codes in our monitoring tool when calling /suggested-fees. It seems that these happen when exclusivity is set but there is no relayer opted in for a specific origin chain


gsteenkamp89 (approved):
good catch!

bmzig (approved):
LGTM

PR #1217, opened 2024-09-23
feat(api): improved error objects

dohaki:
Closes ACX-2705

This PR proposes improved error objects returned by the API. The goal is to have better DX and debugging possibilities.

### DIFF
--- a/api/suggested-fees.ts
+++ b/api/suggested-fees.ts
@@ -30,6 +30,11 @@ import {
 import { selectExclusiveRelayer } from "./_exclusivity";
 import { resolveTiming, resolveRebalanceTiming } from "./_timings";
 import { parseUnits } from "ethers/lib/utils";
+import {
+  InvalidParamError,
+  AmountTooHighError,
+  AmountTooLowError,
+} from "./_errors";
 
 const { BigNumber } = ethers;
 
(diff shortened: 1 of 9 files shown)


bmzig (approved):
LGTM

api/suggested-fees.ts
```diff
@@ -228,7 +236,12 @@ const handler = async (
 
     const skipAmountLimitEnabled = skipAmountLimit === "true";
     if (!skipAmountLimitEnabled && isAmountTooLow) {
-      throw new InputError("Sent amount is too low relative to fees");
+      throw new AmountTooLowError({
+        message: `Sent amount is too low relative to fees. Minimum deposit is ${ethers.utils.formatUnits(
+          minDeposit,
```
pxrl, line 241: There is a UX quirk on sharing `minDeposit`: it's set predominantly based on destination gas, so it's subject to fluctuate slightly between queries. If a caller gets this error response and then resubmits w/ `minDeposit` there's a decent chance that it will fail again, and they'll need to set another `minDeposit`.

idk whether we can (or even should) do anything about that in this context, but it's worth thinking about for a follow-up change. The only thing I can think of is it round `minDeposit` up to some round number to give some padding. That's an approach with its own flaws, though.

pxrl (approved):
Awesome!

api/suggested-fees.ts
dohaki: Hmm yea that's a good point. I will remove the value from he error then for now

PR #1219, opened 2024-09-24
improve: update token logo

james-a-morris:


src/assets/chain-logos/optimism-grayscale.svg
dohaki, line 1: Hmm that doesn't look right?

dohaki: image

dohaki: For OP -> Base

dohaki (changes requested):
Ah nice, wasn't aware of the new logo! But I think the grayscale versions are off. See my comment below.

Also, can we update the logos in
- https://github.com/across-protocol/frontend-v2/blob/ac42b5c72ce7198336129e9cd767e911c5361716/scripts/chain-configs/optimism/assets/grayscale-logo.svg
- https://github.com/across-protocol/frontend-v2/blob/ac42b5c72ce7198336129e9cd767e911c5361716/scripts/chain-configs/optimism/assets/logo.svg
as well? And run 
```
yarn generate:ui-assets
```
This will update the respective files `assets/chain-logos` automatically

dohaki (approved):
Niceee

PR #1222, opened 2024-09-27
perf(api): cache gas price

dohaki:
Closes ACX-2609


gsteenkamp89 (approved):
nice one. Linter says you can remove some unused Errors.

PR #1227, opened 2024-10-01
fix: decrease deposit status retry delay

dohaki:
Try to minimize discrepancies between perceived deposit/fill times and real times


gsteenkamp89 (approved):
hopefully this helps!

PR #1228, opened 2024-10-01
feat: use separate fill times json for exclusivity

dohaki:
Closes ACX-2833


src/data/examples/exclusivity-fill-times.json
bmzig, line 1: Does this mean we only will support exclusive relays for deposits to/from 1,10,324 under 1000 USD?

bmzig (approved):

src/data/examples/exclusivity-fill-times.json
pxrl: This is only the example config, so it's used in case we can't retrieve the actual config. It shouldn't ever be used in practice, though it could be activated if we're deploying and github is having issues.

api/_timings.ts
```diff
@@ -21,71 +21,93 @@ const rebalanceTimeOverrides: {
...
+ */
+export const resolveExclusivityTiming = makeTimingResolver(
+  exclusivityTimingsLookup
+);
 
 export function resolveRebalanceTiming(destinationChainId: string): number {
   return rebalanceTimeOverrides[destinationChainId] ?? 900; // 15 minutes
 }
+
+function makeTimingResolver(lookup: ReturnType<typeof makeLookup>) {
```
pxrl, line 44: @dohaki The linter is complaining that `lookup` is not actually read by this function. Is that intended?

pxrl (commented):
@dohaki Does this PR need to consider the fixes that James prepared in [this PR](https://github.com/across-protocol/frontend/pull/1215) from 2 weeks ago? It was unfortunately never merged.

src/data/examples/exclusivity-fill-times.json
bmzig: Ah of course, my bad.

api/_timings.ts
dohaki: ohh good catch! this would have resulted in a bug actually

dohaki:
> @dohaki Does this PR need to consider the fixes that James prepared in [this PR](https://github.com/across-protocol/frontend/pull/1215) from 2 weeks ago? It was unfortunately never merged.

Yea that's a good point. I will ping @james-a-morris about this.

PR #1229, opened 2024-10-01
nit(ui-assets): regenerate zora

james-a-morris:
This is a change that is created when running `yarn generate:ui-assets`. This change ensures that if we run this command again there will be no changes.


dohaki (commented):
Good catch! But I think we need to update the source files in `scripts/chain-configs/zora` with the existing ones in `src/assets/chain-logos/zora-grayscale.svg`. So basically the other way around. I remember @gsteenkamp89 did some fixes to the SVG so that the animations correctly work

PR #1231, opened 2024-10-03
allow filtering by chainId or list of chainIds

gsteenkamp89:
closes ACX-2842

- allows consumer to filter /chains response by list of chainIds. this means the sdk can easily retrieve and cache chain specific data (spokePool address etc.)


dohaki (approved):
Nice

PR #1233, opened 2024-10-07
fix: remove gcp logger

dohaki:
We don't use GCP logging for our production logs; instead, we use Datadog with log drains from Vercel. Therefore, this PR removes the dependency for now.

I also want to exclude the possibility that this logger might create issues with the new default streaming behavior of serverless functions in the context of ACX-2854


bmzig (approved):

gsteenkamp89 (approved):
🤞

PR #1237, opened 2024-10-07
improve: Isolate host time from deposit fillDeadline

pxrl:
If the host time is out of sync with the origin chain time then it may be possible to set fillDeadline such that it would revert on simulation. Per existing implementation, this could occur if the RPC time was > 60 seconds behind the host time, or if the host time was > 60 seconds ahead. Using the time of the chain and the fillDeadlineBuffer offset should protect against this possibility.

### DIFF
--- a/src/utils/bridge.ts
+++ b/src/utils/bridge.ts
@@ -1,6 +1,4 @@
 import { ethers, BigNumber } from "ethers";
-
-import { callViaMulticall3 } from "../../api/_utils";
 import {
   ChainId,
   fixedPointAdjustment,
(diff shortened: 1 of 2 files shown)


src/utils/bridge.ts
```diff
@@ -1,5 +1,5 @@
 import { ethers, BigNumber } from "ethers";
-
+import { callViaMulticall3 } from "../../api/_utils";
```
pxrl, line 2: Importing from api is a temporary bodge. Should `callViaMulticall3` be relocated?

dohaki (approved):

Emadalshamery:
Emadalshamery

PR #1238, opened 2024-10-08
improve: add exponential smoothing to cached gas fees

james-a-morris:


api/cron-cache-gas-prices.ts
```diff
@@ -50,27 +55,43 @@ const handler = async (
...
+        await cache.set(newGasAverage.div(fixedPointAdjustment));
         // Sleep for `updateIntervalSec` seconds
-        const updateIntervalSec =
-          updateIntervalsSecPerChain[
-            chain.chainId as keyof typeof updateIntervalsSecPerChain
-          ] || updateIntervalsSecPerChain.default;
-        await new Promise((resolve) =>
-          setTimeout(resolve, updateIntervalSec * 1000)
-        );
+        await utils.delay(secondsPerUpdateForChain);
```
james-a-morris, line 94: We're already using the sdk in our other imports to `_utils` and these functions are hot loaded so we don't need to be as cautious to add addtl imports

api/cron-cache-gas-prices.ts
```diff
@@ -3,18 +3,23 @@ import { TypedVercelRequest } from "./_types";
...
 import mainnetChains from "../src/data/chains_1.json";
+import { fixedPointAdjustment } from "utils";
+import { utils } from "@across-protocol/sdk";
 
 const updateIntervalsSecPerChain = {
   default: 10,
   1: 12,
 };
 
+const SECONDS_OF_GAS_AVERAGE = 2 * 60 * 60; // 2 hours
```
james-a-morris, line 21: We can change this as needed

api/cron-cache-gas-prices.ts
```diff
@@ -50,27 +55,43 @@ const handler = async (
...
-        await latestGasPriceCache(chain.chainId).set();
-
+        // Resolve the caching abstraction for the current chain
+        const cache = latestGasPriceCache(chain.chainId);
+        // Grab into the cache to get the previous value of the gas price.
+        // Note: If the cache is empty, this fn will return a direct
+        //       value from the chain.
+        const previousGasPrice = await cache.get();
+        // Directly compute the current value of the gas price
+        const currentGasPrice = await getMaxFeePerGas(chain.chainId);
```
dohaki, line 79: NIT: can we wrap in a `Promise.all`? even though this isn't in a perf critical path

dohaki (approved):
Looks good 👍 Left one nit

api/cron-cache-gas-prices.ts
james-a-morris: fair call - done

api/_utils.ts
```diff
@@ -1924,22 +1924,28 @@ export function latestGasPriceCache(chainId: number) {
...
 
+/**
+ * Resolve the current gas price for a given chain
+ * @param chainId The chain ID to resolve the gas price for
+ * @returns The gas price in the native currency of the chain
+ */
+export async function getMaxFeePerGas(chainId: number): Promise<BigNumber> {
+  if (sdk.utils.chainIsOPStack(chainId)) {
+    const l2Provider = asL2Provider(getProvider(chainId));
+    return l2Provider.getGasPrice();
```
nicholaspai, line 1940: can we also smooth the op stack gas prices?

api/cron-cache-gas-prices.ts
```diff
@@ -50,27 +55,45 @@ const handler = async (
...
         // Stop after `maxDurationSec` seconds
         if (diff >= maxDurationSec * 1000) {
           break;
         }
-
-        // Update gas price every `updateIntervalSec` seconds
-        await latestGasPriceCache(chain.chainId).set();
-
+        // Resolve the caching abstraction for the current chain
+        const cache = latestGasPriceCache(chain.chainId);
```
nicholaspai, line 73: should we move this out of the while loop?

api/cron-cache-gas-prices.ts
```diff
@@ -50,27 +55,45 @@ const handler = async (
...
+        //       value from the chain.
+        const [previousGasPrice, currentGasPrice] = await Promise.all([
+          cache.get(),
+          getMaxFeePerGas(chain.chainId),
+        ]);
+        // Resolve and scale our alpha value
+        const alpha = fixedPointAdjustment
+          .mul(secondsPerUpdateForChain)
+          .div(SECONDS_OF_GAS_AVERAGE);
+        // We are computing an expontential weighted moving average of the
```
nicholaspai, line 86:
```suggestion
        // We are computing an exponential weighted moving average of the
```

api/cron-cache-gas-prices.ts
```diff
@@ -50,27 +55,45 @@ const handler = async (
...
+        // gas price. To do this, we will follow the formula:
+        // newValue = (1 - alpha) * previousValue + alpha * currentValue
```
nicholaspai, line 88: what's the source for this algo?

nicholaspai (approved):
Looks great

api/cron-cache-gas-prices.ts
james-a-morris: It's a type of smoothing done for time-series: https://en.wikipedia.org/wiki/Exponential_smoothing

api/_utils.ts
james-a-morris: This will be smoothed as part of the caching logic. We grab this data for all chains

api/cron-cache-gas-prices.ts
james-a-morris: Fair call - https://github.com/across-protocol/frontend/pull/1238/commits/214e6e992cd151bf5ef82b892f5f902a8fd561f1

PR #1239, opened 2024-10-08
Feat/set default origin chain to connected chain

gsteenkamp89:
closes ACX-2171

Finds a default route where fromChain is the connected chain

### DIFF
--- a/src/views/Bridge/hooks/useSelectRoute.ts
+++ b/src/views/Bridge/hooks/useSelectRoute.ts
@@ -7,7 +7,7 @@ import {
   trackQuickSwap,
   similarTokensMap,
 } from "utils";
-import { useAmplitude } from "hooks";
+import { useAmplitude, useConnection } from "hooks";
 
 import {
   getRouteFromQueryParams,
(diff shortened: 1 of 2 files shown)


src/views/Bridge/hooks/useSelectRoute.ts
```diff
@@ -7,26 +7,36 @@ import {
...
 
   const { addToAmpliQueue } = useAmplitude();
 
+  // set default fromChain when user first connects
+  useEffect(() => {
+    if (isConnected) {
+      setSelectedRoute(getInitialRoute({ fromChain: walletChainId }));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isConnected]);
```
dohaki, line 38: So this will have the highest precedence right now correct? Is this wanted? For example, if I use the link https://app-frontend-v3-git-feat-set-default-origin-chain-to-bcfa47-uma.vercel.app/bridge?fromChainId=10&toChainId=42161 but I am connected to mainnet, it will override that behavior...

I think we should only trigger this effect hook, if there are no query params. Or more specifically if there is no `fromChain` query param

dohaki (commented):
I think we need to refine the behavior a bit more. See my comment below

src/views/Bridge/hooks/useSelectRoute.ts
pxrl: That makes sense to me :+1:

dohaki (approved):
LGTM 👍

PR #1240, opened 2024-10-09
Feat world chain support

gsteenkamp89:
closes ACX-2717


scripts/chain-configs/world-chain/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export default {
+  ...chainInfoBase,
+  fullName: "World Chain",
+  logoPath: "./assets/logo.svg",
+  grayscaleLogoPath: "./assets/grayscale-logo.svg",
+  spokePool: getDeployedAddress("SpokePool", chainId),
+  chainId,
+  publicRpcUrl: "https://worldchain-mainnet.g.alchemy.com/public",
+  blockTimeSeconds: 2,
+  tokens: ["WBTC", "WETH"], // "USDC.e" disabled for now
```
dohaki, line 19: We need to add `ETH` here as well no?

dohaki (commented):
I think we are missing ETH as an input token

PR #1242, opened 2024-10-10
improve: chain selector update

james-a-morris:


src/utils/constants.ts
```diff
@@ -521,8 +521,7 @@ export const gasExpenditureDeposit = BigNumber.from(90_000);
 // Used to determine whether to show the "delayed" warning in the deposits table
 export const pendingStateTimeUntilDelayed = 5 * 60; // 5 mins
 
-export const vercelApiBaseUrl =
-  process.env.REACT_APP_VERCEL_API_BASE_URL_OVERRIDE || "";
```
dohaki, line 525: Hmm, why remove this? I found this to be very useful when developing locally

src/views/Bridge/components/ChainSelector.tsx
```diff
@@ -28,17 +33,64 @@ export function ChainSelector({
...
+    account,
+  });
+
+  const sortOrder = useMemo(() => {
+    const chains = allChains.map((c) => ({
+      ...c,
+      balance: balances?.[c.chainId] ?? BigNumber.from(0),
+      disabled: false,
+    }));
+    if (!balances || isLoading) {
```
dohaki, line 57: This is causing some weird behavior where the displayed chains switch between sorted and unsorted every time the query re-runs. To reproduce try to keep the chain selector modal open for a while and you observe that

dohaki (commented):
Left two remarks

pxrl:
@james-a-morris Great to see this - such a nice UX boost!

src/views/Bridge/components/ChainSelector.tsx
james-a-morris: I had `isLoading` for testing. I've removed it [here](https://github.com/across-protocol/frontend/pull/1242/commits/1b2ed83ca651b716823a8eb8df489a21a1fa47c2)

dohaki (approved):
Nice

PR #1244, opened 2024-10-11
improve: Support USDC on Arbitrum Sepolia

pxrl:
Requested by an integrator for testing.

### DIFF
--- a/src/data/chains_11155111.json
+++ b/src/data/chains_11155111.json
@@ -1,0 +1,334 @@
+[
+  {
+    "chainId": 11155111,
+    "name": "Sepolia",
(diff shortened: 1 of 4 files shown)


src/data/chains_11155111.json
pxrl, line 1: @dohaki I wasn't sure whether this is needed since it's a new file, but the generate-routes script produces it so I've added it here. wdyt?

dohaki: yea makes sense! I also added it here https://github.com/across-protocol/frontend/pull/1245

dohaki (approved):

PR #1247, opened 2024-10-15
fix: worldchain cron cache balances

dohaki:
Cron job was failing due to missing World Chain Mulicall3 address. Fixed via SDK update


pxrl (approved):

pxrl:
> Cron job was failing due to missing World Chain Mulicall3 address. Fixed via SDK update

Nice catch - thanks for being on top of this!

PR #1248, opened 2024-10-15
chore: Enable USDC.e on World Chain

pxrl:


pxrl:
I've tested this OK on my end - works OK to bridge USDC in and out of World Chain.

james-a-morris (approved):

PR #1254, opened 2024-10-18
improve handling of non eth chains

gsteenkamp89:
closes ACX-2874


scripts/generate-routes.ts
```diff
@@ -264,13 +265,8 @@ function transformChainConfigs(
...
 
-        // Handle WETH Polygon
-        if (
-          tokenSymbol === "WETH" &&
-          [CHAIN_IDs.POLYGON, CHAIN_IDs.POLYGON_AMOY].includes(
-            toChainConfig.chainId
-          )
-        ) {
+        // Handle WETH Polygon & other non-eth chains
+        if (tokenSymbol === "WETH" && !toChainConfig.tokens.includes("ETH")) {
```
gsteenkamp89, line 269: @dohaki is my assumption correct here? Can we equate not supporting ETH with not being an ETH native chain? Will we ever **not** support a native currency if it is ETH?

dohaki: yes I think you are correct

james-a-morris (approved):

PR #1255, opened 2024-10-18
fix: separate upstash redis env vars

dohaki:
I had to handle Upstash env vars separately because Vercel won't let me override the reserved system env vars without deleting the existing storage. Therefore using the approach of this PR


pxrl (approved):
LGTM!

PR #1259, opened 2024-10-23
feat(bridge page): estimated fees redesign

james-a-morris:


dohaki (commented):
Nice work! Looks awesome 👍 I found one bug though:

https://github.com/user-attachments/assets/fc77457d-7d2b-40c8-ae64-7faaa73dd997

james-a-morris:
> a205e09

Resolved

dohaki (commented):
Bug seems to be fixed! Here are some other discrepancies with the Figma design:
### Validation error text placement
In Figma
image
In preview
image


### USD decimals (NIT)
This is a nit! We always truncate USD values to 2 decimals. I was wondering if we could truncate to 3 or 4 decimals when the USD value is around 0? Below a screenshot where we show 0.00 USD for fees even though the fee is >0 USD. I just think it could be a little bit misleading because we don't show token values anymore.
image

dohaki (approved):
As said already this looks so good 👍 One last nit: if the fees for a previous input already loaded and I change the input, the loader overlay doesn't apply to the rows in the expanded fees box. But we can tackle that in a separate PR
image

gsteenkamp89 (approved):

james-a-morris:
> 8490d4b

https://github.com/across-protocol/frontend/pull/1259/commits/8490d4b4b9cbfd4d07aa24c7f56ee10dbe16ba5a

james-a-morris:
> As said already this looks so good 👍 One last nit: if the fees for a previous input already loaded and I change the input, the loader overlay doesn't apply to the rows in the expanded fees box. But we can tackle that in a separate PR image

https://github.com/across-protocol/frontend/pull/1259/commits/8490d4b4b9cbfd4d07aa24c7f56ee10dbe16ba5a

PR #1260, opened 2024-10-24
fix: avoid polygon from referencing native ETH

james-a-morris:


dohaki (approved):

src/hooks/useBalance.ts
```diff
@@ -157,44 +159,48 @@ export function useBalancesBySymbols({
...
         ReturnType<typeof balanceQueryKey>
       >
-    >((chainId) => ({
-      queryKey: balanceQueryKey(account, chainId, tokenSymbol),
+    >(({ chainId, symbol }) => ({
+      queryKey: balanceQueryKey(account, chainId, symbol),
       queryFn: ({ queryKey }) => {
         const [, chainIdToQuery, tokenSymbolToQuery, accountToQuery] = queryKey;
+        if (
+          utils.chainIsMatic(chainIdToQuery!) &&
```
pxrl, line 182: This is OK for the current set of supported chains but would probably break again for Aleph Zero, which is coming rsn cc @gsteenkamp89

pxrl (approved):

PR #1262, opened 2024-10-29
fix casing of zksync logo served from /chains

gsteenkamp89:
closes ACX-2707

- find the relative path of the logo in `chainConfig` and resolves it. testnet chains do not have duplicate assets so this will direct base-sepolia to base/assets/logo.svg. 
- directories for multi word names use dashes (-) as separator, while we export the chainConfig map keys with underscores. when resolving the path we need to replace underscores with dashes. this fixes the zksync logo resolution

### DIFF
--- a/src/components/AmountInput/AmountInput.tsx
+++ b/src/components/AmountInput/AmountInput.tsx
@@ -66,6 +66,7 @@ export function AmountInput({
         ) : null}
         <Input
           type="number"
+          inputMode="numeric"
           valid={isAmountValid}
           placeholder="Enter amount"
           value={amountInput}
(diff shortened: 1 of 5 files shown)


src/components/AmountInput/AmountInput.tsx
```diff
@@ -66,6 +66,7 @@ export function AmountInput({
         ) : null}
         <Input
           type="number"
+          inputMode="decimal"
```
dohaki, line 69: This breaks unfortunately on mobile for german keyboard.

gsteenkamp89: oh really that sucks. i'll remove it. it's out of scope anyway 😅

dohaki (approved):

PR #1266, opened 2024-11-04
feat(multicall): add for alephzero and zksync

james-a-morris:
Add AlephZero multicall address. Also added ZKSync since it does not follow the standard Multicall3 address call


dohaki (approved):
Looks good! As an FYI, we seem to have two different ways of resolving a multicall3 address. The function covered in this PR is only used for L1/Mainnet multicalls. But we also use multicall here https://github.com/across-protocol/frontend-v2/blob/a2a9e4a4263cac6eac3085674ff650c6e85c62c3/api/_utils.ts#L1057. But it relies on the correct addresses upstream in the SDK https://github.com/across-protocol/sdk/blob/b43acb896801647dce6bdea82dfad90b908f7224/src/utils/Multicall.ts#L19

PR #1268, opened 2024-11-08
feat: uniswap v3 exact output support + UniversalSwapAndBridge

dohaki:
- Adds generic helper method that allows us to query a Uniswap quote for `EXACT_OUTPUT`
- Migrates existing `SwapAndBridge` logic to newly deployed `UniversalSwapAndBridge` contracts

Closes ACX-3280


scripts/generate-routes.ts
```diff
@@ -91,16 +91,16 @@ const enabledRoutes = {
     },
     swapAndBridgeAddresses: {
       "1inch": {
-        [CHAIN_IDs.POLYGON]: "0xaBa0F11D55C5dDC52cD0Cb2cd052B621d45159d5",
-        [CHAIN_IDs.OPTIMISM]: "0x3E7448657409278C9d6E192b92F2b69B234FCc42",
-        [CHAIN_IDs.ARBITRUM]: "0xC456398D5eE3B93828252e48beDEDbc39e03368E",
-        [CHAIN_IDs.BASE]: "0x7CFaBF2eA327009B39f40078011B0Fb714b65926",
+        [CHAIN_IDs.POLYGON]: "0xf9735e425a36d22636ef4cb75c7a6c63378290ca",
```
bmzig, line 94: This address doesn't look checksummed. Is that fine?

bmzig (approved):
I think this makes sense. `AlphaRouter` seems really handy.

scripts/generate-routes.ts
dohaki: good point! should be fine but for consistency I changed

PR #1272, opened 2024-11-13
get token details for any address in any chainId

gsteenkamp89:
closes ACX-3284

While calculating swap routes using uniswap's sdk, we need to find the token symbol and decimals for any token address on any chain,
Since there is no on-chain way of getting the token symbol for a native token, we need to use an external api. we can use that exposed by chainlist.org and cache any values for a pretty long ttl.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -2003,7 +2003,7 @@ const TTL_TOKEN_INFO = 30 * 24 * 60 * 60; // 30 days
 
 function tokenInfoCache(params: TokenOptions) {
   return makeCacheGetterAndSetter(
-    buildInternalCacheKey("tokenInfo", ...Object.values(params)),
+    buildInternalCacheKey("tokenInfo", params.address, params.chainId),
     TTL_TOKEN_INFO,
     () => getTokenInfo(params),
     (tokenDetails) => tokenDetails
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -54,7 +54,7 @@ import {
   maxRelayFeePct,
   relayerFeeCapitalCostConfig,
 } from "./_constants";
-import { PoolStateOfUser, PoolStateResult } from "./_types";
+import { PoolStateOfUser, PoolStateResult, TokenInfo } from "./_types";
 import {
   buildInternalCacheKey,
   getCachedValue,
@@ -65,6 +65,7 @@ import {
   InvalidParamError,
   RouteNotEnabledError,
 } from "./_errors";
+import { ZERO_ADDRESS } from "@across-protocol/sdk/dist/esm/constants";
 
 export { InputError, handleErrorCondition } from "./_errors";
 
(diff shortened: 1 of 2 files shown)


api/_utils.ts
```diff
@@ -1992,3 +1992,72 @@ export function paramToArray<T extends undefined | string | string[]>(
...
+    const provider = getProvider(chainId);
+
+    const erc20 = ERC20__factory.connect(
+      ethers.utils.getAddress(address),
+      provider
+    );
+
+    const [decimals, symbol] = await Promise.all([
+      erc20.decimals(),
+      erc20.symbol(),
```
dohaki, line 2047: we should get `name` as well

api/_utils.ts
```diff
@@ -1992,3 +1992,72 @@ export function paramToArray<T extends undefined | string | string[]>(
...
+
+    // ERC20 resolved dynamically
+    const provider = getProvider(chainId);
+
+    const erc20 = ERC20__factory.connect(
+      ethers.utils.getAddress(address),
+      provider
+    );
+
+    const [decimals, symbol] = await Promise.all([
```
pxrl, line 2045: There's admittedly limited benefit because of the long cache duration, but is it worth bundling this (+ erc20.name()) into a multicall aggregate call?

api/_utils.ts
```diff
@@ -1992,3 +1992,72 @@ export function paramToArray<T extends undefined | string | string[]>(
...
+      symbol,
+    };
+  } catch (err) {
+    throw new Error(
+      `Unable to find tokenDetails for address: ${address}, on chain with id: ${chainId}`,
+      {
+        cause: err instanceof Error && err,
+      }
+    );
+  }
```
pxrl, line 2062: `api/_errors.ts` defines a bunch of errors that can be thrown elsewhere - it'd be nice to slot this definition over there and follow the pattern that's established for passing errors back to the caller.

pxrl (commented):
> Since there is no on-chain way of getting the token symbol for a native token, we need to use an external api. we can use that exposed by chainlist.org and cache any values for a pretty long ttl.

Are we OK to revert to the definitions that we have in constants for this? I _think_ it should be OK in that we know we have the definitions (We can resolve chain -> gasToken symbol via `constants/networks`, and then we can resolve the token information from the gasToken symbol by lookup in `constants/tokens`.

api/_utils.ts
gsteenkamp89: We could do that, but as you say, probably not much benefit since we're caching this for quite a while

api/_utils.ts
```diff
@@ -1992,3 +1993,89 @@ export function paramToArray<T extends undefined | string | string[]>(
...
+export type TokenOptions = {
+  chainId: number;
+  address: string;
+};
+
+const TTL_TOKEN_INFO = 30 * 24 * 60 * 60; // 30 days
+
+function tokenInfoCache(params: TokenOptions) {
+  return makeCacheGetterAndSetter(
+    buildInternalCacheKey("tokenInfo", ...Object.values(params)),
```
dohaki, line 2006: NIT: should we make this positional and explicit args? like
```suggestion
    buildInternalCacheKey("tokenInfo", params.address, params.chainId),
```
This would prevent duplicate records in the cache for 
```js
getCachedTokenInfo({ chainId, address })
// below would create separate cache key with current implementation
getCachedTokenInfo({ address, chainId })
```

api/_utils.ts
```diff
@@ -1992,3 +1993,89 @@ export function paramToArray<T extends undefined | string | string[]>(
...
+      address,
+      decimals,
+      symbol,
+      name,
+    };
+  } catch (error) {
+    throw new TokenNotFoundError({
+      chainId,
+      address,
+    });
```
dohaki, line 2079: NIT: Would be helpful to be able to distinguish whether there was a revert or the provided inputs really are wrong 🤔 Or at least if we could pass the thrown error as `cause`

dohaki (approved):
LGTM

api/_utils.ts
gsteenkamp89: I've updated this function to validate the 2 params, throw input errors.
Also added the root error as cause, as you suggested.
If the address just doesn't exist on the given chain, then any reverts from the provider will be caught and added to cause.

PR #1278, opened 2024-11-18
Feat/support arb tokens on coingecko endpoint

gsteenkamp89:
related to ACX-3283

- allows caller to specify chainId which we'll pass to CG client. CG client is responsible for resolving the platformId based on the chainId. we default to mainnet, as before.


dohaki (approved):
Nice!

PR #1281, opened 2024-11-21
feat: add allowance and balance check

dohaki:
Closes ACX-3365

- Moves `/swap` to route `/swap/allowance`
- Add balance and allowance check to above
- Adds support for `SpokePoolVerifier`


api/_dexes/cross-swap.ts
```diff
@@ -216,10 +202,23 @@ export async function buildCrossSwapTx(
       }
     );
     toAddress = swapAndBridge.address;
-  } else {
-    const spokePool = getSpokePool(
-      crossSwapQuotes.crossSwap.inputToken.chainId
+  } else if (isInputNative && spokePoolVerifier) {
+    tx = await spokePoolVerifier.populateTransaction.deposit(
```
pxrl, line 206: Hm - one thing to be aware of is that we can't issue deposit refunds in this case. We need to deploy new SpokePoolVerifier contracts for that. Audit is wrapping up so we should be able to do that shortly.

api/_dexes/cross-swap.ts
```diff
@@ -242,23 +241,39 @@ export async function buildCrossSwapTx(
...
+    depositor: crossSwapQuotes.crossSwap.depositor,
+    recipient:
+      message && message !== "0x"
+        ? getMultiCallHandlerAddress(destinationChainId)
+        : crossSwapQuotes.crossSwap.recipient,
+    inputToken: crossSwapQuotes.bridgeQuote.inputToken.address,
+    outputToken: crossSwapQuotes.bridgeQuote.outputToken.address,
+    inputAmount: crossSwapQuotes.bridgeQuote.inputAmount,
+    outputAmount: crossSwapQuotes.bridgeQuote.outputAmount,
+    destinationChainid: crossSwapQuotes.bridgeQuote.outputToken.chainId,
```
pxrl, line 266:
```suggestion
    destinationChainid: crossSwapQuotes.bridgeQuote.outputToken.chainId,
```

We also have `destinationChainId = crossSwapQuotes.crossSwap.outputToken.chainId`. Is it worth asserting that these are equal? Then we could just use `destinationChainId` here.

api/_dexes/cross-swap.ts
```diff
@@ -242,23 +241,39 @@ export async function buildCrossSwapTx(
...
+  const originChainId = crossSwapQuotes.crossSwap.inputToken.chainId;
+  const destinationChainId = crossSwapQuotes.crossSwap.outputToken.chainId;
+  const spokePool = getSpokePool(originChainId);
+  const message = crossSwapQuotes.bridgeQuote.message || "0x";
+  const deposit = {
+    depositor: crossSwapQuotes.crossSwap.depositor,
+    recipient:
+      message && message !== "0x"
+        ? getMultiCallHandlerAddress(destinationChainId)
+        : crossSwapQuotes.crossSwap.recipient,
```
pxrl, line 261: I don't think `message` can ever be undefined, and we have a utility function in the SDK that centralises the "message is empty" check:

```suggestion
    recipient: isMessageEmpty(message)
        ? crossSwapQuotes.crossSwap.recipient
        : getMultiCallHandlerAddress(destinationChainId),
```

api/_erc20.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export async function getBalance(params: {
+  chainId: number;
+  tokenAddress: string;
+  owner: string;
+}) {
+  const erc20 = getErc20(params);
+  return erc20.balanceOf(params.owner);
+}
```
pxrl, line 21: Since these are queried in parallel by the caller,  do you think it could make sense to multicall them?

api/swap/_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+    ? getWrappedNativeTokenAddress(destinationChainId)
+    : utils.getAddress(_outputTokenAddress);
+
+  if (!_minOutputAmount && !_exactInputAmount && !_exactOutputAmount) {
+    throw new MissingParamError({
+      param: "minOutputAmount, exactInputAmount, exactOutputAmount",
+      message:
+        "One of 'minOutputAmount', 'exactInputAmount' or 'exactOutputAmount' is required",
+    });
+  }
```
pxrl, line 77: Should we also enforce that these are mutually exclusive? If not then the behaviour that the caller sees is implementation-defined, based on the order of evaluation down at line 105.

api/swap/allowance.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+      | undefined;
+    if (allowance.lt(inputAmount)) {
+      allowanceTx = {
+        to: crossSwapQuotes.crossSwap.inputToken.address,
+        chainId: originChainId,
+        from: crossSwapQuotes.crossSwap.depositor,
+        data: encodeApproveCalldata(crossSwapTx.to, inputAmount),
+      };
+    }
```
pxrl, line 71: Do we also need to consider quirks like USDT, where a reset to 0 is required first? Iirc there's a utility function that can resolve true/false for this requirement based on the token address and chain ID.

api/swap/allowance.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const isSwapTxEstimationPossible =
+      !skipOriginTxEstimation && !allowanceTx && balance.gte(inputAmount);
```
pxrl, line 74: It's a bit unfortunate that the depositor needs to have an approval at least as large as their expected swap/bridge amount in order to determine whether they can do this at all. I can't think of anything off the top of my head to ease this without forking during simulation, but it'd be super nice if we could figure out how to bundle the approval into the simulation (or make it unnecessary). Might be something to think about.

pxrl (commented):
Only minor comments - this looks good.

api/_dexes/cross-swap.ts
bmzig: Also, is it safe to assume that if we have a non-empty message, then the recipient is the multicall handler, couldn't an integrator implement their own version of `HandleV3AcrossMessage`?

api/_dexes/cross-swap.ts
dohaki: ok yea that would be good if we can redeploy.

api/_dexes/cross-swap.ts
dohaki: @bmzig I think in general yes. But for the initial scope of this swap endpoint, we should be able to assume that API consumers won't use their own handler contract. Because we can not do the swapping for them then

api/swap/_utils.ts
dohaki: Yea.. Or we could require the caller to specify the type + amount like @gsteenkamp89 suggested here https://github.com/across-protocol/frontend/pull/1281#discussion_r1851984703. I think that would be the cleanest

api/swap/allowance.ts
dohaki: Hmm.. yea that's a good call. I will add a todo here and tackle that separately

dohaki: After thinking a bit more about this, maybe a better option would be to omit the approval transaction? We can provide the data but it should be the consumer in the end who has to handle the approval-flow 🤔

api/swap/_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  const amountType = _minOutputAmount
+    ? AMOUNT_TYPE.MIN_OUTPUT
+    : _exactInputAmount
+      ? AMOUNT_TYPE.EXACT_INPUT
+      : AMOUNT_TYPE.EXACT_OUTPUT;
+  const amount = BigNumber.from(
+    _minOutputAmount || _exactInputAmount || _exactOutputAmount
+  );
```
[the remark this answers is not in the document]
dohaki: Yea that would be the cleanest option probably.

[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/frontend/pull/1281/commits/f37db730caafdcab11ac8dba79ad12c17f3d7cd9

api/swap/_utils.ts
```diff
@@ -0,0 +1,125 @@
...
+  originChainId: positiveIntStr(),
+  destinationChainId: positiveIntStr(),
+  depositor: validAddress(),
+  recipient: optional(validAddress()),
+  integratorId: optional(string()),
+  refundAddress: optional(validAddress()),
+  refundOnOrigin: optional(boolStr()),
+  slippageTolerance: optional(positiveFloatStr(50)), // max. 50% slippage
+  skipOriginTxEstimation: optional(boolStr()),
+});
```
gsteenkamp89, line 35: This looks good, I think in subsequent PRs we can refine this schema even more to be a union discriminated  by trade type. This should make validation a bit more straightforward

```ts
const schema = union([
   object({
      tradeType: literal('minOutput'),
      ...minOutputParams
   }),
   object({
      tradeType: literal('exactOutput'),
      ...exactOutputParams
   })
])
```

gsteenkamp89 (approved):
Looks good, just left one comment but we can address that later i think

pxrl (approved):
LGTM!

PR #1284, opened 2024-11-22
feat: refund on origin flag

dohaki:
Closes ACX-3322


api/_dexes/utils.ts
```diff
@@ -162,3 +160,9 @@ export function buildMinOutputBridgeTokenMessage(
     ],
   });
 }
+
+export function getFallbackRecipient(crossSwap: CrossSwap) {
+  return crossSwap.refundOnOrigin
+    ? constants.AddressZero
+    : crossSwap.refundAddress ?? crossSwap.depositor;
```
dohaki, line 167: Maybe we should remove this query param for now until we have a custom MultiCallHandler

gsteenkamp89 (approved):

api/_dexes/utils.ts
bmzig, line 167: Should it be `crossSwap.refundAddress ?? crossSwap.depositor` or `crossSwap.refundAddress ?? crossSwap.recipient`? Not sure if the recipient is the same as the depositor in this context.

bmzig: Sorry, somewhat irrelevant, but is it `crossSwap.depositor` here since `crossSwap.recipient` is set as the multicall handler address?

PR #1285, opened 2024-11-22
Gerhard/acx 3321 calculate fee components

gsteenkamp89:


api/swap/_utils.ts
```diff
@@ -121,5 +125,102 @@ export async function handleBaseSwapQueryParams({
...
+    bridgeFeePromise,
+    originSwapFeePromise,
+    destinationSwapFeePromise,
+  ]);
+
+  return {
+    bridgeFees,
+    originSwapFees,
+    destinationSwapFees,
+  };
```
gsteenkamp89, line 225: Should we revert to fees in input token if coingecko fails tor esolve prices in usd?

dohaki: Yea would be nice if possible. But we can tackle in a separate PR

api/swap/_utils.ts
```diff
@@ -121,5 +127,102 @@ export async function handleBaseSwapQueryParams({
...
+);
+
+async function calculateSwapFee(
+  swapQuote: SwapQuote,
+  baseCurrency: string
+): Promise<Record<string, number>> {
+  const { tokenIn, tokenOut, expectedAmountOut, expectedAmountIn } = swapQuote;
+  const [[, inputTokenPriceBase], [, outputTokenPriceBase]] = await Promise.all(
+    [
+      coingeckoClient.getCurrentPriceByContract(
```
dohaki, line 155: I noticed that this will return 0 if the price can not be looked up on CG... For example, if I want to query the price of APE on Arbitrum this returns 0 because the platform doesn't know of the contract even though it exists.

We probably need some mapping or fallback where we can try to look up based on the token symbol -> CG id. And we should ask integrators of this endpoint to provide us a list of used tokens so that we can add them to this mapping

dohaki (approved):
Left some remarks. They might not be directly actionable in this PR though

PR #1290, opened 2024-11-25
improve: swap deposit exclusive with depositV3

bmzig:
Once the spoke pools are upgraded with the most recent audit changes, we will want to stop using deposit exclusive and start using depositV3. This also means we need to pass in the exclusivity parameter as either an offset or a timestamp, depending on whether the origin chain re-orgs.


dohaki:
Closing in favor of https://github.com/across-protocol/frontend/pull/1383

PR #1294, opened 2024-11-27
feat: use trading api + support universal router

dohaki:
Closes ACX-3382

Sorry for the big PR but had to refactor quite a lot to support the Trading API and `UniversalRouter`. This PR allows the usage of different quote resolving + tx building strategies. We also make use the trading API for improved performance.

Note that it is possible to use the Uniswap API to retrieve a quote and use the old SwapRouter02 implementation to build the respective tx. One problem we need to address is the requirement for permit2 support when using the UniversalRouter contract.

### DIFF
--- a/api/_dexes/cross-swap.ts
+++ b/api/_dexes/cross-swap.ts
@@ -14,7 +14,7 @@ import {
   getUniswapCrossSwapQuotesForOutputA2B,
   getUniswapCrossSwapQuotesForOutputB2A,
 } from "./uniswap/quote-resolver";
-import { getUniversalRouterStrategy } from "./uniswap/universal-router";
+import { getSwapRouter02Strategy } from "./uniswap/swap-router-02";
 import { UniswapQuoteFetchStrategy } from "./uniswap/utils";
 import { CrossSwap, CrossSwapQuotes } from "./types";
 import {
@@ -54,7 +54,7 @@ export const LEFTOVER_TYPE = {
 export const PREFERRED_BRIDGE_TOKENS = ["WETH"];
 
 const defaultQuoteFetchStrategy: UniswapQuoteFetchStrategy =
-  getUniversalRouterStrategy();
+  getSwapRouter02Strategy();
 const strategyOverrides = {
   [CHAIN_IDs.BLAST]: defaultQuoteFetchStrategy,
 };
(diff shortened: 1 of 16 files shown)


scripts/generate-routes.ts
```diff
@@ -116,6 +116,10 @@ const enabledRoutes = {
         [CHAIN_IDs.ZORA]: "0x8EB5FF2e23FD7789e59989aDe055A398800E394e",
         [CHAIN_IDs.MAINNET]: "0x8EB5FF2e23FD7789e59989aDe055A398800E394e",
       },
+      "uniswap-universalRouter": {
+        [CHAIN_IDs.OPTIMISM]: "0xaED9bBFdCC63219d77D54F8427aeb84C2Be46c5f",
+        [CHAIN_IDs.ARBITRUM]: "0xaED9bBFdCC63219d77D54F8427aeb84C2Be46c5f",
+      },
```
dohaki, line 122: Note that we can not use these currently bc we need to support permit2 in the periphery contract when calling a dex

api/_dexes/cross-swap.ts
```diff
@@ -48,7 +51,13 @@ export const LEFTOVER_TYPE = {
...
 } as const;
 
-export const PREFERRED_BRIDGE_TOKENS = ["WETH", "USDC"];
+export const PREFERRED_BRIDGE_TOKENS = ["WETH"];
+
+const defaultQuoteFetchStrategy: UniswapQuoteFetchStrategy =
+  getSwapRouter02Strategy();
+const strategyOverrides = {
+  [CHAIN_IDs.BLAST]: defaultQuoteFetchStrategy,
+};
```
gsteenkamp89, line 60: what dictates what strategy to use for a chain?

api/_utils.ts
```diff
@@ -201,6 +201,7 @@ export const getLogger = (): LoggingUtility => {
  * @returns A valid URL of the current endpoint in vercel
  */
 export const resolveVercelEndpoint = () => {
+  return "https://app.across.to";
```
gsteenkamp89, line 204: do we want to keep this?

api/_dexes/uniswap/trading-api.ts
```diff
@@ -0,0 +1,139 @@
...
+    {
+      headers: {
+        "x-api-key": UNISWAP_API_KEY,
+      },
+    }
+  );
+  return response.data;
+}
+
+export async function getUniswapClassicIndicativeQuoteFromApi(
```
gsteenkamp89, line 73: What is the purpose of an `"indicative quote"`? and what determines this?

gsteenkamp89 (approved):
Just have a few questions for my own understanding, but otherwise LGTM!

api/_dexes/cross-swap.ts
dohaki: Some chains we support might not have Uniswap deployed. And we'd need to eventually use a different strategy, I.e. Aggregator/Dex.

In the context of Uniswap, I thought that the `UniversalRouter` is only deployed to a subset of chains if compared to `SwapRouter02`. So I thought we'd need to use different strategies as well. That's why added this refactor here. But for initial launch we won't need multiple strategies AFAICT.

gsteenkamp89: That makes sense, ty!

api/_dexes/uniswap/trading-api.ts
gsteenkamp89: Aah wow good find. the docs are not super clear on that afaict

PR #1301, opened 2024-12-02
measure swap/allowance performance

gsteenkamp89:
- measure full running time of swap/allowance endpoint (add requests data as well as `crossSwapQuotes`).
- measure running time `getCrossSwapQuotes`, we can separate these by `amountType` eg, "exactInput" | "exactOutput".
- measure time to get crossSwapQuote by crossSwapType. eg "b2b" | "b2a" etc.

### DIFF
--- a/api/_dexes/cross-swap.ts
+++ b/api/_dexes/cross-swap.ts
@@ -9,7 +9,6 @@ import {
   getBridgeQuoteForMinOutput,
   getSpokePool,
   Profiler,
-  getLogger,
 } from "../_utils";
 import {
   getBestUniswapCrossSwapQuotesForOutputA2A,
@@ -82,7 +81,7 @@ export async function getCrossSwapQuotes(
 export async function getCrossSwapQuotesForOutput(crossSwap: CrossSwap) {
   const profiler = new Profiler({
     at: "api/cross-swap#getCrossSwapQuotesForOutput",
-    logger: getLogger(),
+    logger: console,
   });
   const crossSwapType = getCrossSwapType({
     inputToken: crossSwap.inputToken.address,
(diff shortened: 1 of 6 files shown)


api/_dexes/cross-swap.ts
```diff
@@ -78,6 +80,10 @@ export async function getCrossSwapQuotes(
 }
 
 export async function getCrossSwapQuotesForOutput(crossSwap: CrossSwap) {
+  const profiler = new Profiler({
+    at: "api/cross-swap#getCrossSwapQuotesForOutput",
+    logger: getLogger(),
```
dohaki, line 85: Do we need to specify some log level here?

dohaki (commented):
Left a question

api/_dexes/cross-swap.ts
gsteenkamp89: I think this logger only accepts a global log level.. I might have to change this

dohaki (approved):

api/_dexes/cross-swap.ts
```diff
@@ -86,19 +91,35 @@ export async function getCrossSwapQuotesForOutput(crossSwap: CrossSwap) {
...
+    return await profiler.measureAsync(
+      getCrossSwapQuotesForOutputA2B(crossSwap),
+      "getCrossSwapQuotesForOutputA2B",
+      crossSwap
+    );
   }
 
   if (crossSwapType === CROSS_SWAP_TYPE.ANY_TO_ANY) {
-    return getCrossSwapQuotesForOutputA2A(crossSwap);
+    return await profiler.measureAsync(
```
james-a-morris, line 118: Do we need to run `return await`? You can usually omit the `await` because we're returning a promise anyway

james-a-morris (commented):
one question

api/_dexes/cross-swap.ts
gsteenkamp89: Yeah you're right this is not necessary!

PR #1304, opened 2024-12-02
improve: use deposit tracking from indexer in confirmation page

james-a-morris:


nicholaspai (approved):

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
```diff
@@ -48,6 +49,10 @@ export function DepositStatusUpperCard({
     fromBridgePagePayload
   );
 
+  void useIndexerDepositsTracking([
```
dohaki, line 52: TIL about the operator `void`! Never used it so far.

But just so that I understand: we void this here because we just want to load-test this hook right?

dohaki (commented):
Left a few questions!

src/views/DepositStatus/components/DepositStatusUpperCard.tsx
james-a-morris: Correct - we don't want the response (for now)

src/hooks/useIndexerDepositTracking.ts
```diff
@@ -17,12 +18,13 @@ export function useIndexerDepositsTracking(
...
+        deposit.depositTxnHash,
         deposit.originChainId,
         deposit.depositId,
-      ] as [string, number, number],
+      ] as [string, string, number, number],
       enabled:
-        isDefined(deposit.originChainId) &&
-        isDefined(deposit.depositId) &&
+        ((isDefined(deposit.originChainId) && isDefined(deposit.depositId)) ||
+          isDefined(deposit.depositTxnHash)) &&
```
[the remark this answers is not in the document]
james-a-morris: @dohaki nope - it is not a requirement

dohaki (approved):

PR #1307, opened 2024-12-03
chore: deprecate legacy fee response values

dohaki:


james-a-morris (commented):
Should we telegraph this to integrators?

gsteenkamp89 (dismissed):

dohaki:
> Should we telegraph this to integrators?

Yep in progress

gsteenkamp89 (approved):

james-a-morris (approved):
Approving so it won't block. LGTM only when the integrators are ready

PR #1309, opened 2024-12-03
fix: use more accurate ENS

james-a-morris:
We were previously relying on onboard to generate our ENS addresses. This is no longer reliable.

### DIFF
--- a/src/components/DepositsTable/cells/AddressCell.tsx
+++ b/src/components/DepositsTable/cells/AddressCell.tsx
@@ -5,6 +5,7 @@ import { Deposit } from "hooks/useDeposits";
 
 import { BaseCell } from "./BaseCell";
 import { shortenAddress } from "utils";
+import { useEnsQuery } from "hooks/useEns";
 
 type Props = {
   deposit: Deposit;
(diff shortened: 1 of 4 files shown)


dohaki (commented):
Shouldn't our dev wallet 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D resolve to some ENS?

image

james-a-morris:
> Shouldn't our dev wallet 0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D resolve to some ENS?
> 
> image

It seems as though in order for an ENS address to be resolved its [primary name (reverse registrar)](https://docs.ens.domains/registry/reverse#set-primary-name) must be set. [EthersJS Docs](https://docs.ethers.org/v5/api/providers/provider/#Provider-lookupAddress). This code snippet confirms that this is how EthersJS resolves it:

```
        const node = address.substring(2).toLowerCase() + ".addr.reverse";

        const resolverAddr = await this._getResolver(node, "lookupAddress");
        if (resolverAddr == null) { return null; }
```

I don't believe `0x91` has this set on its ENS registrar.

PR #1314, opened 2024-12-09
use only chain address to avoid reverts

gsteenkamp89:
some token addresses are not on some chains, the config client just inserts the l1 address into the `address` slot. when we go to check a users balance for a token on that chain we are currently using the mainnet address, causing a revert.


james-a-morris (approved):
fair call - does this not throw elsewhere?

gsteenkamp89:
> fair call - does this not throw elsewhere?

We can skip trying to fetch balances for a token that doesn't exist. It currently just a bit noisy in the console

PR #1315, opened 2024-12-09
improve: reduce edgecase on delayed deposits

james-a-morris:
Currently we have a case where if a deposit takes longer than 5 minutes and is larger than the max deposit instant then we display a 3-hour warning. This change updates the deposit to only mark delay if the deposit amount is larger than the max deposit short delay


gsteenkamp89 (approved):
LGTM

PR #1319, opened 2024-12-11
feat: relayer authentication submission method

dylanuma:


james-a-morris (changes requested):
Is this good to close?

PR #1322, opened 2024-12-11
feat(api): relayer config auth

md0x:


api/relayer-config.ts
```diff
@@ -0,0 +1,65 @@
...
+) => {
+  if (request.method === "GET") {
+    await handleGetRequest(request, response);
+  } else if (request.method === "POST") {
+    await handlePostRequest(request, response);
+  } else {
+    // Method not allowed
+    response.setHeader("Allow", ["GET", "POST"]);
+    response.status(405).end(`Method ${request.method} Not Allowed`);
+  }
```
james-a-morris, line 62:
```suggestion
  if(["GET","POST"].includes(request.method) {
    return request.method === "GET" ?  handleGetRequest(request, response) : handlePostRequest(request, response);
   } else {
     response.setHeader("Allow", ["GET", "POST"]);
     response.status(405).end(`Method ${request.method} Not Allowed`);
  }

Comment (james-a-morris) [reply api/relayer-config.ts:62]: we implicitly return a promise here - we can use the async nature of this function call and drop the async

File: api/_exclusivity/utils.ts [hunk 2]
```diff
@@ -0,0 +1,21 @@
+import { ethers } from "ethers";
+
+// TODO: get this from gh
+export const whiteListedRelayers = [
+  "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D", // dev wallet
+];
+
+export const getRelayerFromSignature = async (
```
pxrl, line 8: `verifyMessage()` isn't async; can we drop the `async` here ?

```suggestion
export const getRelayerFromSignature = (
```

api/relayer-config.ts
```diff
@@ -1,18 +1,70 @@
...
+} from "./_exclusivity/utils";
 import { TypedVercelRequest } from "./_types";
 
+export const MAX_MESSAGE_AGE_SECONDS = 300;
+
+const handleGetRequest = async (
+  request: TypedVercelRequest<any>,
+  response: VercelResponse
+) => {
+  const { signature, message } = request.query;
```
pxrl, line 15: @james-a-morris Would it make sense for `message` to be submitted in the message body? Otherwise we have to submit a lot of stringified json in the query string.

api/relayer-config.ts
```diff
@@ -1,18 +1,70 @@
...
+
+  const { timestamp, ...restOfMessage } = JSON.parse(message);
+  if (!isTimestampValid(timestamp, MAX_MESSAGE_AGE_SECONDS)) {
+    return response.status(400).json({ message: "Message too old" });
+  }
```
pxrl, line 20: I think the timestamp would need to be covered by the signature, otherwise it doesn't protect against replay.

api/relayer-config.ts
pxrl: Scrap that - this was on the get request.

james-a-morris: The message should be in the header field in the auth header

api/_types/exclusivity.types.ts
```diff
@@ -16,6 +17,16 @@ export const RelayerFillLimitSchema = object({
 
 export type RelayerFillLimit = Infer<typeof RelayerFillLimitSchema>;
 
+export type RelayerConfigUpdate = {
+  timestamp: number;
+  relayerFillLimits: RelayerFillLimit[];
+};
+
+export type TypedRelayerConfigUpdateRequest = TypedVercelRequest<
+  any,
```
james-a-morris, line 26:
```suggestion
  none,
```

api/relayer-config.ts
```diff
@@ -1,19 +1,49 @@
...
+  TypedVercelRequest,
+} from "./_types";
+
+const handlePostRequest = async (
+  request: TypedRelayerConfigUpdateRequest,
+  response: VercelResponse
+) => {
+  const body = request.body as RelayerConfigUpdate;
+  const { authorization } = request.headers;
+  const { timestamp, relayerFillLimits } = body;
```
james-a-morris, line 20:
```suggestion
  const { timestamp } = body;
```

api/relayer-config.ts
```diff
@@ -1,19 +1,49 @@
...
   response: VercelResponse
 ) => {
-  if (request.method === "GET") {
-    // Handle GET request
-    response.status(200).json({ message: "GET request received" });
-  } else if (request.method === "POST") {
-    // Handle POST request
-    response.status(200).json({ message: "POST request received" });
+  if (request.method === "POST") {
+    return handlePostRequest(request, response);
```
james-a-morris, line 43: Would probably make sense to internalize `handlePostRequest` in this function and have a check at the top for `request.method !== POST`;

james-a-morris (approved):

PR #1323, opened 2024-12-11
Add scaffolding for config unpacking

pxrl:


api/relayer-config.ts
```diff
@@ -52,7 +53,14 @@ const handlePostRequest = async (
     return response.status(401).json({ message: "Unauthorized" });
   }
 
-  // Handle POST request
+  if (!RelayerFillLimitsSchema.is(message)) {
+    return response
+      .status(400)
+      .json({ message: "Invalid configuration payload" });
+  }
```
james-a-morris, line 60: We should assert here - we should only send status in the try catch. See https://github.com/across-protocol/frontend-v2/blob/98f3e618c61607870aabc8a625570481413c955f/api/account-balance.ts

md0x (approved):

james-a-morris (approved):
No major blockers

PR #1331, opened 2024-12-12
Update format

pxrl:


james-a-morris (approved):
Fair call - so we're changing the typing such that we have a separate fill config + update config

md0x (approved):

PR #1334, opened 2024-12-17
chore: Drop max message length description

pxrl:
This encourages integrators to work around the limit. Better to have it fail and let them reach out.


james-a-morris (approved):
Good call. To push past Gitbook's caching we may have to make a blank commit there

PR #1339, opened 2024-12-19
improve(API): Apply gas mark up to base fee rather than gas cost

nicholaspai:
depends on https://github.com/across-protocol/sdk/pull/801

Fixes ACX3600

## Expected changes to API

See notes here about impact of bumping the `sdk` version: https://github.com/across-protocol/frontend/pull/1339/files#r1903169393

If we set the environment variables `BASE_FEE_MARKUP` and `PRIORITY_FEE_MARKUP` equal to the existing `GAS_MARKUP` values today then the gas fees should be quoted the exact same. We can fast follow by reducing these markups.

The main differences will be:
- ~7777777: Zora will be re-added as an OPStack chain which means the l1 gas cost will be added to all estimation. This will make zora quotes a lot more accurate. This is a good thing. Currently, zora quotes are under-estimating gas~ This is no longer a change since https://github.com/across-protocol/frontend/commit/3568cea23d3b6a7576657c1e4e4e1979d0d15992 was already merged before this

## QA Testing 

We can test this by querying `/api/gas-prices` against the same endpoint on the prod deployment. This will show roughly how the gas costs will change with prod.

### Comparing against production fee quoting API

Ultimately, what matters is the `gasFeeTotal` in the `/limits` end point and we should compare with prod. 
- For WETH: 
- query /limits for every single destination chain once on Prod API and once on Preview API
- Write down the gasFeeTotal and compare with Preview API
- Note any differences:
- If Prod API has higher gas fee total, then preview API will reduce gas fees but we need to make sure these gas fees are still deemed profitable by relayer
- If preview API has higher gas fee total, then that’s notable as well
- Can use the preview API from this PR https://github.com/across-protocol/frontend/pull/1349 to see the gas cost breakdown from the production API

### Comparing against Relayer

Relayer's will only fill deposits that leave enough profits to cover for destination chain gas costs. This means that the relayer's computed `tokenGasCosts`, including any scaling, should be less than the fee quoting API's `tokenGasCosts`, otherwise the relayer will consider the API's quotes as unprofitable. A simple way to check this is to check the relayer's log that includes the message "Updated gas cost" (in the "ProfitClient") that will list configs for each chain like:

```
"at": "ProfitClient",
"message": "Updated gas cost",
...
"1": {
      "nativeGasCost": "131167",
      "tokenGasCost": "961072227772860",
      "gasPrice": "7327088580",
      "scaledTokenGasCost": "276308265484697",
      "gasPadding": "1.15",
      "gasMultiplier": "0.25"
    },
...
```

I recommend using this relayer PR https://github.com/across-protocol/relayer/pull/1969 to print the above logs and compare with the quoted preview API's fees


james-a-morris (commented):
We should let integrators know about this prior to rolling out. This will change expected fees

api/_utils.ts
```diff
@@ -1972,7 +1974,10 @@ export async function getMaxFeePerGas(chainId: number): Promise<BigNumber> {
   }
   const { maxFeePerGas } = await sdk.gasPriceOracle.getGasPriceEstimate(
```
nicholaspai, line 1975: The main thing to check in this PR is to make sure that this gasPrice used by the /limits endpoint to estimate gas fees is not double-scaling any of the gas prices using the `baseFeeMultiplier`. Reviewers should review this [upstream SDK PR](https://github.com/across-protocol/sdk/pull/801) to trace the full computation from this `getMaxFeePerGas()` method call

gsteenkamp89: So essentially all that needs to change is ensuring the sdk handles computing the base fee multiplier and then applying it to the base fee here in /limits?

api/_utils.ts
```diff
@@ -589,14 +589,17 @@ export const getHubPoolClient = () => {
...
   if (typeof gasMarkup[chainId] === "number") {
     return gasMarkup[chainId];
   }
 
-  return sdk.utils.chainIsOPStack(Number(chainId))
-    ? gasMarkup[CHAIN_IDs.OPTIMISM] ?? DEFAULT_GAS_MARKUP
-    : DEFAULT_GAS_MARKUP;
+  return (
+    1 +
+    (sdk.utils.chainIsOPStack(Number(chainId))
```
nicholaspai, line 599: One difference here. In the current multiplier use case, the SDK adds the `gasMarkup` to 1 [here](https://github.com/across-protocol/sdk/blob/a7615d0c68d847cfd8a42c992df32390b5931e01/src/relayFeeCalculator/chain-queries/baseQuery.ts#L89).

Whereas, the new SDK change I propose assumes that the passed in multiplier is complete already and doesn't need to be added to anything

api/_utils.ts
```diff
@@ -1966,13 +1969,12 @@ export function latestGasPriceCache(chainId: number) {
  * @returns The gas price in the native currency of the chain
  */
 export async function getMaxFeePerGas(chainId: number): Promise<BigNumber> {
-  if (sdk.utils.chainIsOPStack(chainId)) {
```
nicholaspai, line 1969: Important LOC we're removing. We should use gas price oracle for all estimates since it handles OP stack now

nicholaspai:
Note: because this PR uses the SDK's GasPriceOracle to compute gas prices even for OpStack chains we should NOT MERGE this PR until we set the GAS_PRICE_EIP1559_RAW_<chainId> variables equal to true (and probably also the NEW_GAS_PRICE_ORACLE_59144 ). Otherwise the API will give higher gas price estimates than we use currently

api/gas-prices.ts
```diff
@@ -1,31 +1,139 @@
 import { VercelResponse } from "@vercel/node";
 import {
+  buildDepositForSimulation,
```
nicholaspai, line 3: This file has a lot of changes but its an API endpoint we don't depend on in production as its mostly here to compare against the current existing API to compare the expected gas total changes. The main changes are:
- Adding gas costs (native and token gas cost) to the API response
- breaking down gas price by base fee and priority fee

api/limits.ts
```diff
@@ -194,6 +194,8 @@ const handler = async (
         tokenPriceNative,
         relayer,
         gasUnits,
+        // !!gas price should be defined and passed into getRelayerFeeDetails so we don't recompute using default
```
nicholaspai, line 197: I think its important to highlight the importance of not letting gasPrice be undefined here, otherwise the effects of the `baseFeeMultiplier`, based on the developer-set GAS_MARKUP will be discarded

package.json
```diff
@@ -7,7 +7,7 @@
     "@across-protocol/constants": "^3.1.24",
     "@across-protocol/contracts": "^3.0.19",
     "@across-protocol/contracts-v3.0.6": "npm:@across-protocol/contracts@3.0.6",
-    "@across-protocol/sdk": "^3.3.27",
```
nicholaspai, line 10: This SDK change has two important changes that will materially impact the `gasFeeTotal` value returned in the API response of the `/limits` (and `/suggested-fees`) endpoint:
- Zora chain gets [re-added](https://github.com/across-protocol/sdk/commit/1249377b565c2ffa2f40923502df61c06c8a9716) back to [OP_STACK family](https://github.com/across-protocol/constants/commit/a6129368ef538c4cdb76426af9a842ac16a4f58c) which means the fees will go up since OPStack chains [add an L1GasCost component](https://github.com/across-protocol/sdk/blob/291212bb7055d5cb6d8e6724dcc3b631702e8792/src/relayFeeCalculator/chain-queries/baseQuery.ts#L150)
- `GAS_MARKUP` will be applied to base fee, not total token gas cost, which will cut many gas fee totals down significantly because base fees are much smaller than priority fees on most chains, for example most OP Stack chains

api/_utils.ts
```diff
@@ -671,10 +708,10 @@ export const getRelayerFeeDetails = async (
     recipientAddress: string;
     message?: string;
   },
-  tokenPrice?: number,
-  relayerAddress?: string,
-  gasUnits?: sdk.utils.BigNumberish,
-  gasPrice?: sdk.utils.BigNumberish
+  tokenPrice: number,
+  relayerAddress: string,
+  gasPrice: sdk.utils.BigNumberish,
```
nicholaspai, line 713: Made gasPrice, tokenPrice and relayerAddress all required since they're never set undefined in this repo and its better to be more constricting on this type where possible

api/limits.ts
```diff
@@ -193,8 +193,9 @@ const handler = async (
         depositArgs,
         tokenPriceNative,
         relayer,
-        gasUnits,
-        gasPrice
+        gasPrice,
+        gasCosts?.nativeGasCost,
+        gasCosts?.tokenGasCost
```
nicholaspai, line 198: In other words, the change here is: load both nativeGasCost and tokenGasCost from the SDK and pass them both here when computing gasFeePercent. This is useful because the function `getCachedFillGasUsage()` call uses the `opStackL1GasCostMultiplier` when computing the `tokenGasCost` which will give us more configuration options when quoting gas costs.

We now can play with the following configs:
- `baseFeeMultiplier` set in `latestGasPriceCache`
- `priorityFeeMultiplier` set in `latestGasPriceCache`
- `opStackL1GasCostMultiplier` set in `getCachedFillGasUsage`

We take the resultant gas costs and gas prices and pass them into `relayerFeeDetails` which simply does some simple multiplication and division to return the fee percentages. There should be no more RPC queries in this `getRelayerFeeDetails()` call

gsteenkamp89 (approved):

james-a-morris (approved):
I tested the endpoint and ran the FE - the values look to be expected.

PR #1341, opened 2024-12-20
support transfer with auth

gsteenkamp89:


api/_transfer-with-auth.ts
```diff
@@ -0,0 +1,137 @@
...
+  return {
+    domainSeparator,
+    eip712: {
+      types: {
+        EIP712Domain: [
+          { name: "name", type: "string" },
+          { name: "version", type: "string" },
+          { name: "chainId", type: "uint256" },
+          { name: "verifyingContract", type: "address" },
+        ],
```
dohaki, line 103: This can be removed actually. Ethers throws an error for some reason because this is already defined internally when calling `signTypedData`

api/relay/_utils.ts
```diff
@@ -63,12 +67,33 @@ export const SwapAndDepositWithPermitArgsSchema = type({
   deadline: positiveIntStr(),
 });
 
+export const DepositWithAuthArgsSchema = type({
+  signatureOwner: validAddress(),
+  depositData: DepositDataSchema,
+  validAfter: positiveIntStr(),
+  validBefore: positiveIntStr(),
+  nonce: hexString(),
```
dohaki, line 75: Do we need to enforce a certain length here?

scripts/tests/swap-auth.ts
```diff
@@ -0,0 +1,69 @@
+import { Wallet } from "ethers";
+
+import { getProvider } from "../../api/_utils";
+import { fetchSwapQuote, SWAP_API_BASE_URL } from "./_swap-utils";
+import { buildAuthTxPayload } from "../../api/swap/auth/_utils";
+import axios from "axios";
+
+type AuthPayload = Awaited<ReturnType<typeof buildAuthTxPayload>>;
+
+async function swapWithPermit() {
```
dohaki, line 10: Did you test a relay via auth already? Just out of curiosity

dohaki (commented):
Looks great! Just a few remarks and questions

api/_transfer-with-auth.ts
gsteenkamp89: ah okay will remove!

api/relay/_utils.ts
gsteenkamp89: yes probably a good idea! will add a modifier for this validator

scripts/tests/swap-auth.ts
gsteenkamp89: Not yet, got rate limited by CG on first attempt. will address your comments and try again. 👍

dohaki (approved):
LGTM

PR #1344, opened 2024-12-23
use fill deadline buffer from env

gsteenkamp89:
closes ACX-3511


src/utils/bridge.ts
```diff
@@ -237,6 +238,8 @@ export async function sendDepositV3Tx(
   );
   fillDeadline ??= await getFillDeadline(spokePool);
 
+  console.log("###", fillDeadline);
+
```
dohaki, line 242:
```suggestion
```

dohaki (approved):
One nit! but looks good otherwise

PR #1346, opened 2024-12-26
feat: gelato and local signers

dohaki:
Closes ACX-3568, ACX-3569 and ACX-3570

This PR builds on top of #1345 and implements two relay strategies:
- Gelato
- Local Signers

The high-level lifecycle of a relay request is implemented as follows:
1. User sends request via `POST /relay`
    - validates body
    - if valid publishes a message to a chain- and strategy-specific queue (currently we use [QStash](https://upstash.com/docs/qstash/overall/getstarted))
    - returns `requestHash`
2. QStash processes messages by sending to `POST /relay/jobs/process`
    - makes sure to retry messages, takes care of parallelism, etc.
    - uses configured strategy to relay tx, i.e. via Gelato or local signers
3. User can request status via `GET /relay/status?requestHash`
   - we can move this logic into `POST /relay` as well if the user wants a long-running atomic request


api/relay/_utils.ts
```diff
@@ -5,11 +5,17 @@ import { hexString, positiveIntStr, validAddress } from "../_utils";
...
   getDepositTypedData,
   getSwapAndDepositTypedData,
 } from "../_spoke-pool-periphery";
+import { RelayRequest } from "./_types";
+import { redisCache } from "../_cache";
 
-export const GAS_SPONSOR_ADDRESS = "0x0000000000000000000000000000000000000000";
+export const GAS_SPONSOR_ADDRESS =
+  process.env.GAS_SPONSOR_ADDRESS ??
+  "0x0000000000000000000000000000000000000000";
```
bmzig, line 18: Is there any reason to not have this default to, for example, 0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67? Is this the address only used to get kickbacks from executing transactions?

bmzig (approved):
No major comments from me. I see there are still some TODOs, but not sure if you want to get them addressed in this PR or a follow-up.

api/relay/_utils.ts
dohaki: Yea setting to a default address like our relayer makes sense 👍 And yes this address is only used to get kickbacks from executing transactions.

api/relay/jobs/process.ts
```diff
@@ -0,0 +1,127 @@
...
+      to,
+      methodNameAndArgs,
+      signatures,
+    };
+
+    // Get cached request
+    const cachedRequest = await getCachedRelayRequest(relayRequest);
+
+    if (
+      !cachedRequest ||
```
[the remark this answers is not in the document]
dohaki: The error message is probably not explicit enough. But we only want to throw if the request is unknown or succeeded already. There are cases where the transaction gets the status `failure` but might succeed at a later point.

dohaki:
> No major comments from me. I see there are still some TODOs, but not sure if you want to get them addressed in this PR or a follow-up.

Thanks for reviewing! Yea the TODOs will be handled in follow up PRs. I opened separate tickets for them as well

api/relay/_queue.ts
```diff
@@ -0,0 +1,46 @@
...
+  const strategyName = strategy.strategyName;
+  const queue = getRelayRequestQueue(strategyName, request.chainId);
+  await queue.upsert({
+    parallelism: strategy.queueParallelism,
+  });
+
+  const baseUrl = resolveVercelEndpoint(true);
+  const response = await queue.enqueueJSON({
+    retries: 3,
+    contentBasedDeduplication: true,
```
gsteenkamp89, line 26: does this solve nonce collision?

api/relay/_queue.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function pushRelayRequestToQueue({
+  request,
+  strategy,
+}: {
+  request: RelayRequest;
+  strategy: RelayStrategy;
+}) {
+  const strategyName = strategy.strategyName;
+  const queue = getRelayRequestQueue(strategyName, request.chainId);
```
gsteenkamp89, line 18: Is each queue/strategy using different EOAs? If not, how do we avoid nonce collisions between txns submitted within the same block between these queues?

api/relay/_strategies/gelato.ts
```diff
@@ -0,0 +1,92 @@
...
+            `Can not relay request via Gelato due to task state ${taskStatus.taskState}`
+          );
+        }
+
+        if (taskStatus.transactionHash) {
+          txHash = taskStatus.transactionHash;
+          break;
+        }
+
+        await new Promise((resolve) => setTimeout(resolve, 1_000));
```
gsteenkamp89, line 40: nit: do you think there's value in setting the polling interval to the block time for the chain? or maybe make this an ENV?

api/relay/_strategies/local-signers.ts
```diff
@@ -0,0 +1,92 @@
...
+        const wallet = new Wallet(signerPrivateKey, provider);
+        try {
+          await lockSigner(wallet.address, request.chainId);
+
+          const balance = await wallet.getBalance();
+          if (balance.lt(balanceAlertThreshold)) {
+            // TODO: Send PD alert
+          }
+
+          const txRequest = {
```
gsteenkamp89, line 36: would it make sense to simulate this tx first?

api/relay/_strategies/local-signers.ts
```diff
@@ -0,0 +1,92 @@
...
+            chainId: request.chainId,
+            to: request.to,
+            data: encodedCalldata,
+            from: wallet.address,
+          };
+          const tx = await wallet.sendTransaction(txRequest);
```
gsteenkamp89, line 42: would it make sense to simulate this tx first?

api/relay/_utils.ts
```diff
@@ -0,0 +1,320 @@
...
+    throw new InvalidParamError({
+      message: "Invalid deposit signature",
+      param: "signatures.deposit",
+    });
+  }
+}
+
+export function encodeCalldataForRelayRequest(request: RelayRequest) {
+  let encodedCalldata: string;
+
```
gsteenkamp89, line 176: should we add a todo for transfer with auth?

api/relay/status.ts
```diff
@@ -0,0 +1,37 @@
...
+  logger.debug({
+    at: "Relay/status",
+    message: "Request params",
+    params: request.query,
+  });
+
+  try {
+    assert(request.query, RelayRequestStatusSchema);
+
+    const cachedRelayRequest = await getCachedRelayRequest(
```
gsteenkamp89, line 29: nit: can we handle if a request hash is not found

api/swap/permit/_utils.ts
```diff
@@ -115,13 +137,22 @@ export async function buildPermitTxPayload(
   ]);
   return {
     eip712: {
-      permit: permitTypedData.eip712,
-      deposit: depositTypedData.eip712,
+      permit: {
+        ...permitTypedData.eip712,
+        message: stringifyBigNumProps(permitTypedData.eip712.message),
```
gsteenkamp89, line 142: nit: since `stringifyBigNumProps` runs recursively, can we call it once at the root of this object, just in case we might miss a value, or if we add fields in later PRs we might forget to also serialize that object correctly

gsteenkamp89 (approved):
Awesome work! I only have a few questions, for my own understanding

api/relay/_queue.ts
dohaki: yea that's the idea. by using a queue and setting parallelism to the number of signers we have, we should be able to mitigate nonce collisions on a single chain.

api/relay/_queue.ts
dohaki: it depends on the strategy. for example, if using Gelato, we don't need to handle nonce collisions because they take care of it. if using local signers, then we need to implement a mechanism/handler that's able to prevent nonce collisions.

api/relay/_strategies/gelato.ts
dohaki: hmm yea block time could make sense actually. a bit hesitant on using an env var though

api/relay/_strategies/local-signers.ts
dohaki: `sendTransaction` actually does an internal `estimateGas` call already if no `gasLimit` is provided

api/relay/_utils.ts
dohaki: yep will do

api/relay/status.ts
dohaki: yes good call!

api/swap/permit/_utils.ts
dohaki: yep good point. I hoisted it up to api/swap/permit now

PR #1349, opened 2025-01-03
Test: Add gas costs to /gas-prices endpoint

nicholaspai:
This PR is not designed to be merged. I only want to print out the current gas fee computation components on the production API to compare against those with https://github.com/across-protocol/frontend/pull/1339

Signed-off-by: nicholaspai <npai.nyc@gmail.com>


api/gas-prices.ts
```diff
@@ -1,31 +1,95 @@
...
+          outputToken: tokenAddress,
+          recipientAddress: DEFAULT_SIMULATED_RECIPIENT_ADDRESS,
+          originChainId: 0, // Shouldn't matter for simulation
+          destinationChainId: Number(chainId),
+        };
+        const relayerFeeCalculatorQueries = getRelayerFeeCalculatorQueries(
+          Number(chainId)
+        );
+        return relayerFeeCalculatorQueries.getGasCosts(
+          buildDepositForSimulation(depositArgs),
```
nicholaspai, line 66: This should parallel how the /iimits endpoint computes the `gasFeePercent` today:

- Passes the `gasPrice` computed using `getMaxFeePerGas` and the `gasUnits` passed from a `getGasCosts()` call with `omitMarkup = true` into [getRelayerFeeDetails](https://github.com/across-protocol/frontend/blob/177ef0f3f3c40ffd10cb3ca80b959724327c756a/api/limits.ts#L192) which [ultimately](https://github.com/across-protocol/sdk/blob/d7940448b315610ff2c0efadadd6042aae82e73a/src/relayFeeCalculator/relayFeeCalculator.ts#L351) [calls](https://github.com/across-protocol/sdk/blob/d7940448b315610ff2c0efadadd6042aae82e73a/src/relayFeeCalculator/relayFeeCalculator.ts#L250) the SDK's `getGasCosts` function

james-a-morris (changes requested):
proactively throwing a "request change" to ensure this doesn't get merged

PR #1350, opened 2025-01-06
improve(api): make buffer cap logic more intuitive

james-a-morris:
A slight clean up to simplify existing logic


gsteenkamp89 (approved):
Totally agree this is more intuitive

dohaki (approved):
LGTM

PR #1353, opened 2025-01-06
feat: unified swap endpoint

dohaki:
Closes ACX-3576

### DIFF
--- a/api/_permit.ts
+++ b/api/_permit.ts
@@ -26,7 +26,7 @@ export async function getPermitTypedData(params: {
   deadline: number;
   eip712DomainVersion?: number;
 }) {
-  const { domainSeparator, eip712DomainVersion, nonce } =
+  const { name, domainSeparator, eip712DomainVersion, nonce } =
     await getPermitArgsFromContract(params);
 
   return {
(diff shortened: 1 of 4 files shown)


api/_transfer-with-auth.ts
```diff
@@ -35,6 +54,49 @@ export async function getReceiveWithAuthTypedData(params: {
...
+
+export async function getReceiveWithAuthArgsFromContract(params: {
+  tokenAddress: string;
+  chainId: number;
+  ownerAddress: string;
+  spenderAddress: string;
+  value: BigNumberish;
+  eip712DomainVersion?: number;
 }) {
   const provider = getProvider(params.chainId);
```
bmzig, line 102: nit: Should you rename this to erc20Auth or erc20ReceiveWithAuth?

api/_permit.ts
```diff
@@ -44,21 +114,16 @@ export async function getPermitTypedData(params: {
...
     versionFromContractResult.status === "fulfilled"
       ? versionFromContractResult.value
       : undefined;
-  const nonce = nonceResult.value;
-  const domainSeparator = domainSeparatorResult.value;
+  const nonce: BigNumberish = nonceResult.value;
+  const domainSeparator: string = domainSeparatorResult.value;
 
   const eip712DomainVersion = [1, 2, "1", "2"].includes(versionFromContract)
     ? Number(versionFromContract)
```
bmzig, line 129: nit (line 132, which might be out of scope): Can we reuse `hashDomainSeparator` from [here](https://github.com/across-protocol/frontend/blob/7f7aba48466f2045729a07d6df7d7daa69d0712a/api/_transfer-with-auth.ts#L25)

bmzig (approved):
Two (potentially out of scope) nits.

api/_permit.ts
dohaki: yep good point. will tackle in a separate cleanup PR

api/_transfer-with-auth.ts
dohaki: yep good point. will tackle in a separate cleanup PR

PR #1354, opened 2025-01-06
script for comparing gas prices

gsteenkamp89:


scripts/gas-prices.ts
```diff
@@ -0,0 +1,204 @@
...
+    res.json()
+  )) as Array<Route>;
+
+  // Extract all unique destination tokens
+  const uniqueDestinationTokens = Array.from(
+    new Set(routes.map((route) => route.destinationTokenSymbol))
+  );
+
+  // Break the tokens into batches of 5
+  const tokenBatches = chunkArray(uniqueDestinationTokens, BATCH_SIZE);
```
nicholaspai, line 70: I think you'll get more consistent results if you make all queries for the same destination chain at the same time. This is because the API caches gas prices for each destination chain for some amount of time, so all routes with the same `destinationChain` will use the same gas price which affects the gas costs

scripts/gas-prices.ts
```diff
@@ -0,0 +1,204 @@
...
+  for (let i = 0; i < array.length; i += size) {
+    result.push(array.slice(i, i + size));
+  }
+  return result;
+}
+
+// use this script to compare gas prices for relayers between 2 environments
+async function compareGasPrices() {
+  const PROD_URL = "https://app.across.to";
+  const TEST_URL = "https://app-frontend-v3-git-gas-markup-fee-uma.vercel.app";
```
nicholaspai, line 58: Can you make this TEST_URL configurable via env?

nicholaspai (dismissed):
Very nice, just two suggestions

scripts/gas-prices.ts
```diff
@@ -0,0 +1,204 @@
...
+
+const BATCH_SIZE = 5;
+
+function chunkArray<T>(array: T[], size: number): T[][] {
+  const result: T[][] = [];
+  for (let i = 0; i < array.length; i += size) {
+    result.push(array.slice(i, i + size));
+  }
+  return result;
+}
```
james-a-morris, line 53: Do we not have this in the SDK or is the intention to not include dependencies?

scripts/gas-prices.ts
james-a-morris: +1 Not a huge fan of hardcoding this preview URL

scripts/gas-prices.ts
```diff
@@ -0,0 +1,204 @@
...
+  recommendedDepositInstant: string;
+  relayerFeeDetails: {
+    relayFeeTotal: string;
+    relayFeePercent: string;
+    gasFeeTotal: string;
+    gasFeePercent: string;
+    capitalFeeTotal: string;
+    capitalFeePercent: string;
+  };
+};
```
james-a-morris, line 18: We have this definition in other locations of the code - could be worth consolidating to prevent duplication/having to change multiple locations when the limits api updates

scripts/gas-prices.ts
gsteenkamp89: good suggestion, i've made changes

scripts/gas-prices.ts
gsteenkamp89: 👍

scripts/gas-prices.ts
gsteenkamp89: you're right! refactored

james-a-morris (dismissed):
one nit

nicholaspai (dismissed):

nicholaspai (dismissed):

james-a-morris (dismissed):

PR #1355, opened 2025-01-07
refactor: cleanup + align swap response formats

dohaki:
This cleans up a few things and aligns swap response formats to be consistent. Also adds a test script for the unified swap endpoint


gsteenkamp89 (approved):
Very neat, nice one 👏

PR #1357, opened 2025-01-07
Improve: Add new config file

evaldofelipe:


james-a-morris (changes requested):
proactively requesting review until this is ready to go in

PR #1363, opened 2025-01-09
support per-chain fillDeadline env var

gsteenkamp89:


dohaki (commented):
LGTM 👍 We should update the env var on Vercel before merging this one in so it directly builds with the new one

dohaki (approved):

gsteenkamp89:
> LGTM 👍 We should update the env var on Vercel before merging this one in so it directly builds with the new one

Done! I'll wait for the build, then we can merge this and promote it to production

dohaki:
I think you could just merge because once in master it needs to rebuild anyways

PR #1365, opened 2025-01-09
feat: Add origin chain fee multipliers

nicholaspai:
Currently we apply a markup to gas fees based on the chain where the fill transaction is expected to be sent. We can improve this even further by adding additional fee multipliers based on where the deposit is originating from. For example, some chains have higher re-org risk so the deposit to fill latency is expected to be longer than for lower re-org-risk destinations.


api/_utils.ts
```diff
@@ -87,16 +87,29 @@ const {
...
 export const priorityFeeMarkup: {
   [chainId: string]: number;
 } = JSON.parse(PRIORITY_FEE_MARKUP || "{}");
+// Markup applied to base fee and priority fee based on the origin chain of a fill. Should be set based on the expected
+// time it would take for a filler to wait until the deposit is far enough behind HEAD to fill it. For example, some
+// chains have higher re-org rates so fillers are expected to wait longer before filling the deposits and therefore,
+// a higher buffer should be applied on the estimated gas price to allow for more volatility.
+export const originChainBaseFeeMarkup: {
+  [chainId: string]: number;
+} = JSON.parse(ORIGIN_BASE_FEE_MARKUP || "{}");
```
nicholaspai, line 109: we might need to make this a origin-destination keyed dictionary. Otherwise this multiplier isn't very useful since the actual multiplier we want to apply is based on the destination chain gas process

PR #1366, opened 2025-01-09
improve(API): Pass gasPrice into relayerFeeDetails call

nicholaspai:
We should pass the pre-computed gas price into this call so that we don't query the gas price twice. Additionally, we can use the scaled up gas price value to derive the gas fee totals, which currently are not using the scaled up values

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1948,6 +1948,7 @@ export function isContractCache(chainId: number, address: string) {
 
 export function getCachedFillGasUsage(
   deposit: Parameters<typeof buildDepositForSimulation>[0],
+  gasPrice: BigNumber,
   overrides?: Partial<{
     spokePoolAddress: string;
     relayerAddress: string;
@@ -1975,6 +1976,7 @@ export function getCachedFillGasUsage(
       buildDepositForSimulation(deposit),
       overrides?.relayerAddress,
       {
+        gasPrice,
         // Scale the op stack L1 gas cost component by the base fee multiplier.
         // Consider adding a new environment variable OP_STACK_L1_GAS_COST_MARKUP if we want finer-grained control.
         opStackL1GasCostMultiplier: getGasMarkup(deposit.destinationChainId)
--- a/api/limits.ts
+++ b/api/limits.ts
@@ -164,7 +164,7 @@ const handler = async (
       message,
     };
 
-    const [tokenPriceNative, _tokenPriceUsd, latestBlock, gasCosts, gasPrice] =
+    const [tokenPriceNative, _tokenPriceUsd, latestBlock, gasPrice] =
       await Promise.all([
         getCachedTokenPrice(
           l1Token.address,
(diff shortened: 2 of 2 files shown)


api/_utils.ts
```diff
@@ -1975,6 +1976,7 @@ export function getCachedFillGasUsage(
       buildDepositForSimulation(deposit),
       overrides?.relayerAddress,
       {
+        gasPrice,
```
nicholaspai, line 1979: Passing this in avoids a redundant SDK.GasPriceOracle.getGasPriceEstimate call

api/_cache.ts
```diff
@@ -73,7 +73,10 @@ export function buildCacheKey(
 }
 
 export function buildInternalCacheKey(...args: (string | number)[]): string {
-  return buildCacheKey("QUOTES_API", ...args);
+  return buildCacheKey(
+    `${process.env.CACHE_PREFIX ? process.env.CACHE_PREFIX + "-" : ""}QUOTES_API`,
```
nicholaspai, line 77: I found this useful for testing and isolating my cache usage away from other preview deployments

api/gas-prices.ts
```diff
@@ -70,50 +71,32 @@ const handler = async (
           const relayerFeeCalculatorQueries = getRelayerFeeCalculatorQueries(
             Number(chainId)
           );
-          const opStackL1GasCostMultiplier = getGasMarkup(
```
nicholaspai, line 73: All the changes in this file align it with the /limits implementation for computing gas costs:
 making sure for Linea we don't pass in gasPrice and let SDK.getGasCosts estimate gas costs using the unsignedTx from the deposit

api/gas-prices.ts
```diff
@@ -70,50 +71,32 @@ const handler = async (
...
+                    : gasPrices[i].maxFeePerGas,
+                opStackL1GasCostMultiplier: opStackL1DataFeeMarkup,
+                baseFeeMultiplier: baseFeeMarkup,
+                priorityFeeMultiplier: priorityFeeMarkup,
               }
             );
-          // OPStack chains factor in the L1 gas cost of including the L2 transaction in an L1 rollup batch
-          // into the total gas cost of the L2 transaction.
-          let opStackL1GasCost: ethers.BigNumber | undefined = undefined;
-          if (sdk.utils.chainIsOPStack(Number(chainId))) {
```
nicholaspai, line 91: All this `opStackL1GasCost` recomputation can be replaced because the SDK as of 3.4.8 returns the `opStackL1GasCost`

api/gas-prices.ts
```diff
@@ -124,12 +107,23 @@ const handler = async (
         Object.keys(chainIdsWithToken).map((chainId, i) => [
           chainId,
           {
-            gasPrice: gasPrices[i].maxFeePerGas.toString(),
+            gasPrice:
```
nicholaspai, line 110: For linea we can infer the gas fee components trivially because the maxFee is hardcoded at 7

api/_utils.ts
nicholaspai: Also, if we don't pass in this gas price, then this function will return gasCosts that we use in the relayerFeeCalculator but it recomputes the gas price without any multipliers. So, the resultant tokenGasCost is likely too low

api/limits.ts
```diff
@@ -164,39 +164,34 @@ const handler = async (
...
       transferRestrictedBalances,
       fullRelayerMainnetBalances,
     ] = await Promise.all([
-      getRelayerFeeDetails(
-        depositArgs,
-        tokenPriceNative,
-        relayer,
-        gasPrice,
-        gasCosts?.nativeGasCost,
-        gasCosts?.tokenGasCost
```
nicholaspai, line 198: Currently, we are using this tokenGasCost that wasn't dependent on the `gasPrice` we computed in the `latestGasPriceCache(destinationChainId).get()` call, which means that this tokenGasCost estimate is low

api/_cache.ts
dohaki: +1

dohaki (approved):
Nice this looks good to me

PR #1367, opened 2025-01-09
improve(API): Derive tokenGasCost locally in /limits

nicholaspai:
Rather than pass the `tokenGasCosts` derived from the SDK.estimateGasCosts call, derive it locally. The current code doesn't use the `gasPrice` fetched from the SDK to compute the `tokenGasCosts` so the token gas costs are inaccurate


nicholaspai:
Going to replace this with https://github.com/across-protocol/frontend/pull/1366

PR #1368, opened 2025-01-10
feat: add Lens Sepolia

gsteenkamp89:


james-a-morris (dismissed):

dohaki (commented):
NIT: There are two linter warnings we should address

dohaki (approved):
LGTM

PR #1369, opened 2025-01-10
improve(API): Make gasFeeTotal more accurate in /limits and increase cache hit frequency

nicholaspai:
This PR contains several optimizations that ultimately make the `gasFeeTotal` as accurate as possible while ensuring that the `/limits` endpoint is retrieving gas fee components from the cache as often as possible, keeping the endpoint latency down.

## `stale-while-revalidate`

We can reduce the /limits endpoint cache time to 1s while the revalidate period is 59s so that after the cached value is >1s old we can immediately start recomputing the limits value. This means in the best case we'll have as fresh gas cost data as possible--essentially, as quickly as we can re-compute the gas fee totals.

## Gas price caching:

We should ideally use less stale gas price data. This PR reduces the gas price caching time to 10s from 5s.

## Gas cost caching:

We currently cache the entire `tokenGasCost` value but this means we are negating some of the optimizations we have made to use as accurate gas fee estimation as possible.  This is because `tokenGasCost` is dependent on the `gasPrice` so if we cache the `tokenGasCost` longer than the gas price gets updated, then we're ultimately returning stale gas price data. 

This PR breaks down the `tokenGasCost` computation into `nativeGasCost` and `l1DataFee` values and caches them individually. This allows us to use a longer TTL for these values which are expected to change less frequently than the `gasPrice` estimation which we can reset the cached value very frequently.

## Cron job 

This PR adds `nativeGasCost` and `l1DataFee` cache resets to the cron job to keep these values fresh in the cache as often as possible. The ultimate effect is that the `/limits` endpoint should always hit the cache for `nativeGasCost`, `gasPrice` and `l1DataFee` and that these cached values are as fresh as possible.

This PR depends on https://github.com/across-protocol/sdk/pull/826

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1979,7 +1979,7 @@ export function getCachedNativeGasCost(
   // We can use a long TTL since we are fetching only the native gas cost which should rarely change.
   // Set this longer than the secondsPerUpdate value in the cron cache gas prices job.
   const ttlPerChain = {
-    default: 60,
+    default: 120,
   };
   const cacheKey = buildInternalCacheKey(
     "nativeGasCost",
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -2079,7 +2079,8 @@ export function getCachedOpStackL1DataFee(
 }
 
 export function latestGasPriceCache(
-  deposit: Parameters<typeof buildDepositForSimulation>[0],
+  chainId: number,
+  deposit?: Parameters<typeof buildDepositForSimulation>[0],
   overrides?: Partial<{
     relayerAddress: string;
   }>
@@ -2088,37 +2089,39 @@ export function latestGasPriceCache(
     default: 5,
   };
   return makeCacheGetterAndSetter(
-    buildInternalCacheKey("latestGasPriceCache", deposit.destinationChainId),
+    buildInternalCacheKey("latestGasPriceCache", chainId),
     ttlPerChain.default,
-    async () => (await getMaxFeePerGas(deposit, overrides)).maxFeePerGas,
+    async () =>
+      (await getMaxFeePerGas(chainId, deposit, overrides)).maxFeePerGas,
     (bnFromCache) => BigNumber.from(bnFromCache)
   );
 }
 
 export async function getMaxFeePerGas(
-  deposit: Parameters<typeof buildDepositForSimulation>[0],
+  chainId: number,
+  deposit?: Parameters<typeof buildDepositForSimulation>[0],
   overrides?: Partial<{
     relayerAddress: string;
   }>
 ): Promise<sdk.gasPriceOracle.GasPriceEstimate> {
-  const { destinationChainId } = deposit;
+  if (deposit && deposit.destinationChainId !== chainId) {
+    throw new Error(
+      "Chain ID must match the destination chain ID of the deposit"
+    );
+  }
   const {
     baseFeeMarkup: baseFeeMultiplier,
     priorityFeeMarkup: priorityFeeMultiplier,
-  } = getGasMarkup(destinationChainId);
-  const unsignedFillTxn = await getUnsignedFillTxnFromDeposit(
-    deposit,
-    overrides?.relayerAddress
-  );
-  return sdk.gasPriceOracle.getGasPriceEstimate(
-    getProvider(destinationChainId),
-    {
-      chainId: destinationChainId,
-      unsignedTx: unsignedFillTxn,
-      baseFeeMultiplier,
-      priorityFeeMultiplier,
-    }
-  );
+  } = getGasMarkup(chainId);
+  const unsignedFillTxn = deposit
+    ? await getUnsignedFillTxnFromDeposit(deposit, overrides?.relayerAddress)
+    : undefined;
+  return sdk.gasPriceOracle.getGasPriceEstimate(getProvider(chainId), {
+    chainId,
+    unsignedTx: unsignedFillTxn,
+    baseFeeMultiplier,
+    priorityFeeMultiplier,
+  });
 }
 
 /**
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -1979,7 +1979,7 @@ export function getCachedNativeGasCost(
   // We can use a long TTL since we are fetching only the native gas cost which should rarely change.
   // Set this longer than the secondsPerUpdate value in the cron cache gas prices job.
   const ttlPerChain = {
-    default: 120,
+    default: 60,
   };
   const cacheKey = buildInternalCacheKey(
     "nativeGasCost",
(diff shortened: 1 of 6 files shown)


api/limits.ts
```diff
@@ -164,34 +165,47 @@ const handler = async (
...
-        destinationChainId === CHAIN_IDs.LINEA
-          ? undefined
-          : latestGasPriceCache(destinationChainId).get(),
-      ]);
+    const [
+      tokenPriceNative,
+      _tokenPriceUsd,
+      latestBlock,
+      gasPrice,
+      nativeGasCost,
```
nicholaspai, line 173: Added `nativeGasCost` to this promise.all since now its no longer dependent on `gasPrice`.

api/limits.ts
```diff
@@ -164,34 +165,47 @@ const handler = async (
...
+    ] = await Promise.all([
+      getCachedTokenPrice(
+        l1Token.address,
+        sdk.utils.getNativeTokenSymbol(destinationChainId).toLowerCase()
+      ),
+      getCachedTokenPrice(l1Token.address, "usd"),
+      getCachedLatestBlock(HUB_POOL_CHAIN_ID),
+      // We only want to derive an unsigned fill txn from the deposit args if the destination chain is Linea
+      // because only Linea's priority fee depends on the destination chain call data.
+      latestGasPriceCache(
```
nicholaspai, line 183: `latestGasPriceCache` gets called now for every single chain, as opposed to skipping for Linea, but for Linea, the `depositArgs` are defined

api/_utils.ts
```diff
@@ -1970,94 +1970,158 @@ export function isContractCache(chainId: number, address: string) {
...
+  chainId: number,
+  deposit?: Parameters<typeof buildDepositForSimulation>[0],
+  overrides?: Partial<{
+    relayerAddress: string;
+  }>
+) {
   const ttlPerChain = {
-    default: 30,
-    [CHAIN_IDs.ARBITRUM]: 15,
+    default: 5,
```
dohaki, line 2074: When we consider the cron job `cron-cache-gas-prices`, this TTL doesn't matter much. The more correct value to tweak would be this one https://github.com/across-protocol/frontend/blob/9e18813baa5797418b98860e3884542a58a10d82/api/cron-cache-gas-prices.ts#L15-L18

dohaki (commented):
This looks great 👍 I left one comment on the TTL. I think we could achieve a higher cache hit rate if we keep the gas price cache TTL slightly longer than the `updateInterval` of the cron job in `cron-cache-gas-prices.ts`.

api/_utils.ts
nicholaspai: Good call, updated this again. I've also updated the cron-cache-gas-prices file to also cache native gas cost and L1 data fees in the cron job. This should be a one-stop cron job we can use to keep all caches pretty warm

nicholaspai: Can you check the TTL's in `_utils` and the cron-cache-gas-price `secondsToUpdate` to make sure the values make sense?

api/_utils.ts
```diff
@@ -1970,94 +1970,169 @@ export function isContractCache(chainId: number, address: string) {
...
   overrides?: Partial<{
     relayerAddress: string;
   }>
 ) {
+  // We can use a long TTL since we are fetching only the native gas cost which should rarely change.
+  // Set this longer than the secondsPerUpdate value in the cron cache gas prices job.
   const ttlPerChain = {
-    default: 10,
-    [CHAIN_IDs.ARBITRUM]: 10,
+    default: 60,
```
nicholaspai, line 1982: We should ideally update this once per cron job if it runs every 60 seconds

api/_utils.ts
```diff
@@ -1970,94 +1970,169 @@ export function isContractCache(chainId: number, address: string) {
...
+  overrides?: Partial<{
+    relayerAddress: string;
+  }>
+) {
+  // The L1 data fee should change after each Ethereum block since its based on the L1 base fee.
+  // However, the L1 base fee should only change by 12.5% at most per block.
+  // We set this higher than the secondsPerUpdate value in the cron cache gas prices job which will update this
+  // more frequently.
+  const ttlPerChain = {
+    default: 60,
```
nicholaspai, line 2031: We should update this ideally every 12s--every ethereum block

api/_utils.ts
```diff
@@ -1970,94 +1970,169 @@ export function isContractCache(chainId: number, address: string) {
...
+  chainId: number,
+  deposit?: Parameters<typeof buildDepositForSimulation>[0],
+  overrides?: Partial<{
+    relayerAddress: string;
+  }>
+) {
+  // We set this higher than the secondsPerUpdate value in the cron cache gas prices job which will update this
+  // more frequently.
   const ttlPerChain = {
     default: 30,
```
nicholaspai, line 2085: This gas price should be updated every 5s

api/cron-cache-gas-prices.ts
```diff
@@ -68,8 +111,75 @@ const handler = async (
...
+      const depositArgs = getDepositArgsForChainId(chainId, outputTokenAddress);
+      const gasCostCache = getCachedNativeGasCost(depositArgs);
+
+      while (true) {
+        const diff = Date.now() - functionStart;
+        // Stop after `maxDurationSec` seconds
+        if (diff >= maxDurationSec * 1000) {
+          break;
+        }
+        const gasCost = await gasCostCache.get();
```
nicholaspai, line 138: So if this loop runs every 10s and we're updating the OpStackL1DataFee every 10s, we also should update the native gas cost once per 60s assuming one of these loops the `gasCostCache.get()` call forces the cache to reset its data

api/limits.ts
```diff
@@ -221,15 +237,21 @@ const handler = async (
         )
       ),
     ]);
-    // This call should not make any additional RPC queries if gasCosts is defined--for any deposit
-    // with an empty message.
+    const tokenGasCost =
```
nicholaspai, line 240: We no longer cache the tokenGasCost as its dependent on the gasPrice, the nativeGasCost and the opStackL1GasCost, each of which should be cached at different intervals and using those values we can trivially derive the tokenGasCost

api/limits.ts
```diff
@@ -221,15 +237,21 @@ const handler = async (
...
+    // and tokenGasCost.
     const relayerFeeDetails = await getRelayerFeeDetails(
       depositArgs,
       tokenPriceNative,
       relayer,
       gasPrice,
-      gasCosts?.nativeGasCost,
-      gasCosts?.tokenGasCost
+      nativeGasCost,
+      tokenGasCost
```
dohaki, line 254: NIT: Should we also return those values as part of the response? I think it could be useful.

dohaki (dismissed):
Looking good 👍

api/limits.ts
nicholaspai: https://github.com/across-protocol/frontend/pull/1369/commits/5441dc6e37999b35256b213d481c1f2ec3f1bde0

dohaki (approved):
Looking gooood

PR #1372, opened 2025-01-14
feat(API): Split gas component cron jobs up into different files

nicholaspai:
Updating the gas costs and gas prices in the same cron job makes the cache resets slower due to the memory constraints of a single file, and when updating the gas price every 5s, this makes a difference.

This PR splits the gas cost updates, which are 10s and 30s for l1 data fees and native gas costs, separate from the gas price updates so that the gas price updates can be closer to 5s. Currently its anywhere from 5s to 10s.

Additionally, this PR adds try-catches around the `await cache.set()` calls so that the a failure due to an upstream RPC call (e.g. I've seen this happen with linea_estimateGas calls a lot) doesn't take down the cron job for the full run.

Finally, there was an issue with the current cron job where all combinations of chainId+outputToken were repeated many times, leading to tons of wasteful computation. This PR uses `Set's` to make sure costs are computed once for each destinationChain+outputToken combination

### DIFF
--- a/api/cron-cache-gas-costs.ts
+++ b/api/cron-cache-gas-costs.ts
@@ -108,10 +108,17 @@ const handler = async (
       }
     };
 
-    const getOutputTokensToChain = (chainId: number) =>
+    const getOutputTokensToChain = (chainId: number) => {
+      const destinationTokens = new Set<string>();
       availableRoutes
         .filter(({ destinationChainId }) => destinationChainId === chainId)
-        .map(({ destinationToken }) => destinationToken);
+        .forEach(({ destinationToken }) => {
+          if (!destinationTokens.has(destinationToken)) {
+            destinationTokens.add(destinationToken);
+          }
+        });
+      return Array.from(destinationTokens);
+    };
 
     const cacheUpdatePromise = Promise.all(
       mainnetChains.map(async (chain) => {
--- a/api/cron-cache-gas-costs.ts
+++ b/api/cron-cache-gas-costs.ts
@@ -159,32 +159,34 @@ const handler = async (
       }
     };
 
+    const getOutputTokensToChain = (chainId: number) =>
+      availableRoutes
+        .filter(({ destinationChainId }) => destinationChainId === chainId)
+        .map(({ destinationToken }) => destinationToken);
+
     // The minimum interval for Vercel Serverless Functions cron jobs is 1 minute.
     // But we want to update gas data more frequently than that.
     // To circumvent this, we run the function in a loop and update gas prices every
     // `secondsPerUpdateForChain` seconds and stop after `maxDurationSec` seconds (1 minute).
-    await Promise.all(
+    const cacheUpdatePromise = Promise.all([
       mainnetChains.map(async (chain) => {
-        const routesToChain = availableRoutes.filter(
-          ({ destinationChainId }) => destinationChainId === chain.chainId
+        await Promise.all(
+          getOutputTokensToChain(chain.chainId).map((outputToken) =>
+            updateNativeGasCostPromise(chain.chainId, outputToken)
+          )
         );
-        const outputTokensForChain = routesToChain.map(
-          ({ destinationToken }) => destinationToken
+      }),
+      mainnetChains.map(async (chain) => {
+        await Promise.all(
+          getOutputTokensToChain(chain.chainId).map((outputToken) =>
+            updateL1DataFeePromise(chain.chainId, outputToken)
+          )
         );
-        await Promise.all([
-          Promise.all(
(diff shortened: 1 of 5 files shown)


api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,209 @@
+import { VercelResponse } from "@vercel/node";
```
nicholaspai, line 1: Everything in this file is copied from cron-cache-gas-prices except where I add comments

api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,209 @@
...
+      while (true) {
+        const diff = Date.now() - functionStart;
+        // Stop after `maxDurationSec` seconds
+        if (diff >= maxDurationSec * 1000) {
+          break;
+        }
+        const gasCost = await gasCostCache.get();
+        if (utils.chainIsOPStack(chainId)) {
+          const cache = getCachedOpStackL1DataFee(depositArgs, gasCost);
+          try {
```
nicholaspai, line 112: This try-catch is added around the cache.set

api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,212 @@
...
+            ),
+          ]);
+        })
+      ),
+    ]);
+
+    // The above promises are supposed to complete after maxDurationSec seconds, but there are so many of them
+    // (one per route) that they last one can run quite a bit longer, so force the function to stop after maxDurationSec
+    // so that the serverless function can exit successfully.
+    await Promise.race([cacheUpdatePromises, utils.delay(maxDurationSec)]);
```
nicholaspai, line 194: Force the many promises to stop after 60s, since there are so many of them this Promise.all wasn't completing in under 90s before, evevn though each promise was designed to finish in 60s

api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,212 @@
...
+// Set lower than TTL in getCachedOpStackL1DataFee
+// Set lower than the L1 block time so we can try to get as up to date L1 data fees based on L1 base fees as possible.
+const updateL1DataFeeIntervalsSecPerChain = {
+  default: 10,
+};
+
+// Set lower than TTL in getCachedNativeGasCost. This should rarely change so we should just make sure
+// we keep this cache warm.
+const updateNativeGasCostIntervalsSecPerChain = {
+  default: 20,
```
nicholaspai, line 37: Lowered this to 20s from 30s since there are so many routes to update, the real latency of updating this per route is practically quite a bit higher (5-10s) than whatever is set here. We want to make sure this cache gets reset once per run so that the cache stays warm. The TTL is 120s so one update per run should be totally fine

api/cron-cache-l1-data-fee.ts
```diff
@@ -0,0 +1,155 @@
...
+    };
+
+    const getOutputTokensToChain = (chainId: number) =>
+      availableRoutes
+        .filter(({ destinationChainId }) => destinationChainId === chainId)
+        .map(({ destinationToken }) => destinationToken);
+
+    const cacheUpdatePromise = Promise.all(
+      mainnetChains
+        .filter((chain) => utils.chainIsOPStack(chain.chainId))
```
nicholaspai, line 126: Added this filter here since we don't need to update l1 date fees for non op stack chains

api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,152 @@
...
+        await utils.delay(secondsPerUpdate);
+      }
+    };
+
+    const getOutputTokensToChain = (chainId: number) => {
+      const destinationTokens = new Set<string>();
+      availableRoutes
+        .filter(({ destinationChainId }) => destinationChainId === chainId)
+        .forEach(({ destinationToken }) => {
+          if (!destinationTokens.has(destinationToken)) {
```
nicholaspai, line 116: Using a set here cuts down computation a lot for duplicate routes

api/cron-cache-gas-costs.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      while (true) {
+        const diff = Date.now() - functionStart;
+        // Stop after `maxDurationSec` seconds
+        if (diff >= maxDurationSec * 1000) {
+          break;
+        }
+        try {
+          await cache.set();
+          updateCounts[chainId][outputTokenAddress]++;
```
nicholaspai, line 98: This updateCount helped debug a lot that I was sending tons of duplicate queries for chain+outputToken combinations

dohaki (approved):
This makes sense to me

PR #1373, opened 2025-01-14
feat: add stablecoins for a2a bridging

dohaki:


api/_dexes/cross-swap-service.ts
NSN-WORK, line 1: the suggested-fees interface has down, it's return the message that the invalid params. 

plz, try a transactions and you will get the error.

NSN-WORK: the interface suggested-fees did not work for 40 minutes, maybe your pull request have trigger the bug of the paramaters checker.

dohaki: hey thanks for flagging! the issue was a malformed object from an upstream provider we use.

PR #1378, opened 2025-01-15
fix: override vercel dev typescript version

dohaki:
There was an annoying issue for a while with the local usage of the Vercel CLI. When testing serverless functions it would always error with TSC errors which were cause due to the usage of Typescript <v5. And these error logs would make application logs unreadable. This PR overrides the resolution for the dependencies now.

Note that you have to use `yarn dev:api` now instead of the global Vercel CLI installation.


gsteenkamp89 (approved):
This is such HUGE help! 👏

PR #1381, opened 2025-01-17
improve(API): Change default fee markup values to be more developer-friendly

nicholaspai:
In case developer forgets to set fee markups explicitly, set more reasonable defaults such that the default suggested fees are conservative enough to avoid relayers from viewing the quotes as unprofitable.


dohaki (approved):
LGTM

james-a-morris (approved):

PR #1382, opened 2025-01-21
ENH: Set default relayer weight to 0

cc7768:
This PR changes the default relayer weight to 0.


dohaki (approved):

james-a-morris (commented):
confirming we've tested this prior to merging

james-a-morris (approved):

PR #1384, opened 2025-01-22
feat(hyperliquid): estimate fees properly

james-a-morris:
This update sets the fees more accurately by calling into the suggested fees API endpoint with a simulated message payload. 

~~The caveat here is that due to rounding issues there's sometimes a 1 unit refund against the recipient (0.000001 USDC). Examples:~~
~~* https://arbiscan.io/tx/0x0ad884bf8eadcda92a2ad3fc55d62d79bac30ac80c83f0d0c32e04b06e8ace56~~
~~* https://arbiscan.io/tx/0xc800914ff0d79bc6c5032abfe3a90cfbd35c48109420b1d0f1b04d4720dab0f5~~


dohaki (approved):
LGTM

PR #1404, opened 2025-01-29
feat: add wagmi + viem

dohaki:
Closes ACX-3682

- Adds wagmi and viem 
- Enhances helper scripts to support initialization of wagmi

Note that this PR is just a prerequisite for the upcoming custom wallet UI work. This doesn't break anything nor does it introduce user-facing functionality or changes.


james-a-morris (commented):
Is the goal to eventually move off of Ethers/onboard?

james-a-morris (dismissed):

dohaki:
> Is the goal to eventually move off of Ethers/onboard?

Eventually, yes. But with our custom wallet UI in mind, we should probably do a staged approach:
1. Move off onboard and use wagmi: Wagmi can still be used with ethers. So in this stage we don't need to migrate to viem yet.
2. Move off ethers and use viem

gsteenkamp89 (approved):
Okay I like that we're not throwing out ethers in this PR just yet. Your comment about a staged approach seems logical to me.

PR #1405, opened 2025-01-30
fix: custom gas token bridging

dohaki:
When bridging from a chain with a custom gas token (e.g. GRASS Lens Sepolia), the dApp was not correctly handling it as a native token.


gsteenkamp89 (approved):
good catch!

PR #1413, opened 2025-01-30
improve: don't show intermediary route as potential option

james-a-morris:
Across currently does not support same-chain bridging. This change filters external projects out when they appear as an origin/destination to the intermediary chain.

I.e. if you have Arbitrum selected in the origin chain you should not see Hyperliquid as a destination. If you have Hyperliquid selected on the destination you shouldn't see Arbitrum as an option in the origin.


dohaki (approved):
LGTM

PR #1415, opened 2025-01-30
feat(op-rewards): disable in UI once ceiling is hit

james-a-morris:


dohaki (approved):
LGTM

gsteenkamp89 (approved):
LGTM! looks like we're getting close to the limit here

PR #1434, opened 2025-02-07
Feat/doctor who

gsteenkamp89:


src/utils/constants.ts
```diff
@@ -539,6 +539,10 @@ export const disabledChainIdsForAvailableRoutes = (
   process.env.REACT_APP_DISABLED_CHAINS_FOR_AVAILABLE_ROUTES || ""
 ).split(",");
 
+export const disabledChainIdsForUI = (
```
james-a-morris, line 542: Maybe in the future we could collapse these three into enumerated types: disabledChains: { [chain: number]: "disabled" | "ui-only" | "available-routes-only"}

james-a-morris (approved):

PR #1435, opened 2025-02-10
feat: add capability to cap limits via env var

mrice32:


james-a-morris (approved):

dohaki (dismissed):

melisaguevara:
@mrice32 I just checked and we do have the env var `CHAIN_USD_MAX_DEPOSITS` to handle this.

For example for Lisk we have it set to:
```
CHAIN_USD_MAX_DEPOSITS = {
 "1135":
 {
  "USDT":"25000",
  "WETH":"35000",
  "ETH": "35000",
  "LSK":"10000"
 }
}
```

Values are expressed in USD and the code converts those to token amounts. We decided at that time this was easier to handle.
Do you think we can continue using that one?

melisaguevara (commented):
@mrice32 I just checked and we do have the env var `CHAIN_USD_MAX_DEPOSITS` to handle this.

For example for Lisk we have it set to:
```
CHAIN_USD_MAX_DEPOSITS = {
 "1135":
 {
  "USDT":"25000",
  "WETH":"35000",
  "ETH": "35000",
  "LSK":"10000"
 }
}
```

The values are expressed in USD, and the code converts them to token amounts. We initially thought this was easie but we could also change it to use token amounts directly if needed.

Do you think we can continue using this setup?

mrice32:
> @mrice32 I just checked and we do have the env var `CHAIN_USD_MAX_DEPOSITS` to handle this.
> 
> For example for Lisk we have it set to:
> 
> ```
> CHAIN_USD_MAX_DEPOSITS = {
>  "1135":
>  {
>   "USDT":"25000",
>   "WETH":"35000",
>   "ETH": "35000",
>   "LSK":"10000"
>  }
> }
> ```
> 
> The values are expressed in USD, and the code converts them to token amounts. We initially thought this was easie but we could also change it to use token amounts directly if needed.
> 
> Do you think we can continue using this setup?

This is great! No, USD amounts is perfect.

PR #1438, opened 2025-02-10
fix: deposit status tracking post upgrade

dohaki:
Correctly parse new depositor address in status tracking.


gsteenkamp89 (approved):

james-a-morris (changes requested):
looks good but do we need to remove out the indexer load test?

james-a-morris (dismissed):
don't want to block we can add this in later

james-a-morris (approved):

PR #1450, opened 2025-02-13
improve: external config files

evaldofelipe:


api/_env.ts
dohaki, line 1: I think we need some sort of equivalent of this for the FE build as well 🤔

scripts/pre-build-env.sh
dohaki, line 1: NIT: Can we format this correctly?

dohaki (commented):
The API part looks good 👍 But we need to add similar logic to the FE build. Should be some config in the vite.config.js we can use. See
- https://vite.dev/guide/env-and-mode.html#env-files
- https://vite.dev/guide/api-javascript.html#loadenv

api/_env.ts
dohaki: Resolved

scripts/pre-build-env.sh
evaldofelipe: good catch! fixed

.gitignore
james-a-morris, line 1: will this impact the .example.env changes if we ever need to make them?

api/_env.ts
```diff
@@ -0,0 +1,14 @@
+import path from "path";
+import fs from "fs";
+import dotenv from "dotenv";
+
+let envPath = path.join(process.cwd(), "output_vercel_api.env");
+let envFile = fs.readFileSync(envPath, "utf-8");
```
james-a-morris, line 6: Is this something we should do synchronously?

api/_env.ts
```diff
@@ -0,0 +1,14 @@
...
+dotenv.populate(
+  process.env as dotenv.DotenvPopulateInput,
+  dotenv.parse(envFile)
+);
+
+export const getEnvs = () => {
+  return process.env;
+};
```
james-a-morris, line 14: OOC is this to ensure that by importing `getEnvs` into the file we get the code above loaded?

package.json
```diff
@@ -55,8 +55,9 @@
     "start": "export REACT_APP_GIT_COMMIT_HASH=$(git rev-parse HEAD) && vite",
     "dev": "export REACT_APP_GIT_COMMIT_HASH=$(git rev-parse HEAD) && vite --port $PORT --host",
     "dev:api": "vercel dev --listen 127.0.0.1:3000",
-    "build": "yarn prebuild && export REACT_APP_GIT_COMMIT_HASH=$(git rev-parse HEAD) && tsc && vite build",
+    "build": "yarn prebuild-env && yarn prebuild && if [ -f ./output_vercel_build.env ]; then . ./output_vercel_build.env; fi && echo $GIT_ENV_EXPORTED && export REACT_APP_GIT_COMMIT_HASH=$(git rev-parse HEAD) && tsc && vite build",
```
james-a-morris, line 58: Should this be inlined in a shell script? Inline conditionals could be complex

api/_env.ts
mrice32: Yep, that's the idea at least.

api/_env.ts
mrice32: if we wanted to do it async, we'd have to make getEnvs async, right?

package.json
mrice32: Agree. @evaldofelipe I would suggest moving this conditional logic into a separate bash script.

mrice32 (commented):
Looks great! Left a few comments on @james-a-morris's comments.

Omitting approval for now, since I want to make sure @james-a-morris and @dohaki take a close look at this.

package.json
evaldofelipe: I can't insert this logic inside a script, its need to be executed on the build shell environment. but to simplify, I've added the logic to another yarn command, staying on the same bash env.

api/_env.ts
james-a-morris: Yes - but I believe we can include async calls in the top level

james-a-morris (commented):
replied to comments. Otherwise LGTM but this should be thoroughly tested in a preview ENV prior to merging to avoid crashing the FE

james-a-morris (approved):
lgtm

dohaki (approved):

PR #1460, opened 2025-02-18
feat: l1 token config fallback

dohaki:
This adds some fallback logic to the API in case of an un-parsable JSON. We have a CRON job which runs every hour to cache parseable L1 token configs. If we fail to parse the string directly fetched from the ConfigStore, we use the cached one.

### DIFF
--- a/api/cron-cache-l1-token-configs.ts
+++ b/api/cron-cache-l1-token-configs.ts
@@ -1,0 +1,73 @@
+import { VercelResponse } from "@vercel/node";
+import { TypedVercelRequest } from "./_types";
+
+import {
+  HUB_POOL_CHAIN_ID,
+  getLogger,
+  handleErrorCondition,
+  getL1TokenConfigCache,
+} from "./_utils";
+import { UnauthorizedError } from "./_errors";
+
+import mainnetChains from "../src/data/chains_1.json";
+
+const handler = async (
+  request: TypedVercelRequest<Record<string, never>>,
+  response: VercelResponse
+) => {
+  const logger = getLogger();
+  logger.debug({
+    at: "CronCacheL1TokenConfigs",
+    message: "Starting cron job...",
+  });
+  try {
+    const authHeader = request.headers?.["authorization"];
+    if (
+      !process.env.CRON_SECRET ||
+      authHeader !== `Bearer ${process.env.CRON_SECRET}`
+    ) {
+      throw new UnauthorizedError();
+    }
+
+    // Skip cron job on testnet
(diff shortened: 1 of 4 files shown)


vercel.json
```diff
@@ -20,6 +20,10 @@
     {
       "path": "/api/cron-ping-endpoints",
       "schedule": "* * * * *"
+    },
+    {
+      "path": "/api/cron-cache-l1-token-configs",
+      "schedule": "0 * * * *"
```
dohaki, line 26: Could also once a day? L1 token configs don't get updated often anyways

nicholaspai: So let's say we update an L1 token config. We want it to update pretty quickly. It'd be ideal if we could update this say once an hour but then be able to force an update easily by requesting a certain path

nicholaspai (approved):
Looks great, left some q's

james-a-morris (commented):
Do we have a way to call into pager duty in the event that this is triggered?

vercel.json
dohaki: Yep, we can always manually trigger this job by sending a request to this endpoint.

Also note that this implementation still tries to query the token config directly from the contract first. Only if the raw JSON string is not parseable, we use the cached one (which is always valid)

api/cron-cache-l1-token-configs.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    if (!mainnetChain) {
+      throw new Error("Mainnet chain not found");
+    }
+
+    const l1TokenAddresses = mainnetChain.inputTokens.map(
+      (token) => token.address
+    );
+
+    for (const l1TokenAddress of l1TokenAddresses) {
```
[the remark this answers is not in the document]
dohaki: Sounds good

dohaki:
> Do we have a way to call into pager duty in the event that this is triggered?

Yea, I was thinking of something like that too. But IMO it makes more sense to add PD integration into Datadog. So basically if the number of error logs where the `JSON.parse` fails hit a certain threshold then trigger alert. Wdyt?

api/cron-cache-l1-token-configs.ts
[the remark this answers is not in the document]
dohaki: https://github.com/across-protocol/frontend/pull/1460/commits/8249ba28645f490ff65cd5eb692dd2647b538f27

gsteenkamp89 (approved):
Looks good nice work

pxrl (approved):

PR #1469, opened 2025-02-24
fix: recompute fees when switching arb <-> hl

dohaki:
The fee query when switching from Arbitrum to HL didn't always update. Even though the fees query got recomputed, the quote query did not.


gsteenkamp89 (approved):
Good catch

PR #1485, opened 2025-03-05
Feat/generate routes for solana

gsteenkamp89:
closes ACX-3680

Makes edits to the routes generation script to handle Solana addresses.


james-a-morris (commented):
Is this needed so early? Or is this to test the scripts?

gsteenkamp89:
> Is this needed so early? Or is this to test the scripts?

We need this for testing. This is to be merged into the epic branch, not master

PR #1488, opened 2025-03-06
fix: weighted random

dohaki:


pxrl (approved):
I alternatively had:

```
@@ -12,6 +13,10 @@ export function weightedRandom(relayers: string[]): string {
     })
     .filter(({ effectiveWeight }) => effectiveWeight > 0.0);
 
+  if (relayerWeights.length === 0) {
+    return sdk.constants.ZERO_ADDRESS;
+  }
+
   // Select the relayer with the highest effective weight.
   const { relayer: exclusiveRelayer } = relayerWeights
     .slice(1)
```

PR #1491, opened 2025-03-07
fix: add refunded and slowFillRequested status

KanishkKhurana:
Adding 2 more status enums that i have observed while using the status api.


dohaki (approved):
LGTM

PR #1496, opened 2025-03-12
feat: approval swap flow

dohaki:
Subset of https://github.com/across-protocol/frontend/pull/1269 that only contains approval-based swap flows.

Closes ACX-3900


james-a-morris (commented):
This diff is very large. We should schedule some time to review this in the order you recommend or provide some guidance on where the priority items are?

dohaki:
> This diff is very large. We should schedule some time to review this in the order you recommend or provide some guidance on where the priority items are?

Yes, fair point. I will set up a call

james-a-morris (approved):

PR #1501, opened 2025-03-18
feat: uniswap sdk-based quote retrieval

dohaki:
Allows to specify an optional param of type `type QuoteSource = "trading-api" | "sdk"` when using the `SwapRouter02` strategy for Uniswap. Specifying the value `"sdk"` will result in the usage of a quote retrieval via the SDK for cases where the Trading API doesn't support certain chains.

Closes ACX-3903

~Based on #1500, therefore draft~


gsteenkamp89 (approved):
Looks great, are these tested?

james-a-morris (approved):

PR #1504, opened 2025-03-20
feat: utilize rl disallow list in FE

james-a-morris:
This PR enables Risk Labs disallow list to be referenced by the FE.


daywiss (approved):
lgtm. out of curiousity did u have any concerns about response times

dohaki (approved):
Nice LGTM

james-a-morris:
> lgtm. out of curiousity did u have any concerns about response times

This was a quick endpoint - I'm not concerned about response times. The calls are done out of the hot path so we are not impacting user experience

james-a-morris:
> lgtm. out of curiousity did u have any concerns about response times

This was a quick endpoint - I'm not concerned about response times. The calls are done out of the hot path so we are not impacting user experience

PR #1509, opened 2025-03-25
add rpc proxy endpoint

gsteenkamp89:
## motivation

For some new chains we might have test RPC endpoints we may use that are protected via headers.
This endpoint acts as a proxy for the frontend to use as normal.

The API may skip this step since we can safely add headers there without a problem.

TODOs:
- The frontend must be smart enough to know when to use proxy for a certain chainId


james-a-morris (commented):
failing vercel build

gsteenkamp89:
> failing vercel build

I mistakenly thought we run the server when our test suites run, so we can test against our endpoints. but we actually do not do this, hence why this test was failing.

dohaki (dismissed):
LGTM

dohaki (approved):

PR #1533, opened 2025-04-07
sort tokens by balance

gsteenkamp89:
closes ACX-3973

Sorts tokens in TokenSelector by balance (desc.)

### DIFF
--- a/src/views/Bridge/components/TokenSelector.tsx
+++ b/src/views/Bridge/components/TokenSelector.tsx
@@ -111,7 +111,9 @@ export function TokenSelector({
         ...token,
         balance: balances?.[i] ? balances[i] : BigNumber.from(0),
       }))
-      .sort((a, b) => (b.balance.lt(a.balance) ? -1 : b.balance.gt(a.balance) ? 1 : 0));
+      .sort((a, b) =>
+        b.balance.lt(a.balance) ? -1 : b.balance.gt(a.balance) ? 1 : 0
+      );
   }, [orderedTokens, balances]);
 
   return (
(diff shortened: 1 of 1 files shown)


src/views/Bridge/components/TokenSelector.tsx
```diff
@@ -105,9 +106,26 @@ export function TokenSelector({
...
+        ...token,
+        balance: balances?.[i] ? balances[i] : BigNumber.from(0),
+      }))
+      .sort((a, b) => {
+        return Number(formatUnits(b.balance, b.decimals)) <
+          Number(formatUnits(a.balance, a.decimals))
+          ? -1
+          : Number(formatUnits(b.balance, b.decimals)) >
+              Number(formatUnits(a.balance, a.decimals))
+            ? 1
```
gsteenkamp89, line 121: comparing BigNumbers would place `0.001 WETH` above `100 USDC`

james-a-morris: OOC why are we sorting different token values based on their raw numbers? I feel like this should be normalized in USDC values or something similar

james-a-morris (commented):
one comment

src/views/Bridge/components/TokenSelector.tsx
gsteenkamp89: That's a great point. 
Will have to double check this with product before moving further

gsteenkamp89:
Closing this. 
We want to respect the ordering defined in `orderedTokenLogos`

PR #1534, opened 2025-04-07
rework remote env fetcher script, refactor prebuild command

gsteenkamp89:
## motivation

Issues solved:
1. current `prebuild` scripts runs twice on build
2. bash script to fetch remote env file fails if dev doesn't have some environment variables set manually (or globally) 
3. env vars live in too many places: `output_vercel.env`, `.env.local` & `.env`
4. script to fetch env fails silently if dev does not have required initial env setup

rename some scripts:
- `prebuild.ts` => `fetch-remote-config.ts`
- `prebuild-env.sh` => `fetch-remote-env.ts` (rewritten in typrescript for uniformity)
- ~~load-env~~

# Update

We do not merge env files, because there could be collisions and it can be hard to reason about which vars would take precedence over others.
Now we:
Fetch remote env, save it to `.env`
When we parse env files, it is done in this order:
1. `.env.local` (local should override remote env)
2. `.env` 

this is the order in which Vite will parse them so we follow the same order in the api by explicitly parsing them in this order with dotenv


james-a-morris (dismissed):
lgtm

james-a-morris (approved):

dohaki (approved):

PR #1535, opened 2025-04-07
fix: improve swap quote error handling

dohaki:


src/hooks/useUniversalSwapQuote.ts
```diff
@@ -81,5 +82,17 @@ export function useUniversalSwapQuote(
...
+      if (axios.isAxiosError(error)) {
+        const responseData = error.response?.data as { code?: string };
+        if (
+          responseData?.code === "AMOUNT_TOO_LOW" ||
+          responseData?.code === "SWAP_QUOTE_UNAVAILABLE"
+        ) {
+          return false;
+        }
+      }
+      return true;
```
gsteenkamp89, line 95: do we want to retry for other 400 + errors eg. `InvalidParamError` ?

gsteenkamp89 (dismissed):
very nice :)

pxrl (commented):
@dohaki Nice work to make the error diagnosis a lot more granular :+1: No other comments from me - I'll defer to @gsteenkamp89 on his open question.

src/hooks/useUniversalSwapQuote.ts
dohaki: Yeah, we can definitely improve error handling further. But for now the two errors here should be fine. `InvalidParamError` _should_ theoretically not be thrown in the context of our dApp. Sometimes we do throw a `SimulationError` where retries could make sense. Handling everything a bit more streamlined, should be a separate PR though.

src/hooks/useUniversalSwapQuote.ts
```diff
@@ -81,18 +81,25 @@ export function useUniversalSwapQuote(
...
+    const responseData = error.response?.data as { code?: string };
+    if (
+      responseData?.code === "AMOUNT_TOO_LOW" ||
+      responseData?.code === "SWAP_QUOTE_UNAVAILABLE"
+    ) {
+      return false;
+    }
+  }
+  return true;
+}
```
gsteenkamp89, line 105: Nice, we can expand on this in subsequent PRs 👍

gsteenkamp89 (approved):
Nice one, just one nit

src/hooks/useUniversalSwapQuote.ts
```diff
@@ -81,18 +81,25 @@ export function useUniversalSwapQuote(
...
-        const responseData = error.response?.data as { code?: string };
-        if (
-          responseData?.code === "AMOUNT_TOO_LOW" ||
-          responseData?.code === "SWAP_QUOTE_UNAVAILABLE"
-        ) {
-          return false;
-        }
+    refetchInterval: (query) => {
+      if (query.state.error && !shouldRetry(query.state.error)) {
+        return false;
```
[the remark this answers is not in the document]
dohaki: Yea I noticed that if `retry: false` but `refetchInterval: 5000`, it would still refetch every 5 seconds.

pxrl (approved):

PR #1536, opened 2025-04-07
improve: use generic speed up txn

james-a-morris:
We are moving to version-generic event names. We should move forward with using the generic speed-up deposit fn.

Note: this is an opinionated change

### DIFF
--- a/src/views/Transactions/hooks/useSpeedUp.tsx
+++ b/src/views/Transactions/hooks/useSpeedUp.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect } from "react";
-import { BigNumber } from "ethers";
+import { BigNumber, ethers } from "ethers";
 import { useMutation } from "@tanstack/react-query";
 
 import { useConnection, useBridgeFees, useIsWrongNetwork } from "hooks";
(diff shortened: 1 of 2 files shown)


nicholaspai (commented):
This won't work unless `depositor` is in bytes32 format, which it doesn't seem like it is

src/views/Transactions/hooks/useSpeedUp.tsx
```diff
@@ -86,9 +86,14 @@ export function useSpeedUp(transfer: Deposit, token: Token) {
       );
 
       const depositor = await signer.getAddress();
+      // Scale the depositor address to 32 bytes
+      const paddedDepositor = ethers.utils.hexZeroPad(
+        ethers.utils.hexlify(depositor),
+        32
+      );
```
gsteenkamp89, line 93: can use [this](https://github.com/across-protocol/sdk/blob/24d2de2a402a1ec7f2379ec4cd376d77e6d0bf46/src/utils/AddressUtils.ts#L46)  utility from the sdk

src/views/Transactions/hooks/useSpeedUp.tsx
```diff
@@ -86,9 +86,14 @@ export function useSpeedUp(transfer: Deposit, token: Token) {
...
       const spokePool = config.getSpokePool(transfer.sourceChainId, signer);
-      const txResponse = await spokePool.speedUpV3Deposit(
-        depositor,
+      const txResponse = await spokePool.speedUpDeposit(
+        paddedDepositor,
         BigNumber.from(transfer.depositId),
         updatedOutputAmount,
         newRecipient,
```
gsteenkamp89, line 99: `newRecipient` will also have to be padded

james-a-morris: https://github.com/across-protocol/frontend/pull/1536/commits/bf8e93d54a9d9cedb16ceeeadfd9e3a2c0fadbd9

gsteenkamp89 (approved):
Looks good, thanks for getting this in!

PR #1538, opened 2025-04-09
Feat/deafult out token stable in

gsteenkamp89:
closes ACX-3976
## motivation

If Lens is destination chain and token IN is a stable coin, we want to set the default out token to `GHO` (user may change this)


dohaki (approved):
LGTM

PR #1547, opened 2025-04-14
use 700px for chain selector

gsteenkamp89:
closes ACX-3977
So we can show 12 chains without scroll


gsteenkamp89:
> love this - do we have any considerations if the user has a screen with a < 400px display?

@james-a-morris There's still some weirdness on mobile browsers. I'll have to play around a bit more.

dohaki (dismissed):
LGTM! Do we want to apply the same for the token selector?

dohaki (approved):
Nice

PR #1557, opened 2025-04-16
feat(utils): batch balance queries via multicall

gsteenkamp89:
## motivation

Each new added token/chain increases the number of balance queries exponentially.
This PR adds a class we can use to batch any balance queries (made within a specified interval) via multicall, reducing the number of requests to public RPCs

### DIFF
--- a/src/utils/balances.ts
+++ b/src/utils/balances.ts
@@ -1,0 +1,229 @@
+import { ERC20__factory } from "@across-protocol/contracts";
+import { providers } from "ethers";
+import { callViaMulticall3 } from "../../api/_utils";
+import { getProvider } from "./providers";
+import { getMulticall3, ZERO_ADDRESS } from "./sdk";
+import { useQuery } from "@tanstack/react-query";
+
+export type MultiCallResult = {
+  blockNumber: providers.BlockTag;
+  balances: Record<string, Record<string, string>>;
+};
+
+/**
+ * Fetches the balances for an array of addresses on a particular chain, for a particular erc20 token
+ * @param chainId The blockchain Id to query against
+ * @param addresses An array of valid Web3 wallet addresses
+ * @param tokenAddress The valid ERC20 token address on the given `chainId` or ZERO_ADDRESS for native balances
+ * @param blockTag Block to query from, defaults to latest block
+ * @returns a Promise that resolves to an array of BigNumbers
+ */
+export const getBatchBalanceViaMulticall3 = async (
+  chainId: string | number,
+  addresses: string[],
+  tokenAddresses: string[],
+  blockTag: providers.BlockTag = "latest"
+): Promise<MultiCallResult> => {
+  const chainIdAsInt = Number(chainId);
+  const provider = getProvider(chainIdAsInt);
+
+  const multicall3 = getMulticall3(chainIdAsInt, provider);
+
+  if (!multicall3) {
+    throw new Error("No Multicall3 deployed on this chain");
+  }
+
+  let calls: Parameters<typeof callViaMulticall3>[1] = [];
+
+  for (const tokenAddress of tokenAddresses) {
+    if (tokenAddress === ZERO_ADDRESS) {
+      // For native currency
+      calls.push(
+        ...addresses.map((address) => ({
+          contract: multicall3,
+          functionName: "getEthBalance",
+          args: [address],
+        }))
+      );
+    } else {
+      // For ERC20 tokens
+      const erc20Contract = ERC20__factory.connect(tokenAddress, provider);
+      calls.push(
+        ...addresses.map((address) => ({
+          contract: erc20Contract,
+          functionName: "balanceOf",
+          args: [address],
+        }))
+      );
+    }
+  }
+
+  const inputs = calls.map(({ contract, functionName, args }) => ({
+    target: contract.address,
+    callData: contract.interface.encodeFunctionData(functionName, args),
+  }));
+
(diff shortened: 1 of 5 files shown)


src/utils/balances.ts
```diff
@@ -0,0 +1,229 @@
...
+      balances[address][tokenAddress] = decodedResults[resultIndex].toString();
+      resultIndex++;
+    });
+  }
+
+  return {
+    blockNumber: blockNumber.toNumber(),
+    balances,
+  };
+};
```
gsteenkamp89, line 94: copied from here => https://github.com/across-protocol/frontend/blob/9c2986734641b2f4a3222aa860a04bfbcc31a64d/api/_utils.ts#L1393

src/utils/balances.ts
```diff
@@ -0,0 +1,229 @@
...
+export const balanceBatcher = new BalanceBatcher(getBatchBalanceViaMulticall3);
+
+/**
+ * @param chainId The blockchain chain ID.
+ * @param tokenAddress The ERC20 token address.
+ * @param walletAddress The wallet address whose balance will be fetched.
+ * @param blockTag Optional blockTag (defaults to "latest").
+ * @returns An object containing the balance (a string), loading state, and any error.
+ */
+export function useBalance({
```
gsteenkamp89, line 197: might need to add ability to resolve `tokenSymbol => address`.

src/utils/balances.ts
```diff
@@ -0,0 +1,229 @@
...
+  const multicall3 = getMulticall3(chainIdAsInt, provider);
+
+  if (!multicall3) {
+    throw new Error("No Multicall3 deployed on this chain");
+  }
+
+  let calls: Parameters<typeof callViaMulticall3>[1] = [];
+
+  for (const tokenAddress of tokenAddresses) {
+    if (tokenAddress === ZERO_ADDRESS) {
```
gsteenkamp89, line 39: We'll need to add logic somewhere to handle ETH since we use WETH address internally not the zero address

src/utils/balances.ts
```diff
@@ -0,0 +1,229 @@
...
+interface BatchRequest {
+  chainId: number;
+  blockTag: providers.BlockTag;
+  aggregatedTokens: Set<string>;
+  aggregatedAddresses: Set<string>;
+  requests: BalanceRequest[];
+  timer?: NodeJS.Timeout;
+}
+
+const DEFAULT_BATCH_INTERVAL = 100;
```
gsteenkamp89, line 114: this interval adds an implicit delay. is this too aggressive? is 10ms better?

src/utils/balances.ts
```diff
@@ -0,0 +1,230 @@
...
+          args: [address],
+        }))
+      );
+    }
+  }
+
+  const inputs = calls.map(({ contract, functionName, args }) => ({
+    target: contract.address,
+    callData: contract.interface.encodeFunctionData(functionName, args),
+  }));
```
pxrl, line 65: @gsteenkamp89 I think this could be reordered to squash it down. It should be possible to instantiate a single generic ERC20 contract (not connected to any specific address or provider) and then generate all the balanceOf() calldata. Then that same calldata can be reused across all of the ERC20s we want to query.

So I think it could be something like:

```
const erc20 = ERC20__factory.connect(ZERO_ADDRESS, ...);
const inputs: { target: string, callData: string }[] = [];

for (const address of addresses) {  
  const callData = erc20.interface.encodeFunctionData("balanceOf", address);
  for (const tokenAddress of tokenAddresses) {
    inputs.push({ target: tokenAddress, callData });
  }
}
```

If it's done that way then that should save instantiating a bunch of ephemeral ERC20 contract instances.

src/utils/balances.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Fetches the balances for an array of addresses on a particular chain, for a particular erc20 token
+ * @param chainId The blockchain Id to query against
+ * @param addresses An array of valid Web3 wallet addresses
+ * @param tokenAddress The valid ERC20 token address on the given `chainId` or ZERO_ADDRESS for native balances
+ * @param blockTag Block to query from, defaults to latest block
+ * @returns a Promise that resolves to an array of BigNumbers
+ */
+export const getBatchBalanceViaMulticall3 = async (
+  chainId: string | number,
+  addresses: string[],
```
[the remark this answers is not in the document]
gsteenkamp89: yeah that's exactly right, we're repurposing the logic originally written for relayer balances.

src/utils/balances.ts
```diff
@@ -0,0 +1,230 @@
...
+  tokenAddresses: string[],
```
dohaki, line 25: It might make sense to use symbols instead of addresses here?

src/utils/balances.ts
dohaki: Wdyt of using the symbols instead of addresses in the context of our FE? Based on the chain id we can infer which symbols are native

src/utils/balances.ts
```diff
@@ -0,0 +1,230 @@
...
+  return useQuery({
+    queryKey: [
+      "balance",
+      chainId,
+      token,
+      address,
+      blockTag ?? "latest",
+    ] as const,
+    queryFn: () =>
+      balanceBatcher.queueBalanceCall(
```
dohaki, line 222: This looks interesting. I am just wondering if we could reach the same goal by using optimized tanstack/query options 🤔

PR #1558, opened 2025-04-16
feat: support svm useBalance hooks

dohaki:
Aims to close ACX-3987


src/constants/chains/index.ts
```diff
@@ -50,6 +50,7 @@ export const chainInfoList: ChainInfoList = orderedEnabledChainIds.map(
 
 export const chainInfoTable: ChainInfoTable = Object.fromEntries(
   chainInfoList.map((chain) => {
+    console.log(chain.chainId, chain);
```
gsteenkamp89, line 53: want to keep this?

gsteenkamp89 (approved):
very nice.

src/constants/chains/index.ts
dohaki: removed!

PR #1562, opened 2025-04-24
feat(api): field `outputAmount` + param `allowUnmatchedDecimals`

dohaki:
There might be edge cases where the decimals for the same token do not match on different chains (USDC/USDT on BNB). To handle these cases we add a correctly scaled `outputAmount` response field. We also let consumers opt-in for these routes via the query param `allowUnmatchedDecimals`.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -235,6 +235,7 @@ export const validateChainAndTokenParams = (
     outputToken: string;
     originChainId: string;
     destinationChainId: string;
+    allowUnmatchedDecimals: string;
   }>
 ) => {
   let {
@@ -243,6 +244,7 @@ export const validateChainAndTokenParams = (
     outputToken: outputTokenAddress,
     originChainId,
     destinationChainId: _destinationChainId,
+    allowUnmatchedDecimals,
   } = queryParams;
 
   if (!_destinationChainId) {
@@ -295,12 +297,25 @@ export const validateChainAndTokenParams = (
     });
   }
 
+  if (
+    allowUnmatchedDecimals !== "true" &&
+    inputToken.decimals !== outputToken.decimals
+  ) {
+    throw new InvalidParamError({
+      message:
+        `Decimals of input and output tokens do not match. ` +
+        `Set allowUnmatchedDecimals=true to allow this.`,
+      param: "allowUnmatchedDecimals",
+    });
+  }
+
   return {
     l1Token,
     inputToken,
     outputToken,
     destinationChainId,
     resolvedOriginChainId,
+    allowUnmatchedDecimals: allowUnmatchedDecimals === "true",
   };
 };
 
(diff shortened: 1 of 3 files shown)


api/_utils.ts
```diff
@@ -295,12 +297,25 @@ export const validateChainAndTokenParams = (
     });
   }
 
+  if (
+    allowUnmatchedDecimals !== "true" &&
+    inputToken.decimals !== outputToken.decimals
+  ) {
+    throw new InvalidParamError({
+      message:
```
nicholaspai, line 305: should we return the expected input and output decimals?

dohaki: oh yea good point. I will add `inputToken` and `outputToken` fields like in the swap endpoint

api/_utils.ts
```diff
@@ -295,12 +297,25 @@ export const validateChainAndTokenParams = (
...
+    });
+  }
+
   return {
     l1Token,
     inputToken,
     outputToken,
     destinationChainId,
     resolvedOriginChainId,
+    allowUnmatchedDecimals: allowUnmatchedDecimals === "true",
```
nicholaspai, line 318: Does this mean that if `allowUnmatchedDecimals` is unspecified its defaulted to `true`? Doesn't that mean that if someone doesn't set this, they might get back an unexpected output amount? I think the default should be false right?

api/_utils.ts
```diff
@@ -295,12 +297,25 @@ export const validateChainAndTokenParams = (
...
   }
 
+  if (
+    allowUnmatchedDecimals !== "true" &&
+    inputToken.decimals !== outputToken.decimals
+  ) {
+    throw new InvalidParamError({
+      message:
+        `Decimals of input and output tokens do not match. ` +
+        `Set allowUnmatchedDecimals=true to allow this.`,
```
pxrl, line 307: Do you think there's a risk that integrators might just set this and move on without realising the implications? I wonder if we should be pointing them to some guidance in the response, or something.

api/_utils.ts
pxrl, line 318: Minor style nit - what about defining this somewhere above line 300 so it can be used in the if statement at line 300?

api/_utils.ts
```diff
@@ -2061,13 +2078,13 @@ export async function fetchStakingPool(
 export const ConvertDecimals = (fromDecimals: number, toDecimals: number) => {
   // amount: string, BN, number - integer amount in fromDecimals smallest unit that want to convert toDecimals
   // returns: string with toDecimals in smallest unit
```
pxrl, line 2080: Is this comment still correct? Looks like we're returning BNs now.

api/_utils.ts
```diff
@@ -2061,13 +2078,13 @@ export async function fetchStakingPool(
...
-  return (amount: BigNumber): string => {
+  return (amount: BigNumber) => {
```
gsteenkamp89, line 2081:
```suggestion
  return (amount: BigNumber): BigNumber => {
```

Also specify in comment above

api/suggested-fees.ts
```diff
@@ -364,6 +372,19 @@ const handler = async (
...
+        symbol: inputToken.symbol,
+        decimals: inputToken.decimals,
+        chainId: computedOriginChainId,
+      },
+      outputToken: {
+        address: outputToken.address,
+        symbol: outputToken.symbol,
+        decimals: outputToken.decimals,
+        chainId: destinationChainId,
+      },
```
gsteenkamp89, line 387: nice i like how explicit this is

gsteenkamp89 (approved):
excellent!

api/_utils.ts
dohaki: Yea, that's a good point. Probably we should point to the migration guide

nicholaspai (dismissed):
Looks good from my end

api/_utils.ts
dohaki: https://github.com/across-protocol/frontend/pull/1562/commits/cb7d82f77c427214eb17c2fa84a8989b3cdbe21b

api/_utils.ts
dohaki: https://github.com/across-protocol/frontend/pull/1562/commits/cb7d82f77c427214eb17c2fa84a8989b3cdbe21b

api/_utils.ts
dohaki: https://github.com/across-protocol/frontend/pull/1562/commits/cb7d82f77c427214eb17c2fa84a8989b3cdbe21b

pxrl (approved):

PR #1564, opened 2025-04-29
feat: support destination chain relayer fee calculator config overrides

gsteenkamp89:
Part of https://github.com/across-protocol/sdk/pull/997
Adds support for env var `RELAYER_FEE_CAPITAL_COST_DESTINATION_CHAIN_OVERRIDES`
renames ~~`RELAYER_FEE_CAPITAL_COST_OVERRIDES`~~ => `RELAYER_FEE_CAPITAL_COST_ROUTE_OVERRIDES`

### DIFF
--- a/api/_constants.ts
+++ b/api/_constants.ts
@@ -161,6 +161,18 @@ export const coinGeckoAssetPlatformLookup: Record<string, number> = {
 
 export const graphAPIKey = GRAPH_API_KEY;
 
+// {
+//   "TOKEN_SYMBOL": {
+//     "ORIGIN_CHAIN_ID": {
+//       "DESTINATION_CHAIN_ID": {
+//         "lowerBound": "1000000000000000",
+//         "upperBound": "5000000000000000",
+//         "cutoff": "10000000000000000000000",
+//         "decimals": 18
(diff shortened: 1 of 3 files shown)


api/_constants.ts
```diff
@@ -157,11 +161,18 @@ export const coinGeckoAssetPlatformLookup: Record<string, number> = {
 
 export const graphAPIKey = GRAPH_API_KEY;
 
-const relayerFeeCapitalCostOverrides: Record<
+const relayerFeeCapitalCostRouteOverrides: Record<
```
dohaki, line 164: NIT: Shall we add some comments with examples here? So that we know the expected shape

dohaki (commented):
Nice work! One nit

api/_constants.ts
gsteenkamp89: yeah good Idea!, I'll add some

gsteenkamp89: https://github.com/across-protocol/frontend/pull/1564/commits/f970ff7f456cc982895719238f8199cc26a9a92c

dohaki (approved):

PR #1571, opened 2025-05-05
fix: /limits for ultra chain

dohaki:


api/_utils.ts
```diff
@@ -2728,7 +2728,7 @@ export function parseL1TokenConfigSafe(jsonString: string) {
     if (jsonString === "") {
       return {
         rateModel: {
-          UBar: ethers.utils.parseUnits("1").toString(),
+          UBar: "1",
```
nicholaspai, line 2731: Why change this?

api/limits.ts
```diff
@@ -273,6 +275,14 @@ const handler = async (
     const routeInvolvesLiteChain =
       originChainIsLiteChain || destinationChainIsLiteChain;
 
+    const originChainIsUltraLightChain = ULTRA_LIGHT_CHAINS.includes(
+      computedOriginChainId
+    );
+    const destinationChainIsUltraLightChain =
+      ULTRA_LIGHT_CHAINS.includes(destinationChainId);
+    const routeInvolvesUltraLightChain =
```
nicholaspai, line 283: I'm not sure this is exactly parallel to routeInvolvesLiteChain. 

ToLiteChain
- can slow fill

FromLiteChain
- cannot slow fill

ToUltraChain
- cannot slow fill

FromUltraChain
- cannot slow fill

api/limits.ts
```diff
@@ -317,7 +327,7 @@ const handler = async (
       ...transferRestrictedBalances
     ); // balances on destination chain + mainnet
 
-    if (!routeInvolvesLiteChain) {
+    if (!routeInvolvesLiteChain && !routeInvolvesUltraLightChain) {
```
nicholaspai, line 330: So we don't count Lp funds for deposits to lite chains for filling slow fills?

We should, as I think we can send slow fills to a lite chain (but not from)

api/_utils.ts
dohaki: In the SDK, we do substract somewhere 
```
ethers.utils.parseUnits("1").sub(UBar)
```
which led to `division-by-zero`

dohaki: https://github.com/across-protocol/sdk-v2/blob/0a641517321fa152f76e1c411696d7fb80647161/src/lpFeeCalculator/lpFeeCalculator.ts#L83

nicholaspai: Oh you're right, you should change this to 10000000000000000 (0.01) to match on-chain configs for these empty rate models

nicholaspai (approved):

api/_utils.ts
dohaki: ok

PR #1576, opened 2025-05-08
feat: required changes to support BSC

dohaki:
Subset of https://github.com/across-protocol/frontend/pull/1565

Does not enable BSC. Only trying to prevent future merge conflicts with Solana epic branch.


api/_constants.ts
```diff
@@ -259,6 +310,9 @@ export const ENABLED_POOLS_UNDERLYING_TOKENS = [
   TOKEN_SYMBOLS_MAP.SNX,
   TOKEN_SYMBOLS_MAP.POOL,
   TOKEN_SYMBOLS_MAP.WGHO,
+  TOKEN_SYMBOLS_MAP.LSK,
```
pxrl, line 313: @dohaki Any idea why LSK is suddenly showing up here?

api/_constants.ts
```diff
@@ -290,4 +344,5 @@ export const DEFAULT_FILL_DEADLINE_BUFFER_SECONDS = 3.25 * 60 * 60; // 3.25 hour
 export const CUSTOM_GAS_TOKENS = {
   ...sdkConstants.CUSTOM_GAS_TOKENS,
   [CHAIN_IDs.LENS]: "GHO",
+  [CHAIN_IDs.BSC]: "BNB",
```
pxrl, line 347: You could optionally source `nativeToken` from the constants repo. Not important either way; just wanted to flag it.

pxrl (dismissed):
I tested this a bit in the FE and it looks to work fine. Great initiative to proactively reduce the diff.

gsteenkamp89 (dismissed):
good to get this in now 👍

api/_constants.ts
dohaki: We noticed that we missed that. We don't use that endpoint though. So just for completeness. But I actually need to remove BNB and CAKE from this constant. Because they don't have LPs

pxrl (approved):

api/_constants.ts
dohaki: Yes good point. I would need to change a bit more logic in other parts though. Will do that in a follow up

gsteenkamp89 (approved):

PR #1579, opened 2025-05-09
feat: evm/svm recipient edge cases

dohaki:
Closes ACX-3980


src/views/Bridge/components/RecipientRow.tsx
```diff
@@ -1,39 +1,68 @@
...
-  const { data } = useEnsQuery(props.recipient.address);
+  const ecosystem = getEcosystem(props.toChainId);
+
+  const ensCols =
+    ecosystem === "evm" && props.recipient?.address ? (
+      <EnsData evmAddress={props.recipient?.address} />
+    ) : null;
+
+  const recipientAddress = props.recipient?.address;
+  const isRecipientSet = !!recipientAddress;
```
gsteenkamp89, line 26: Not relevant to this PR, but I believe there's a similar service on Solana we could use. SNS I think

src/views/Bridge/hooks/useToAccount.ts
```diff
@@ -1,59 +1,103 @@
...
 
 export type ToAccount = {
   address: string;
   isContract: boolean;
 };
 
 export function useToAccount(toChainId?: number) {
-  const [customToAddress, setCustomToAddress] = useState<string | undefined>();
-  const [toAccount, setToAccount] = useState<ToAccount | undefined>();
+  const [customToAddressEVM, setCustomToAddressEVM] = useState<
```
gsteenkamp89, line 13: What is the default recipient address if I only connect a wallet to the origin chain? Are we forced to connect or can we enter a custom address in this input?

gsteenkamp89 (approved):
Excellent!
I have 1 question but looks good

src/views/Bridge/hooks/useToAccount.ts
dohaki: > Are we forced to connect or can we enter a custom address in this input?

Yep, both options are possible.

src/views/Bridge/components/RecipientRow.tsx
dohaki: Ah good find. Maybe we need to integrate that as well

PR #1584, opened 2025-05-14
feat: Enable Base USDT

pxrl:


pxrl:
> Quotes for routes to Base USDT seems to be failing with insufficient balance
>  
> We probably need to seed the simulated recipient address?

Yeah - it had no balance at the time, but it does now :+1:

PR #1593, opened 2025-05-21
feat: upgrade spoke pool verifier

dohaki:
Closes ACX-4101

The `SpokePoolVerifier` was redeployed with a new `deposit` method signature. This will break when upgrading to the latest contracts package. This PR therefore upgrades the FE to use the new method


scripts/generate-routes.ts
```diff
@@ -130,6 +130,9 @@ const enabledRoutes = {
         CHAIN_IDs.ZORA,
         CHAIN_IDs.WORLD_CHAIN,
         CHAIN_IDs.INK,
+        CHAIN_IDs.ALEPH_ZERO,
+        CHAIN_IDs.SONEIUM,
+        CHAIN_IDs.UNICHAIN,
```
gsteenkamp89, line 135: Why were these not enabled before?

gsteenkamp89 (approved):
Looks good! 
One question

scripts/generate-routes.ts
dohaki: I think we just missed them for some reason

PR #1605, opened 2025-05-28
Fixes filldeadline & empty message for deposits to Solana

gsteenkamp89:
Cherry picked some fixes some this [PR](https://github.com/across-protocol/frontend/pull/1586)

### DIFF
--- a/api/_providers.ts
+++ b/api/_providers.ts
@@ -18,7 +18,7 @@ export const providerCache: Record<string, providers.StaticJsonRpcProvider> =
 
 function getSvmRpcUrl(chainId: number): string {
   const urls = getRpcUrlsFromConfigJson(chainId);
-  const publicNetwork = PUBLIC_NETWORKS[Number(chainId)];
+  const publicNetwork = PUBLIC_NETWORKS[chainId];
   if (urls.length === 0 && !publicNetwork) {
     throw new Error(`No provider URL set for SVM chain: ${chainId}`);
   }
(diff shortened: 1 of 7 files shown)


src/utils/bignumber.ts
```diff
@@ -24,3 +25,20 @@ export function isBigNumberish(value: unknown): value is BigNumberish {
...
+  return BigNumber.from(uin8ArrayToHex(array));
+}
+
+export function uin8ArrayToHex(array: Uint8Array): Hex {
+  return toHex(
+    Array.from(array)
+      .map((b) => b.toString(16).padStart(2, "0"))
+      .join("")
+  );
+}
```
gsteenkamp89, line 44: unrelated but somehow made its way into one of the cherry-picked commits

api/_providers.ts
```diff
@@ -14,21 +16,33 @@ const { RPC_HEADERS } = getEnvs();
 export const providerCache: Record<string, providers.StaticJsonRpcProvider> =
   {};
 
+function getSvmRpcUrl(chainId: number): string {
+  const urls = getRpcUrlsFromConfigJson(chainId);
+  const publicNetwork = PUBLIC_NETWORKS[Number(chainId)];
```
pxrl, line 21: `chainId` is already scoped as a number - should be OK to drop this?

```suggestion
  const publicNetwork = PUBLIC_NETWORKS[chainId];
```

pxrl (approved):

api/_providers.ts
gsteenkamp89: Yep do it

PR #1613, opened 2025-06-02
fix: default separate cache namespace for preview/prod

dohaki:
This PR creates a separate cache namespace for all `preview` deployments by default, which can still be overridden by setting the environment variable `CACHE_PREFIX` for a specific branch or deployment.

Alternatively, the default for previews could be the deployment URL. Not 100% which makes more sense

### DIFF
--- a/api/_cache.ts
+++ b/api/_cache.ts
@@ -10,6 +10,7 @@ const {
   UPSTASH_REDIS_REST_TOKEN,
   UPSTASH_REDIS_READ_ONLY_TOKEN,
   CACHE_PREFIX,
+  VERCEL_ENV,
 } = getEnvs();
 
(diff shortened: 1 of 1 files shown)


mrice32:
> > Alternatively, the default for previews could be the deployment URL. Not 100% which makes more sense
> 
> It might make sense for each deployment to have a unique cache prefix. Otherwise it's very difficult to a/b test to alternative preview deployments. In hindsight I think we've probably been impacted by this previously without realising it.

Totally agree.

mrice32 (approved):

pxrl:
OK for this one to go in? We have a lot of testing based on previews at the moment so it'd be good to ensure there's no cross-pollination happening.

PR #1617, opened 2025-06-03
ci: add job for API tests

ashwinrava:
This PR makes 3 small changes to our `lint-test` workflow.

1. Adds a new job: `api-tests` which runs `yarn test-api`. I added this because the current configuration only runs frontend unit tests and skips the API tests.
2. Combines `yarn format` and `yarn lint` into a single run step.
3. Removed unnecessary second `yarn format`. Since linting will not change any file contents, I don't believe we need a second format.

To verify that adding `yarn test-api` to our workflow won't break anything, I ran it locally and the current tests passed. The next step is to add more coverage.


dohaki:
This looks good

melisaguevara (approved):
looks good!

dohaki (approved):

PR #1618, opened 2025-06-04
fix: handle 7702 correctly

dohaki:
- Suppress the red warning banner when the depositor is a 7702 delegation on the origin chain.
- When the recipient on the destination chain has a 7702 delegation in place, ensure that the outputToken is shown as WETH and not ETH.

Fixes ACX-4114

### DIFF
--- a/src/hooks/useIsContractAddress.ts
+++ b/src/hooks/useIsContractAddress.ts
@@ -1,13 +1,28 @@
 import { useQuery } from "@tanstack/react-query";
+import { isAddress } from "viem";
 import { getCode, noContractCode } from "utils";
 
-export function useIsContractAddress(address?: string, chainId = 1) {
+export function useIsContractAddress(
+  address?: string,
+  chainId = 1,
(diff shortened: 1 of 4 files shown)


src/hooks/useIsContractAddress.ts
```diff
@@ -20,3 +35,12 @@ export function useIsContractAddress(address?: string, chainId = 1) {
 
   return result.data ?? false;
 }
+
+export function is7702Delegate(code: string): boolean {
+  // Sample 7702 delegation bytecode: 0xef010063c0c19a282a1b52b07dd5a65b58948a07dae32b
+  return (
+    code.length === 48 &&
+    code.startsWith("0xef0100") &&
+    isAddress(`0x${code.slice(8)}`)
```
gsteenkamp89, line 44: Wouldn't this always fail? Does isAddress not check byte length?

dohaki: Yep what James said

PR #1619, opened 2025-06-04
refactor: create adapter and handler framework

ashwinrava:
### Summary
This PR aims to modularize and refactor our API handler code by making it framework agnostic and extendable by nature.

### Design
Small diagram I put together in Excalidraw:

Untitled-2025-06-04-1834

### Details
The idea is to create a handler base class which individual handlers can extend so that only core logic needs to be implemented for each new API. We also want to create an adapter class so that each of the handler implementations can be adapted to work with any framework.

- Created `api-handler.ts` which is the base handler abstract class. It allows for easy extendability since individual handlers only need to implement core processing logic.
- Created `api-adapter.ts` which is the base adapter class. It allows for switching between and using new frameworks like Vercel, Hono, etc. Handlers can specify which adapter implementation to use.
- Created `vercel-adapter.ts` which is the specific adapter to conform to Vercel's requirements. This logic is basically what we use today.
- Modified `account-balance.ts` to extend the base handler and leverage the Vercel adapter.

### Testing
- All automated tests pass.

- Deployed APIs locally and tested the one I modified. I am getting back a valid response.

I ran the following: `curl "http://localhost:3000/api/account-balance?token=0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828&account=0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828&chainId=1"`

I got a successful response:

```
[
  {
    at: 'AccountBalance',
    message: 'AccountBalance request received',
    request: [Object: null prototype] {
      token: '0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828',
      account: '0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828',
      chainId: '1'
    }
  }
]
[
  {
    at: 'AccountBalance',
    message: 'AccountBalance request processed',
    response: {
      balance: '2537504860844314319389',
      account: '0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828',
      token: '0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828'
    }
  }
]
```

### DIFF
--- a/api/account-balance.ts
+++ b/api/account-balance.ts
@@ -1,12 +1,7 @@
-import { VercelResponse } from "@vercel/node";
-import { assert, Infer, type, string } from "superstruct";
-import { TypedVercelRequest } from "./_types";
-import {
-  latestBalanceCache,
-  getLogger,
-  handleErrorCondition,
-  validAddress,
-} from "./_utils";
+import { type, string, Infer } from "superstruct";
+import { latestBalanceCache, validAddress } from "./_utils";
+import { ApiHandler } from "./_base/api-handler";
+import { VercelAdapter } from "./_adapters/vercel-adapter";
 
 const AccountBalanceQueryParamsSchema = type({
   token: validAddress(),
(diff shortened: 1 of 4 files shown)


api/account-balance.ts
```diff
@@ -16,44 +11,51 @@ const AccountBalanceQueryParamsSchema = type({
...
-  try {
-    assert(query, AccountBalanceQueryParamsSchema);
-
-    let { token, account, chainId } = query;
+const AccountBalanceResponseSchema = type({
+  balance: string(),
+  account: validAddress(),
+  token: validAddress(),
+});
+
```
melisaguevara, line 19: +1

melisaguevara (approved):
LGTM!

dohaki (approved):
LGTM. Nice work

PR #1623, opened 2025-06-06
tests: rework handler tests

ashwinrava:
### Summary
This PR aims to rework the way we test our handlers. Each API handler does the following:

1. Processes inbound request and returns a response.
2. Calls `handleErrorCondition` for exception handling.

We should have 2 separate sets of unit tests. One which covers all the branches `handleErrorCondition` and then one which covers the core functionality of each handler in success/fail scenarios.

### Details

- `test-helper.ts` - This contains common functions that every handler test can use to setup mocks. I purposely didn't use any Vercel specific logic here so that the helper can be used with any adapter.
- `account-balance.test.ts` - This contains tests to cover all scenarios which can happen within the core handler logic. We can follow this pattern and add tests for each endpoint in future PRs.
- `error-handler.test.ts` - This aims to test every branch of `handleErrorCondition` in `_errors.ts`. Its execution is standalone and doesn't rely on any specific handler invoking it. This way we can ensure that every execution branch is covered.

### Testing
New tests are passing. The next step is to clean up and refactor old tests.

```
Test Suites: 6 passed, 6 total
Tests:       34 passed, 34 total
```

### DIFF
--- a/test/api/handlers/account-balance.test.ts
+++ b/test/api/handlers/account-balance.test.ts
@@ -1,0 +1,95 @@
+import { VercelResponse } from "@vercel/node";
+import { latestBalanceCache } from "../../../api/_utils";
+import handler from "../../../api/account-balance";
+import { TypedVercelRequest } from "../../../api/_types";
+import { createMockRequest, setupTestMocks } from "../helpers/test-helper";
+import { InputError } from "../../../api/_errors";
+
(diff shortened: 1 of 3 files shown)


ashwinrava:
I'll work on top of the Solana branch to avoid merge conflicts later.

PR #1637, opened 2025-06-24
fix: use address instead of depositor when querying deposits

melisaguevara:
This PR updates the `/deposits` query to use `address` instead of `depositor`, this way the indexer returns deposits where the address is either the depositor or the recipient, rather than just the recipient.

Note: this is the way it used to work with the scraper.


gsteenkamp89 (approved):
Nice catch!

PR #1641, opened 2025-06-26
feat: allow granular min deposit config

dohaki:


api/limits.ts
```diff
@@ -574,4 +566,49 @@ const parseAndConvertUsdToTokenUnits = (
...
+ * @param inputTokenSymbol The input token symbol
+ * @returns The minimum deposit in USD
+ */
+function getMinDepositUsd(
+  originChainId: number,
+  destinationChainId: number,
+  inputTokenSymbol: string
+) {
+  const { MIN_DEPOSIT_USD, MIN_DEPOSIT_USD_OVERRIDES } = getEnvs();
+  const minDepositUsdOverrides = JSON.parse(
```
ashwinrava, line 583: Since we're using a nested JSON in an environment variable, it could be useful to wrap this in a try-catch (or maybe in a new util function like `safeParse` which does this) in case the type check fails at runtime. Or maybe it's not needed if we don't expect the environment variable to change much - just a thought.

api/limits.ts
```diff
@@ -574,4 +566,49 @@ const parseAndConvertUsdToTokenUnits = (
...
+    MIN_DEPOSIT_USD_OVERRIDES || "{}"
+  ) as {
+    routes?: { [key: string]: string };
```
ashwinrava, line 586: Think this type could be reusable for other overrides? I like how it goes from most specific to least specific. (Token + Chain) -> (Token) -> (Chain).

ashwinrava (approved):

api/limits.ts
```diff
@@ -574,4 +566,49 @@ const parseAndConvertUsdToTokenUnits = (
...
+    originChains?: { [key: string]: string };
+    destinationChains?: { [key: string]: string };
+    tokens?: { [key: string]: string };
```
ashwinrava, line 589: Think this type could be reusable for other overrides? I like how it goes from most specific to least specific. (Token + Chain) -> (Token) -> (Chain).

PR #1665, opened 2025-07-17
feat(swap): Set body schema and validations for embedded actions

melisaguevara:
This PR introduces the schema for the body of /swap* requests. It defines a structured format for describing destination actions, including the contract to call, the function signature, and a list of positional arguments to populate the call with. Each argument is represented as a structured object, allowing callers to indicate whether the value should be populated dynamically at execution time using the balance of a specified token.

The basic structure looks as follows:
```json
{
  "actions": [
    {
      "target": "0x...", // Contract address to call
      "functionSignature": "function transfer(address to, uint256 amount)", // Full function signature
      "args": [ // Positional inputs for the function
        {
          "value": "0x...", // The argument value
          "populateDynamically": false, // Whether to populate this value dynamically at execution
          "balanceSource": "0x..." // Optional: token to use balance of as dynamic input (required if populateDynamically is true)
        }
        // ... more arguments
      ],
      "value": "0" // ETH value to send with the transaction
    }
    // ... more actions
  ]
}
```

By using positional arguments, this schema also supports complex input types such as tuples, nested tuples, etc.

At runtime, arguments are validated by attempting to ABI-encode them using the provided functionSignature. If encoding fails due to a type mismatch or structural inconsistency, an `AbiEncodingError` is thrown.

This ensures accurate mapping between arguments and Solidity types, strong typing for dynamic and nested structures, and early failure for invalid inputs prior to execution.

### DIFF
--- a/api/swap/_utils.ts
+++ b/api/swap/_utils.ts
@@ -193,16 +193,20 @@ export function handleSwapBody(body: SwapBody) {
   assert(body, SwapBody);
 
   // Helper function to recursively extract only the .value fields from args array
-  const flattenArgs = (args: any[]): any[] =>
-    args.map((arg) => {
+  const flattenArgs = (args: any[], depth: number = 0): any[] => {
+    if (depth > 10) {
+      throw new Error("Arguments array is too deeply nested");
+    }
+    return args.map((arg) => {
       if (Array.isArray(arg)) {
-        return flattenArgs(arg);
+        return flattenArgs(arg, depth + 1);
       } else if (arg && typeof arg === "object" && "value" in arg) {
         return arg.value;
       } else {
         return arg;
       }
     });
+  };
 
   body.actions.forEach((action) => {
     const methodAbi = action.functionSignature;
--- a/api/swap/_utils.ts
+++ b/api/swap/_utils.ts
@@ -171,7 +171,6 @@ const RecursiveArgumentArray: any = lazy(() =>
 // Instructions for a single function call
 const Action = type({
   target: validEvmAddress(),
-  functionName: string(),
   functionSignature: string(), // Will be validated at runtime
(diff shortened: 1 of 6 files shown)


api/swap/_utils.ts
```diff
@@ -130,6 +145,85 @@ export async function handleBaseSwapQueryParams(
...
+// Recursive array type that can have nested arrays at any depth
+// Needed to support arguments of type tuple and array
+const RecursiveArgumentArray: any = lazy(() =>
+  union([ActionArg, array(RecursiveArgumentArray)])
+);
+
+// Instructions for a single function call
+const Action = type({
+  target: validEvmAddress(),
+  functionSignature: string(), // Will be validated at runtime
```
melisaguevara, line 174: I decided to go with the string based approach for functionSignature (e.g., "function transfer(...)") instead of using the full object representation. For complex types like nested tuples, defining a schema for the ABI object becomes way more complicated. Using the string format keeps the request payload simpler and easier to work with.

For the API users it should be easy to get the string representation:
- ethers has a helper function to get it directly from the ABI.
- viem has this [toFunctionSignature](https://viem.sh/docs/utilities/toFunctionSignature) utility which also works just by adding the "function" keyword in front of it.
- There are libraries with utility functions to get it. See for example [ABIType](https://abitype.dev/api/human#formatabiitem-1)

dohaki (commented):
Looking 🔥 Only left a few smaller things

melisaguevara:
> Looks great, but I wonder if we can validate this at the router level? https://docs.nestjs.com/techniques/validation

@james-a-morris What do you mean by "the router level"? My understanding is that we have only handlers, at least for now.

amateima (approved):

PR #1673, opened 2025-07-22
feat: resubmission of PR #1661

ashwinrava:
I had to revert #1661 because of a false alarm. When executing cross swap quotes concurrently using `Promise.all`, the inner exceptions of individual promises are returned as an array of `AggregateError` objects. Also, we were swallowing these exceptions and returning a generic `SwapQuoteUnavailableError` without any cause. Seeing these generic errors without the underlying exception made me think there was a problem with my changes while using the Uniswap strategy. But looking further locally, I was able to successfully reproduce the error and it was actually expected.

The error was that the user was trying to swap too large of an amount from Mainnet USDC to Lens WGHO: `inputToken=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48&outputToken=0x88F08E304EC4f90D644Cec3Fb69b8aD414acf884&originChainId=1&destinationChainId=232&skipAmountLimit=true&recipient=0xc5939F59b3c9662377DdA53A08D5085b2d52b719&amount=12829237738206715000000000000000000`

In this PR, I resubmitted the original contents but added a function to unwrap the errors from individual promises and return the errors as a formatted string that will get logged in the `cause` field.

For testing, I retried the exact call and now we return some more details in the error and also internally log the inner exceptions in the `cause` field. I also tried some valid scenarios and it looks good.

### What the user sees:
```
{
      type: 'AcrossApiError',
      code: 'SWAP_QUOTE_UNAVAILABLE',
      status: 400,
      message: 'Failed to fetch swap quote: No quotes available for exactInput USDC to WGHO from 1 to 232 for amount 12829237738206715000000000000000000'
}
```

### What we see in the logs:
```
{
    at: 'swap/approval',
    code: 'SWAP_QUOTE_UNAVAILABLE',
    message: 'Status 400 - Failed to fetch swap quote: No quotes available for exactInput USDC to WGHO from 1 to 232 for amount 12829237738206715000000000000000000',
    cause: "Quote 1: AxiosError: status: 404, details: No quotes available, Quote 2: AxiosError: status: 400, details: Relayer Address (0x07aE8551Be970cB1cCa11Dd7a11F47Ae82e70E67) doesn't have enough funds to support this deposit; for help, please reach out to https://discord.across.to"
}
```

### DIFF
--- a/api/_dexes/cross-swap-service.ts
+++ b/api/_dexes/cross-swap-service.ts
@@ -9,6 +9,7 @@ import {
   Profiler,
   addMarkupToAmount,
   getBridgeQuoteForExactInput,
+  addTimeoutToPromise,
 } from "../_utils";
 import { CrossSwap, CrossSwapQuotes, QuoteFetchOpts } from "./types";
 import {
@@ -40,7 +41,9 @@ import {
   CrossSwapQuotesRetrievalB2AResult,
 } from "./types";
 
-const indicativeQuoteBuffer = 0.005; // 0.5% buffer for indicative quotes
+const INDICATIVE_QUOTE_BUFFER = 0.005; // 0.5% buffer for indicative quotes
+
+const PROMISE_TIMEOUT_MS = 10000;
 
 export async function getCrossSwapQuotes(
   crossSwap: CrossSwap,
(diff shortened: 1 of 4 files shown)


ashwinrava:
> Nice improvements! Would it be possible to keep the text `"for help, please reach out to https://discord.across.to"` in the error the caller gets?

Discussed offline with @melisaguevara. We'll need a way to consolidate messages from individual promises since there will always be multiple responses. Also, we'd need to decide which ones are relevant to the caller and which ones should be kept internal as debug logs. We're thinking for now we will keep it as is by returning a generic response and using the `cause` field internally for debugging.

amateima (approved):

PR #1676, opened 2025-07-22
EPIC: swap api embedded actions

melisaguevara:


api/swap/approval/_service.ts
```diff
@@ -58,9 +58,9 @@ export async function handleApprovalSwap(
     includeSources,
   } = await handleBaseSwapQueryParams(request.query);
 
-  if (request.body) {
-    handleSwapBody(request.body);
-  }
+  const { actions } = request.body
+    ? handleSwapBody(request.body, Number(request.query.destinationChainId))
+    : { actions: [] };
```
ashwinrava, line 63:
```suggestion
  const { actions } =
    Object.keys(request.body).length > 0
      ? handleSwapBody(request.body, Number(request.query.destinationChainId))
      : { actions: [] };
```

I noticed while testing that this was throwing even when a body was not provided. I think it's because `request.body` will always be present, but the object will be empty in case of GET requests.

dohaki (commented):
Looking great! After https://github.com/across-protocol/frontend/pull/1692 gets merged, I am ready to approve

PR #1679, opened 2025-07-25
fix: misc cross swap service improvements

dohaki:
This PR contains various tweaks and fixes for issues that came up during testing. I will leave some comments in the diff so that you can better understand the reasoning behind the changes. I know that it might be better to split this PR up into multiple smaller ones. So if it is too hard to understand the changes let me know

### DIFF
--- a/api/_dexes/0x/allowance-holder.ts
+++ b/api/_dexes/0x/allowance-holder.ts
@@ -55,107 +55,117 @@ export function get0xStrategy(
     tradeType: TradeType,
     opts?: QuoteFetchOpts
   ) => {
-    let swapAmount = swap.amount;
-    if (tradeType === TradeType.EXACT_OUTPUT) {
-      swapAmount = await estimateInputForExactOutput(
-        swap,
-        `${API_BASE_URL}/price`,
-        API_HEADERS
-      );
-    }
-
-    const sources = opts?.sources;
-    const sourcesParams =
-      sources?.sourcesType === "exclude"
-        ? {
-            excludeSources: sources.sourcesKeys.join(","),
-          }
-        : // We need to invert the include sources to be compatible with the API
-          // because 0x doesn't support the `includeSources` parameter
-          sources?.sourcesType === "include"
-          ? {
-              excludeSources: SOURCES.sources[swap.chainId]
-                .map((s) => s.key)
-                .filter(
-                  (sourceKey) => !sources.sourcesKeys.includes(sourceKey)
-                ),
-            }
-          : {};
-
-    // https://0x.org/docs/api#tag/Swap/operation/swap::allowanceHolder::getQuote
-    const response = await axios.get(
-      `${API_BASE_URL}/${opts?.useIndicativeQuote ? "price" : "quote"}`,
-      {
-        headers: API_HEADERS,
-        params: {
-          chainId: swap.chainId,
-          sellToken: swap.tokenIn.address,
-          buyToken: swap.tokenOut.address,
-          sellAmount: swapAmount,
-          taker: swap.recipient,
-          slippageBps: Math.floor(swap.slippageTolerance * 100),
-          ...sourcesParams,
-        },
+    try {
+      let swapAmount = swap.amount;
+      if (tradeType === TradeType.EXACT_OUTPUT) {
+        swapAmount = await estimateInputForExactOutput(
+          swap,
+          `${API_BASE_URL}/price`,
+          API_HEADERS
+        );
       }
-    );
 
-    const quote = response.data;
+      const sources = opts?.sources;
+      const sourcesParams =
+        sources?.sourcesType === "exclude"
+          ? {
+              excludeSources: sources.sourcesKeys.join(","),
+            }
+          : // We need to invert the include sources to be compatible with the API
+            // because 0x doesn't support the `includeSources` parameter
+            sources?.sourcesType === "include"
+            ? {
+                excludeSources: SOURCES.sources[swap.chainId]
+                  .map((s) => s.key)
+                  .filter(
+                    (sourceKey) => !sources.sourcesKeys.includes(sourceKey)
+                  ),
+              }
+            : {};
 
-    const usedSources = quote.route.fills.map((fill: { source: string }) =>
-      fill.source.toLowerCase()
-    );
-
-    const expectedAmountIn = BigNumber.from(quote.sellAmount);
-    const maximumAmountIn = expectedAmountIn;
-
-    const expectedAmountOut = BigNumber.from(quote.buyAmount);
-    const minAmountOut = BigNumber.from(quote.minBuyAmount);
-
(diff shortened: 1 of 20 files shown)


api/_dexes/0x/allowance-holder.ts
```diff
@@ -55,107 +56,117 @@ export function get0xStrategy(
...
   ) => {
-    let swapAmount = swap.amount;
-    if (tradeType === TradeType.EXACT_OUTPUT) {
-      swapAmount = await estimateInputForExactOutput(
-        swap,
-        `${API_BASE_URL}/price`,
-        API_HEADERS
-      );
-    }
+    try {
```
dohaki, line 59: Wrapping in a try/catch so we can log the exact error from the 0x API. Previously, they would just get buried in one of our `Promise.any`

api/_dexes/0x/allowance-holder.ts
```diff
@@ -55,107 +56,117 @@ export function get0xStrategy(
...
+        expectedAmountOut: swapQuote.expectedAmountOut.toString(),
+        expectedAmountIn: swapQuote.expectedAmountIn.toString(),
+      });
+
+      return swapQuote;
+    } catch (error) {
+      getLogger().debug({
+        at: "0x/fetchFn",
+        message: "Error fetching 0x quote",
+        error: compactAxiosError(error as Error),
```
dohaki, line 162: Added a helper function `compactAxiosError` which logs a compact format of an `AxiosError`. The original error would contain lot of verbose data

api/_dexes/cross-swap-service.ts
```diff
@@ -318,10 +315,6 @@ export async function getCrossSwapQuotesForOutputB2A(
   crossSwap: CrossSwap,
   strategies: QuoteFetchStrategies
 ): Promise<CrossSwapQuotes> {
-  const profiler = new Profiler({
-    at: "api/_dexes/cross-swap-service#getCrossSwapQuotesForOutputB2A",
-    logger: console,
-  });
```
dohaki, line 324: Removed the now obsolete profiler because we now have traces

api/_dexes/cross-swap-service.ts
```diff
@@ -483,28 +482,25 @@ function _prepCrossSwapQuotesRetrievalB2A(
...
-        destinationRouter,
-        depositEntryPoint,
-        bridgeableOutputToken,
-        destinationSwapChainId,
-        destinationStrategy,
-        originStrategy,
-      };
-    })
-  );
+  return destinationStrategies.map((destinationStrategy) => {
```
dohaki, line 485: We don't need to return a combination of each `origin` + `destination` strategy here because we only care about the performance of the `destination` strategy in the case of B2A, i.e. bridge + destination swap.

api/_dexes/cross-swap-service.ts
```diff
@@ -639,55 +623,30 @@ export async function getCrossSwapQuotesForOutputA2B(
...
-    const indicativeOriginSwapQuote = await profiler.measureAsync(
-      result.originStrategy.fetchFn(
-        {
-          ...result.originSwap,
-          depositor: crossSwap.depositor,
-          amount: bridgeQuote.inputAmount.toString(),
-        },
-        TradeType.EXACT_OUTPUT,
-        {
-          useIndicativeQuote: true,
```
dohaki, line 651: Similar to one of above-mentioned comments - we need to use real quote fetching performance when selecting a strategy

api/_dexes/cross-swap-service.ts
```diff
@@ -775,7 +734,7 @@ export async function getCrossSwapQuotesA2A(
     crossSwap.inputToken.chainId,
     crossSwap.outputToken.chainId
   );
-  const bridgeRoutesToCompareChunkSize = 2;
+  const bridgeRoutesToCompareChunkSize = 1;
```
dohaki, line 737: Not 100% sure about this one but trying to reduce the number of concurrent requests when trying to determine the best bridging token route for any-any swaps.

api/_dexes/cross-swap-service.ts
```diff
@@ -895,15 +886,31 @@ export async function getCrossSwapQuotesForExactInputByRouteA2A(
       TradeType.EXACT_INPUT,
       { sources: originSources }
     );
+
+    // 2. Get INDICATIVE destination swap quote for bridgeable output token -> any token
+    //    with exact input amount set to `originSwapQuote.minAmountOut`.
+    const indicativeDestinationSwapQuote =
```
dohaki, line 892: Moving this inside makes sure use both origin + destination fetch performances for determining strategies to use

api/_dexes/cross-swap-service.ts
```diff
@@ -1267,53 +1167,71 @@ function _prepCrossSwapQuotesRetrievalA2A(
...
-        destinationRouter,
-      };
-    })
-  );
+  const baseStrategies =
+    strategiesToUseForComparison === "origin"
+      ? originStrategies
+      : destinationStrategies;
+
+  return baseStrategies.map((strategy) => {
```
dohaki, line 1175: This is an attempt to reduce the number of concurrent requests again. Instead of using every possible origin + destination strategy combination, we use the assumption that if a strategy from `strategiesToUseForComparison` is selected just use the matching destination strategy if possible.

api/_utils.ts
```diff
@@ -177,21 +178,22 @@ export const getLogger = (): LoggingUtility => {
...
       info: (...args) => {
         if (LogLevels[logLevel] <= LogLevels.INFO) {
-          console.info(args);
+          console.info(inspect(args, { depth }));
         }
       },
       warn: (...args) => {
         if (LogLevels[logLevel] <= LogLevels.WARN) {
-          console.warn(args);
+          console.warn(inspect(args, { depth }));
```
dohaki, line 196: That is something I just found out: Allows for better logs, i.e. printing nested fields in objects. Previously it would print `[Object object]` for fields deeper the level 2.

api/swap/approval/index.ts
```diff
@@ -52,6 +52,7 @@ const handler = async (
       );
     } finally {
       span.end();
+      processor.forceFlush();
```
dohaki, line 55: This is an attempt to transport traces to reliably. Sometimes traces would get dropped. I assume something related to the serverless architecture? 🤔

scripts/tests/swap.ts
```diff
@@ -18,7 +19,7 @@ async function swap() {
 
     console.log("\nSwap quote:", JSON.stringify(swapQuote, null, 2));
 
-    if (process.env.DEV_WALLET_PK) {
+    if (process.env.DEV_WALLET_PK && !argsFromCli.skipTxExecution) {
```
dohaki, line 22: Handy if you just want to test retrieving quotes

amateima (approved):

PR #1680, opened 2025-07-26
chore: improve span attributes

dohaki:
Improve span status and request id


api/limits.ts
```diff
@@ -520,7 +519,6 @@ const handler = async (
         message: "Response data",
         responseJson,
       });
-      span.setStatus({ code: SpanStatusCode.OK });
```
dohaki, line 523: Based on the spec [here](https://opentelemetry.io/docs/concepts/signals/traces/#span-status) we don't really need to explicitly set the status to OK

amateima (approved):

PR #1682, opened 2025-07-28
chore: set parent incoming request attr

dohaki:
set more helpful span attributes

### DIFF
--- a/api/_request_utils.ts
+++ b/api/_request_utils.ts
@@ -1,6 +1,8 @@
-import { Span } from "@opentelemetry/api";
-import { ATTR_HTTP_REQUEST_METHOD } from "@opentelemetry/semantic-conventions";
+import { Span, diag } from "@opentelemetry/api";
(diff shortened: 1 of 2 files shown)


api/_request_utils.ts
```diff
@@ -25,5 +27,22 @@ export function setRequestSpanAttributes(
 ) {
   const requestIdToUse = requestId ?? getRequestId(request);
   span.setAttribute("http.request_id", requestIdToUse);
-  span.setAttribute(ATTR_HTTP_REQUEST_METHOD, request.method ?? "unknown");
+
+  if (request.headers) {
+    const attributes = getIncomingRequestAttributes(
```
ashwinrava, line 32: Just curious, what kind of incoming request data does this capture? Couldn't find any documentation on it online.

dohaki: Yea I had to dig a bit as well to find this

PR #1693, opened 2025-07-31
feat: reduce RPC usage in API

mrice32:
This makes three changes to reduce RPC usage in the API:
1. Enable caching at the provider level.
2. Compute fill deadline without calling destination chain RPC.
3. Remove amount dependency in RPC call when checking HubPool utilization, which will allow this call to be identical (and therefore cachable) on different calls to suggested-fees.


api/_utils.ts
```diff
@@ -1300,7 +1302,8 @@ function getProviderFromConfigJson(
     3, // max. concurrency used in `SpeedProvider`
     5, // max. concurrency used in `RateLimitedProvider`
     "RPC_PROVIDER", // cache namespace
-    1 // disable RPC calls logging
+    0, // disable RPC calls logging
+    redisCache
```
dohaki, line 1306: Could you verify that something gets written to Upstash?

api/_dexes/utils.ts
```diff
@@ -355,7 +353,10 @@ export async function extractDepositDataStruct(
     exclusiveRelayer:
       crossSwapQuotes.bridgeQuote.suggestedFees.exclusiveRelayer,
     quoteTimestamp: crossSwapQuotes.bridgeQuote.suggestedFees.timestamp,
-    fillDeadline: await getFillDeadline(spokePool),
+    fillDeadline: getFillDeadline(
```
dohaki, line 356: Nice 💪

api/_utils.ts
mrice32: Good point, doesn't look like it is from the preview. Will take a pass tonight to figure out why.

mrice32: Okay, the issue was fixed. We needed a distance from HEAD.

Fixing this revealed some issues related to differences from our regular redis client vs the upstash one. For some reason, I'm seeing it strip quote characters (which breaks the provider caching). I was able to put in a workaround, but PTAL and let me know if you have a better idea.

melisaguevara (dismissed):
lgtm!

pxrl (dismissed):

dohaki (dismissed):
LGTM

dohaki (dismissed):

dohaki (approved):

PR #1706, opened 2025-08-04
chore: add app fee options to swap quote script

ashwinrava:
It will be useful to have app fees be a part of the new swap quote script for easier testing:

- Add new options and append them to params
- Make `yargs` use strict mode so typos/mistakes in arguments and options will be caught

### DIFF
--- a/scripts/tests/swap-quotes.ts
+++ b/scripts/tests/swap-quotes.ts
@@ -34,6 +34,8 @@ type SwapQuoteResult = {
   outputTokenSymbol: string;
   amount: string;
   tradeType: string;
+  appFeeRecipient?: string;
+  appFeePercent?: string;
   success: boolean;
   data?: any;
   error?: any;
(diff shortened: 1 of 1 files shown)


scripts/tests/swap-quotes.ts
```diff
@@ -34,6 +34,8 @@ type SwapQuoteResult = {
   outputTokenSymbol: string;
   amount: string;
   tradeType: string;
+  appFeeRecipient?: string;
+  appFeePercent?: string;
```
dohaki, line 38: I think it might make more sense to add these to the `swap.ts` script 🤔 Because this script's goal is to just mass retrieve many token/chain combinations. Wdyt?

ashwinrava: Yeah was thinking of adding it there too but this one makes it easy to just quickly test the response JSON for multiple combinations/scenarios.

For example, I ran it with this to see that app fees are being applied to the quote data:
`yarn tsx swap-quotes.ts --oc "10" --dc "42161" --t "USDC, ETH" --afr "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D`

But if we want to only test app fees with full execution, I can remove it from here and add it to the other script.

ashwinrava: Cool, I see you added it in #1711.

PR #1718, opened 2025-08-06
feat: add swap api calldata marker

amateima:
This PR appends `0x73c0de` marker (0x73 = sw) to the calldata in order to identify transactions built by the Swap API.


dohaki (approved):
LGTM

PR #1719, opened 2025-08-06
fix(swap): mismatched decimals

dohaki:
This fixes a few things:
- Convert decimals for indicative quotes (this fixes USDC-BNB related swaps)
- Compact and bubble up more errors
- Allow providing message as POST body in suggested-fees and limits (sometimes Vercel would throw 414 URL_TOO_LONG errors while testing/fixing above)

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -94,6 +94,7 @@ type LoggingUtility = sdk.relayFeeCalculator.Logger;
 type RpcProviderName = keyof typeof rpcProvidersJson.providers.urls;
 
 import { getEnvs } from "./_env";
+import { isMessageTooLong } from "./_message";
 
 const {
   REACT_APP_HUBPOOL_CHAINID,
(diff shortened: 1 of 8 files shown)


api/limits.ts
```diff
@@ -85,6 +90,7 @@ const handler = async (
       const provider = getProvider(HUB_POOL_CHAIN_ID);
 
       assert(query, LimitsQueryParamsSchema);
+      assert(body, LimitsBodySchema);
```
melisaguevara, line 93: I think this breaks GET calls. See for example [this one](https://app-frontend-v3-git-fix-swap-mismatched-decimals-uma.vercel.app/api/suggested-fees?inputToken=0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2&outputToken=0x82aF49447D8a07e3bd95BD0d56f35241523fBab1&destinationChainId=42161&originChainId=1&recipient=0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D&amount=1000000000000000000&skipAmountLimit=true&allowUnmatchedDecimals=true). 

We should check if body is present or maybe the method used before asserting.

api/suggested-fees.ts
```diff
@@ -96,15 +102,19 @@ const handler = async (
       const hubPool = getHubPool(provider);
 
       assert(query, SuggestedFeesQueryParamsSchema);
+      assert(body, SuggestedFeesBodySchema);
```
melisaguevara, line 105: same here

melisaguevara (commented):
Nice!
There's only one issue with the body handling.

api/_utils.ts
```diff
@@ -1006,20 +1007,28 @@ export const getCachedLimits = async (
...
-        destinationChainId,
-        amount,
-        message,
-        recipient,
-        relayer,
-        allowUnmatchedDecimals,
-      },
+      params: messageTooLong ? paramsWithoutMessage : params,
+      method: messageTooLong ? "POST" : "GET",
+      data: messageTooLong ? { message: _message } : undefined,
```
gsteenkamp89, line 1031: would it not be safer to _always_ use a POST then? or do we lose the ability to cache the response then?

gsteenkamp89 (approved):
Looks good!
Just one question

api/_utils.ts
dohaki: > or do we lose the ability to cache the response then?
Yea, that's the case unfortunately...

PR #1725, opened 2025-08-07
fix: align percentage query params `slippage` and `appFee`

dohaki:
Attempt to align percentage query param values `slippage` and `appFee`:
- `slippage` and `appFee` are expressed in range `0 <= x <= 1`, i.e. 0.01 = 1%
- `slippageTolerance` kept for backwards compatibility

### DIFF
--- a/api/swap/_utils.ts
+++ b/api/swap/_utils.ts
@@ -62,11 +62,13 @@ export const BaseSwapQueryParamsSchema = type({
   integratorId: optional(string()),
   refundAddress: optional(validAddress()),
   refundOnOrigin: optional(boolStr()),
+  // DEPRECATED: Use `slippage` instead
   slippageTolerance: optional(positiveFloatStr(50)), // max. 50% slippage
+  slippage: optional(positiveFloatStr(0.5)), // max. 50% slippage
   skipOriginTxEstimation: optional(boolStr()),
   excludeSources: optional(union([array(string()), string()])),
   includeSources: optional(union([array(string()), string()])),
-  appFeePercent: optional(positiveFloatStr(1)),
+  appFee: optional(positiveFloatStr(1)),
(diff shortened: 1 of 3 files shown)


api/swap/_utils.ts
```diff
@@ -160,10 +160,11 @@ export async function handleBaseSwapQueryParams(
...
   const amount = BigNumber.from(_amount);
 
-  const slippageToleranceNum = parseFloat(slippageTolerance);
-  const appFeePercentNum = appFeePercent
-    ? parseFloat(appFeePercent)
+  const slippageToleranceNum = slippageTolerance
+    ? parseFloat(slippageTolerance)
     : undefined;
+  const slippageNum = parseFloat(slippage);
+  const appFeePercentNum = appFee ? parseFloat(appFee) : undefined;
```
ashwinrava, line 167: Nit for consistency.
```suggestion
  const appFeeNum = appFee ? parseFloat(appFee) : undefined;
```

api/swap/approval/_service.ts
```diff
@@ -68,6 +69,8 @@ export async function handleApprovalSwap(
       ? handleSwapBody(request.body, Number(request.query.destinationChainId))
       : { actions: [] };
 
+  const slippageTolerance = _slippageTolerance ?? slippage * 100;
```
ashwinrava, line 72: I think the null coalescing operator takes precedence right? Then we'd need the brackets so that `_slippageTolerance` doesn't get multiplied by 100 too.

```suggestion
  const slippageTolerance = _slippageTolerance ?? (slippage * 100);
```

ashwinrava (commented):
Looks good, added 2 suggestions.

api/swap/approval/_service.ts
dohaki: ohh good catch!

PR #1736, opened 2025-08-12
chore: disable any-any for strategies with unsupported `sellEntireBalance`

dohaki:


api/_dexes/cross-swap-service.ts
```diff
@@ -1429,10 +1424,19 @@ function _prepCrossSwapQuotesRetrievalA2A(
...
+              // If strict trade type is enabled, we need to check if the destination strategy supports
+              // selling the entire balance.
+              if (crossSwap.strictTradeType) {
+                try {
+                  destinationStrategy.assertSellEntireBalanceSupported();
+                  return true;
+                } catch (error) {
+                  return false;
+                }
+              }
```
ashwinrava, line 1437: Is it intended that we would now potentially pick non-matching strategies? For example, if strictTradeType is true and the origin strategy is LiFi, then we would choose 0x at the destination since it would be the first one which supports `sellEntireBalance`. Just double checking if we also need to compare strategy name on top of this.

ashwinrava (approved):

api/_dexes/cross-swap-service.ts
dohaki: Yes exactly thats what I was aiming for

PR #1743, opened 2025-08-14
chore: re-gen dex sources + tweaks

dohaki:
- Regenerate DEX sources
- Include Uniswap API sources to `/swap/sources` endpoint
- Bump Uniswap API prio
- Tweak LiFi API speed
- Improve upstream RPC error

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "generate:swap-routes": "tsx scripts/generate-swap-routes.ts",
     "generate:swap-routes:mainnet": "yarn generate:swap-routes 1",
     "generate:swap-routes:testnet": "yarn generate:swap-routes 11155111",
+    "generate:dex-sources": "tsx scripts/generate-dex-sources.ts",
     "gas-prices": "tsx scripts/gas-prices.ts"
   },
   "lint-staged": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -111,6 +112,10 @@
     "scripts/generate-swap-routes.ts": [
       "tsx",
       "git add src/data/"
+    ],
+    "api/**/*.ts": [
+      "yarn generate:dex-sources",
+      "git add api/_dexes/*/utils/sources.ts"
```
ashwinrava, line 118: Unfortunately `lint-staged` passes the matched files as arguments to the list of commands. So if `api/1.ts` and `api/2.ts` were being committed, the actual commands run will be like:

1. `yarn generate:dex-sources api/1.ts api/2.ts`
2. `git add api/_dexes/*/utils/sources.ts api/1.ts api2.ts`

It wouldn't matter in this case since the `generate-dex-sources` script doesn't take any arguments and those will be ignored. But if you ever want to not pass these matched files in the future, you can always do this: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments

gsteenkamp89 (approved):

PR #1768, opened 2025-08-20
fix: Solana cron cache gas prices

melisaguevara:
We were trying to estimate Solana gas fees without providing a transaction, but the gas oracle needs one in order to return an estimate. See [here](https://github.com/across-protocol/sdk/blob/master/src/gasPriceOracle/adapters/solana.ts#L16).

### DIFF
--- a/api/cron-cache-gas-prices.ts
+++ b/api/cron-cache-gas-prices.ts
@@ -12,7 +12,7 @@ import {
   resolveVercelEndpoint,
 } from "./_utils";
 import { UnauthorizedError } from "./_errors";
-import { CHAIN_IDs } from "./_constants";
+import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "./_constants";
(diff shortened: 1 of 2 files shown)


api/cron-cache-gas-prices.ts
```diff
@@ -123,7 +123,14 @@ const handler = async (
       Promise.all(
         mainnetChains
           .filter((chain) => chain.chainId !== CHAIN_IDs.LINEA)
-          .map((chain) => updateGasPricePromise(chain.chainId))
+          .map((chain) =>
+            chain.chainId === CHAIN_IDs.SOLANA
+              ? updateGasPricePromise(
+                  chain.chainId,
+                  TOKEN_SYMBOLS_MAP.USDC.addresses[CHAIN_IDs.SOLANA]
```
melisaguevara, line 130: By providing an output token we will use a mock deposit to estimate gas.

ashwinrava (approved):

PR #1773, opened 2025-08-22
drop exclusivity for native deposits

gsteenkamp89:
closes FE-46

### DIFF
--- a/src/views/Bridge/hooks/useBridgeAction/strategies/evm.ts
+++ b/src/views/Bridge/hooks/useBridgeAction/strategies/evm.ts
@@ -155,18 +155,19 @@ export class EVMBridgeActionStrategy extends AbstractBridgeActionStrategy {
     }
 
     let tx: providers.TransactionResponse;
-    const isExclusive =
-      depositArgs.exclusivityDeadline > 0 &&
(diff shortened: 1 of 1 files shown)


src/views/Bridge/hooks/useBridgeAction/strategies/evm.ts
```diff
@@ -155,18 +155,19 @@ export class EVMBridgeActionStrategy extends AbstractBridgeActionStrategy {
...
+
+    if (shouldUseSpokePoolVerifier && spokePoolVerifier) {
       tx = await sendSpokePoolVerifierDepositTx(
         signer,
         {
           ...depositArgs,
           fillDeadline,
+          // drop exclusivity for native deposits
+          exclusivityDeadline: 0,
+          exclusiveRelayer: constants.AddressZero,
```
pxrl, line 170: @gsteenkamp89 These arguments are just passed directly through to the SpokePool, so I think it should be OK to use the arguments provided by `suggested-fees` here.

gsteenkamp89: https://github.com/across-protocol/frontend/pull/1773/commits/2abef491078839a621f09050f4a8ac755d481fa9

PR #1775, opened 2025-08-22
fix: populate transactions page with svm connection

melisaguevara:


gsteenkamp89 (approved):
good catch! thank you!

PR #1779, opened 2025-08-25
chore: Cleanup

pxrl:


melisaguevara (approved):
I pushed a commit to regenerate 0x sources which were being deleted. We'll add a check to avoid these kind of issues in the future.

dohaki (approved):

PR #1791, opened 2025-08-29
Revert "chore: temporary disable Lens"

melisaguevara:
Reverts across-protocol/frontend#1790


gsteenkamp89 (approved):
This looks good. I've tested regular and swap routes to and from Lens and all seems good. 👍

PR #1798, opened 2025-09-02
Fix/send unchecked tx

gsteenkamp89:
closes FE-47

### Problem
Users experienced a 20+ second delay when submitting bridge transactions before being redirected to the deposit status page.

### Root Cause
The `Signer.sendUncheckedTransaction()` method was not behaving as expected - the promise only resolved once the transaction was confirmed, not when submitted to the mempool as it should. 

### Solution
- Updated lockfile to fix `sendUncheckedTransaction` behavior. Updating transitive dependencies (multiple sub-packages of `@ethersproject`) was the real fix.
- Applied unchecked transaction sending to all chains for consistent UX


dohaki (approved):
Nice

PR #1801, opened 2025-09-03
fix: stringify all values in response JSON

ashwinrava:
Sometimes `instanceof BigNumber` doesn't work due to:

- Serialized/deserialized BigNumber that could have lost its prototype
- Conflicting ethers versions from upstream calls could cause type mistmatch
- Bundler optimization

This fixes it:
Screenshot 2025-09-03 at 11 19 21 AM


dohaki (approved):
Great work 🔥

PR #1803, opened 2025-09-03
improve(api): Use host time as fillDeadline base

pxrl:
Using destination chain time (as reported by an RPC provider) has proven to be unreliable when the chain or provider is experiencing issues such as nodes lagging. This can subsequently produce deposits that are unfillable because the fillDeadline is in the past.


api/_dexes/utils.ts
```diff
@@ -491,14 +491,11 @@ export async function extractSwapAndDepositDataStruct(
 }
 
 async function getFillDeadline(spokePool: SpokePool): Promise<number> {
```
pxrl, line 493: nb. we have two `getFillDeadline` utiltiies. One is querying the fillDeadlineBuffer onchain and the other uses a locally-configured value. Do we want to consolidate these?

melisaguevara:
Closing this PR since these changes were included here: https://github.com/across-protocol/frontend/pull/1831

melisaguevara:
Closing this PR since these changes were included here: https://github.com/across-protocol/frontend/pull/1831

PR #1850, opened 2025-09-24
refactor: prepare multi-bridge-strategy support

dohaki:
The goal of this refactor is to set the stage for implementing different bridging strategies (e.g. direct bridge HyperEVM -> HyperCore, OFT, CCTP, etc.). We employ a strategy pattern (similar to the one used for DEXes). Some things I want to add as a fast follow, but left out in this PR to keep the diff small:
- Move a bunch of helpers in `api/_dexes/utils.ts` to `api/_bridges/across` that are only relevant for Across-specific message building
- Move a bunch of helpers in `api/swap/approval/_utils.ts` to `api/_bridges/across` that are only relevant for Across-specific tx building
- Add more safeguards related to bridge capabilities in `api/_dexes/cross-swap-service.ts` (makes most sense probably when we add a new bridge strategy)

### DIFF
--- a/api/_bridges/across/strategy.ts
+++ b/api/_bridges/across/strategy.ts
@@ -42,7 +42,7 @@ export function getAcrossBridgeStrategy(): BridgeStrategy {
       return getBridgeQuoteMessage(crossSwap, appFee);
     },
 
-    getQuoteExactInput: async ({
+    getQuoteForExactInput: async ({
       inputToken,
       outputToken,
       exactInputAmount,
--- a/api/_bridges/across/strategy.ts
+++ b/api/_bridges/across/strategy.ts
@@ -1,0 +1,92 @@
+import {
+  BridgeStrategy,
+  GetExactInputBridgeQuoteParams,
+  BridgeCapabilities,
+  GetOutputBridgeQuoteParams,
+} from "../types";
+import { CrossSwap, CrossSwapQuotes } from "../../_dexes/types";
+import {
+  getBridgeQuoteForExactInput,
+  getBridgeQuoteForOutput,
+} from "../../_utils";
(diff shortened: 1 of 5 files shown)


api/_bridges/across/strategy.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    getBridgeQuoteRecipient: (crossSwap: CrossSwap) => {
+      return getBridgeQuoteRecipient(crossSwap);
+    },
+
+    getBridgeQuoteMessage: (crossSwap: CrossSwap, appFee?: AppFee) => {
+      return getBridgeQuoteMessage(crossSwap, appFee);
+    },
+
+    getQuoteExactInput: async ({
```
ashwinrava, line 45: Nit to match naming convention with `getQuoteForOutput`.

```suggestion
    getQuoteForExactInput: async ({
```

ashwinrava (approved):
Looks awesome, nice rework!

api/_bridges/across/strategy.ts
dohaki: https://github.com/across-protocol/frontend/pull/1850/commits/95663657d7ee95373cb6fa033b1a314a9b25d739

api/_bridges/across/strategy.ts
```diff
@@ -0,0 +1,92 @@
...
+const name = "across";
+const capabilities: BridgeCapabilities = {
+  ecosystems: ["evm", "svm"],
+  supports: {
+    A2A: true,
+    A2B: true,
+    B2A: true,
+    B2B: true,
+    B2BI: true,
+    crossChainMessage: true,
```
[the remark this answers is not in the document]
dohaki: Yes, that's a good point. I will add that as a follow yp

amateima (approved):

PR #1866, opened 2025-10-01
fix: weth incorrectly resolves to eth info

NikolasHaimerl:
# Fix WETH/ETH Symbol Resolution

**Linear:** [ACX-4369](https://linear.app/uma/issue/ACX-4369/incorrect-token-information-in-response-for-wetheth)

## :memo: Summary

This pull request addresses a bug where the `swap/approval` endpoint was incorrectly resolving the symbol for wrapped native tokens (like WETH) as the native token symbol (e.g., ETH). This was happening on chains like Base and Linea where the wrapped native token address is also present in the `TOKEN_SYMBOLS_MAP` under the `ETH` symbol.

## :wrench: Changes

- Modified the `getTokenByAddress` function in `api/_utils.ts` to correctly resolve the token symbol in cases of ambiguity between `ETH` and `WETH`.
- Added a comprehensive test suite for `getTokenByAddress` to ensure the correct symbol is returned for all supported chains.

## :thought_balloon: Rationale

The root cause of this issue lies in the `@across-protocol/constants` package, where some wrapped native token addresses are associated with both `ETH` and `WETH` symbols. The `getTokenByAddress` function was not deterministically picking the correct symbol, sometimes defaulting to `ETH` when it should have been `WETH`.

This pull request implements a solution that prioritizes `WETH` when a token address is associated with both symbols. This is a robust solution because it correctly handles the ambiguity in the constants file and ensures that the correct token symbol is used throughout the application.

The alternative of fixing the constants file is not feasible as it is an external dependency. This solution provides a local fix that is both effective and maintainable.

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -562,6 +562,14 @@ export const getTokenByAddress = (
       }
     }
 
+    // Due to a misconfiguration in the constants file, the WETH address is sometimes associated with both the ETH and WETH symbols.
+    // This is problematic because the SDK will not be able to resolve the correct token if ETH is picked up first.
+    // To fix this, we will check if there is a WETH match and prioritize it over the ETH match.
+    const wethMatch = matches.find(([symbol]) => symbol === "WETH");
+    if (wethMatch) {
+      return wethMatch[1];
+    }
+
     return matches[0][1];
   } catch (error) {
     return undefined;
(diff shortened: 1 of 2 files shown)


NikolasHaimerl:
Closing in favour of https://github.com/across-protocol/frontend/pull/1867

PR #1880, opened 2025-10-07
Revert "fix: incorrect token information in response for wetheth"

NikolasHaimerl:
Reverts across-protocol/frontend#1867

### DIFF
--- a/api/_utils.ts
+++ b/api/_utils.ts
@@ -562,15 +562,6 @@ export const getTokenByAddress = (
       }
     }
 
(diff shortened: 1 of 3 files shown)


amateima (commented):
What's the revert reason?

NikolasHaimerl:
> What's the revert reason?

Master is currently failing on the CI due to some discrepancy between preview and production deployments to Vercel. 
I could not figure out the reason for this difference in the deployed versions of the API. 
Until I do I want to revert the change so the master CI is succeeding again.

dohaki (approved):

PR #1883, opened 2025-10-07
feat(e2e-api): Add axios instance to fix open handles

NikolasHaimerl:
This PR addresses an issue where E2E API tests would hang and fail to exit gracefully, leading to a `TLSWRAP` open handle error in Jest. This was caused by `axios` keeping TCP connections alive after the tests had completed.

### The Fix

1.  **Enabled Open Handle Detection:** The `detectOpenHandles: true` option has been added to `jest.api.config.cjs`. This will help in diagnosing similar issues in the future by providing detailed information about any open handles that are preventing Jest from exiting.

2.  **Created a Custom Axios Instance:** A new, shared `axios` instance (`axiosInstance`) has been created in `e2e-api/utils/config.ts`. This instance is configured with `keepAlive: false` for both HTTP and HTTPS agents.

3.  **Updated Tests:** The `e2e-api/swap/fetch-approval.test.ts` file has been updated to use the new `axiosInstance` for all API requests.

### Why it Works

By default, `axios` uses the Node.js `http` and `https` agents with `keepAlive` enabled. This is a performance optimisation that keeps TCP connections open for reuse in subsequent requests. However, in a test environment, this can cause problems as the open connections prevent the Node.js event loop from emptying, which in turn prevents Jest from exiting.

By creating a custom `axios` instance with `keepAlive: false`, we ensure that a new TCP connection is established for each request and that it is closed immediately after the response is received. This allows the test process to terminate cleanly without any lingering open handles.

### DIFF
--- a/e2e-api/swap/fetch-approval.test.ts
+++ b/e2e-api/swap/fetch-approval.test.ts
@@ -1,8 +1,7 @@
-import axios from "axios";
 import { BigNumber, ethers } from "ethers";
 import { CHAIN_IDs, TOKEN_SYMBOLS_MAP } from "@across-protocol/constants";
 
-import { e2eConfig } from "../utils/config";
+import { e2eConfig, axiosInstance } from "../utils/config";
 
 const SWAP_API_BASE_URL = e2eConfig.swapApiBaseUrl;
 const SWAP_API_URL = `${SWAP_API_BASE_URL}/api/swap/approval`;
@@ -95,7 +94,7 @@ describe("GET /swap/approval", () => {
       } (${testCase.originChainId}) -> ${
         testCase.outputToken.symbol
       } (${testCase.destinationChainId})`, async () => {
-        const response = await axios.get(SWAP_API_URL, {
+        const response = await axiosInstance.get(SWAP_API_URL, {
           params: {
             amount: testCase.amount,
             tradeType,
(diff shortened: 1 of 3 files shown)


test/api/main.test.ts
```diff
@@ -5,6 +5,12 @@ import poolsHandler from "../../api/pools";
 import coingeckoHandler from "../../api/coingecko";
 import availableRouteHandler from "../../api/available-routes";
 import { TypedVercelRequest } from "../../api/_types";
+import { trace } from "@opentelemetry/api";
+import {
+  InMemorySpanExporter,
+  SimpleSpanProcessor,
+} from "@opentelemetry/sdk-trace-base";
+import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
```
dohaki, line 13: Could we also use [mocks](https://jestjs.io/docs/mock-functions#mocking-modules) here?

api/suggested-fees.ts
```diff
@@ -46,9 +46,10 @@ import { parseRole, Role } from "./_auth";
 import { getEnvs } from "./_env";
 import { getDefaultRelayerAddress } from "./_relayer-address";
 import { getRequestId, setRequestSpanAttributes } from "./_request_utils";
-import { tracer } from "../instrumentation";
+import { tracer as defaultTracer } from "../instrumentation";
```
dohaki, line 49: Wondering if we could mock this module in the unit/integration tests

dohaki (commented):
Nice work! Left a question

api/suggested-fees.ts
NikolasHaimerl: That was my original approach as well; unfortunately, we would have to mock it at every place where tests, test code that uses tracers internally, somewhere. I went with the approach that I did to make it easier to understand what is going on, give control to the user to dependency inject the tracer, and only start the real tracer if we are not in the CI (we don't need tracers in the CI AFAIK).

